PR #11, opened 2019-02-17
Fix/publish callback

cbergmiller:
If i am not mistaken the `PublishResult` should always contain the `SubscriptionId` and we can use it to set `subscription_id` attribute if it is not set yet?

### DIFF
--- a/opcua/client/ua_client.py
+++ b/opcua/client/ua_client.py
@@ -3,6 +3,7 @@ Low level binary client
 """
 import asyncio
 import logging
+from typing import List
 
 from opcua import ua
 from ..ua.ua_binary import struct_from_binary, uatcp_to_binary, struct_to_binary, nodeid_from_binary, header_from_binary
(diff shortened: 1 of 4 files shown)


oroulet:
This needs to be triple checked....

oroulet:
I am quite sure this was added due to some race conditions....

oroulet:
Now I remember, this is that you can get a publish result before we are finished creating the internal subscription object. Maybe this issue is away with asyncio but we really need to check there is no await before creation is finished

cbergmiller:
`SubscriptionService.create_subscription` is synchronous but `InternalSession.create_subscription` is a coroutine (because in case of `UaClient` this is a async. network operation). How hard would it be to write a test that triggers the edge case of a publish result that gets emitted right on creation of the subscription?

cbergmiller:
Fixed by 9591355706d357b8cf0db145aefc1167ed52bb6d

PR #15, opened 2019-02-18
update some examples, make them work and add several missing async methods

oroulet:


oroulet:
@cbergmiller that stuff is hard. I tried some examples against the sync opcua lib, to make them work , I add to add some port a few methods to async. but now it looks like there is a race contition somewhere and the subscription overload tests fail!!!! to solve that I should use a lock on some internal ressource like SubsccritionService.subscriptions, but it looks like any method using asyncio.Lock  must be async, which require their callers to be async, etc.. this seems to have no end and since asyncio.Protocol.datareceived is sync then everythin has to go into tasks pushed in to the loop, that becomes a mess.... should I really do that? any opinion?

opcua/client/client.py
```diff
@@ -361,12 +354,13 @@ def _schedule_renew_session(self, renew_session: bool=False):
     async def _renew_session(self):
         """
         Renew the SecureChannel before the SessionTimeout will happen.
-        ToDo: shouldn't this only be done if there was no session activity?
+        In theory we could do that only if no session activity
+        but it does not cost much..
         """
-        server_state = self.get_node(ua.FourByteNodeId(ua.ObjectIds.Server_ServerStatus_State))
+        state_node = await self.nodes.server_state
```
cbergmiller: Does this work? server_state should be a `Node` instance. So is there a coroutine to await here?

opcua/common/node.py
```diff
@@ -654,33 +654,32 @@ def _fill_delete_reference_item(self, rdesc, bidirectional=False):
             rule = ua.ObjectIds.ModellingRule_Mandatory if mandatory else ua.ObjectIds.ModellingRule_Optional
             await self.add_reference(rule, ua.ObjectIds.HasModellingRule, True, False)
 
-    def add_folder(self, nodeid, bname):
-        return create_folder(self, nodeid, bname)
+    async def add_folder(self, nodeid, bname):
```
cbergmiller: If `create_folder` returns a coroutine you simply can return that coroutine from `add_folder`. Then `add_folder` doesn't need to be another coroutine. This will result in slightly better performance (no task will be created). To document this a type hint `-> Coroutine` can be added.

examples/client-minimal.py
```diff
@@ -1,40 +1,14 @@
...
 from opcua import Client, Node, ua
 
 logging.basicConfig(level=logging.INFO)
 _logger = logging.getLogger('opcua')
 
 
-async def browse_nodes(node: Node):
-    """
-    Build a nested node tree dict by recursion (filtered by OPC UA objects and variables).
-    """
```
oroulet: I remove that code becuase this is supposed to be a minimal example for new users. copy it somewhere else if you want

opcua/client/client.py
oroulet: no that cannot work. it is trange I did not get warning for that one...

oroulet: OK, it was never run, since I did ot tested so long...

opcua/common/node.py
oroulet: OK. intreesting point. Are you sure of that? I thought that if create_folder was a coroutine anyway, python would only create one task for the chain add_folder -> create_folder..... must be some optimalization somewhere

oroulet:
Thanks for the update!. This locking issue is interesting to discuss
* my main concern is Protocol.data_received() is this called within the loop? or from another thread? I could not find any info on that. If it is called from another thread. The first thing to do is push it to the loop using a queue or whatever
* then a few places we have things like
```
 for x in self.my_list:
      await x.something() 
```
that means other tasks can come through while we await and modify my_list. we should probably do something like
`await asyncio.gather([x.something() for x in my_list])`
Does it look ok?

oroulet:
Ok I made some tests with data_received() and it looks like it blocks the asyncio loop, so it is really running inside loop and should not give any issue. so most probably we can get away without using any lock, we should simply not have any await in any sensitive code part to protect structures, etc...

cbergmiller:
`data_received` is called by the event loop (not from a thread).
`gather` will run all the tasks _concurrently_. The `await` in the for loop will run the tasks `successively`. If the latter behavior is needed the list could be cloned before iteration:
```python
for x in self.my_list[:]:
      await x.something()
```

oroulet:
OK. small update now and I get an interesting warning:
base_events.py            1274 ERROR    Task exception was never retrieved
future: <Task finished coro=<UaProcessor.close() done, defined at /home/olivier/opcua-asyncio/opcua/server/uaprocessor.py:401> exception=RuntimeError('dictionary changed size during iteration',)>
  File "/home/olivier/opcua-asyncio/opcua/server/uaprocessor.py", line 408, in close
    await self.session.close_session(True)
  File "/home/olivier/opcua-asyncio/opcua/server/internal_server.py", line 347, in close_session
    await self.delete_subscriptions(self.subscription_service.subscriptions.keys())
  File "/home/olivier/opcua-asyncio/opcua/server/internal_server.py", line 422, in delete_subscriptions
    return await self.subscription_service.delete_subscriptions(ids)
  File "/home/olivier/opcua-asyncio/opcua/server/subscription_service.py", line 43, in delete_subscriptions
    if i not in self.subscriptions:
RuntimeError: dictionary changed size during iteration

I looked hard at code in subscription_service.py and I cannot see where we may have an issue. There is no await in the line: `if i not in self.subscription`: ...
Does Python expect that self.subscription is not modified after the test???

cbergmiller:
I suppose it's the `ids` dict key view (of the `subscriptions` dict) that changes during iteration? There is the `await sub.stop()` line..

```python
    async def delete_subscriptions(self, ids):
        self.logger.info("delete subscriptions: %s", ids)
        res = []
        for i in ids:
            if i not in self.subscriptions:
                res.append(ua.StatusCode(ua.StatusCodes.BadSubscriptionIdInvalid))
            else:
                sub = self.subscriptions.pop(i)
                await sub.stop()
                res.append(ua.StatusCode())
        return res
```

oroulet:
@cbergmiller now it works!?!??!?! that took me an awful amount of time. Let me know if you have any more comments

oroulet:
@cbergmiller any idea what is missing in travis.yml file becuase the tests on travis are failing:

```
language: python
python:
  - "3.6"
# command to install dependencies
install:
  - pip install python-dateutil
  - pip install pytz
  - pip install lxml
  - pip install aiofiles
  - pip install asyncio-contextmanager
  - pip install pytest
  - pip install pytest-asyncio
  - pip install cryptography
# command to run tests
script: pytest -v
```

```
item = <Function 'test_value[client]'>
    def pytest_runtest_setup(item):
        for marker, fixture in _markers_2_fixtures.items():
            if marker in item.keywords and fixture not in item.fixturenames:
                # inject an event loop fixture for all async tests
                item.fixturenames.append(fixture)
>       if item.get_closest_marker("asyncio") is not None:
E       AttributeError: 'Function' object has no attribute 'get_closest_marker'
../../../virtualenv/python3.6.3/lib/python3.6/site-packages/pytest_asyncio/plugin.py:184: AttributeError
___________________ ERROR at setup of test_set_value[client] ___________________
```

PR #16, opened 2019-02-22
Sync api

oroulet:
attempt to write a sync API to completely replace python-opcua and dominate the world
@cbergmiller


oroulet:
test passing!!!! so merging

PR #27, opened 2019-03-28
Fix session renewal bug.

hugosilvaguerreiro:
The renewal loop  was sleeping for hundreds of hours

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -346,9 +346,9 @@ class Client:
         but it does not cost much..
         """
         try:
-            duration = min(self.session_timeout, self.secure_channel_timeout) * 0.7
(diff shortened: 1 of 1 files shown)


cbergmiller:
@hugosilvaguerreiro thanks for your contribution! May i ask how you got aware of this issue? I have noticed client sessions failing after roughly one hour in productive use but i didn't have the time to investigate yet. Did you notice something similar?
However if @oroulet doesn't have any objections i would like to merge this soon.

hugosilvaguerreiro:
@cbergmiller you're welcome. I was having some problems with client sessions stopping to work after some time for no apparent reason. I was then experimenting a bit with the session timeouts and noticed that the log for the renewal was never showing up. I don't recall for how long they stayed up, but I'm guessing it was something close to 1 hour since that's the default.

oroulet:
I though I already merged it... that is strange...

PR #35, opened 2019-04-20
removed async_contextmanager dependency, use sqlite3 shortcut methods

### LINKED ISSUE #28: dependency to context manager
@cbergmiller Why do we have dependency to https://pypi.org/project/asyncio-contextmanager/ ?
It seems to be used at one place in https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/server/history_sql.py
What was the issue that made it needed?

cbergmiller:
Fixes #28 
I removed the contextmanager dependency. Using the sqlite3 shortcut methods described here [https://docs.python.org/3/library/sqlite3.html#using-sqlite3-efficiently](https://docs.python.org/3/library/sqlite3.html#using-sqlite3-efficiently) i removed explicit creation of the cursor altogether.

### DIFF
--- a/asyncua/server/history_sql.py
+++ b/asyncua/server/history_sql.py
@@ -2,8 +2,7 @@ import logging
 from typing import Iterable, Coroutine, Optional
 from datetime import timedelta
 from datetime import datetime
-from asyncio import Lock, get_event_loop
-from aiocontext import async_contextmanager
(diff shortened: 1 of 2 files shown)


oroulet:
great!
you could also have used any of the asyncio sql libraries available, but your solution avoids reduces dependency :+1:

Issue #36, opened 2019-04-20
need cleaner module/class import in __init__.py(s) files


oroulet:
Scrutinizer complains a lot about the code in the __init__,py files. @cbergmiller you wrote/changed them. I am not sure how they should be. Someone should find out what the best practices are...
https://scrutinizer-ci.com/g/FreeOpcUa/opcua-asyncio/

oroulet:
My understanding is that it is better to have as little code as possible in these files. as the libray grows (And it is already quite big) importing these modules cost
Only the stuff used absolutely everywhere llike Node should be imported there
https://stackoverflow.com/questions/35727134/module-imports-and-init-py-in-python

cbergmiller:
I can change the import strategy and remove the `__all__` magic. Should improve the Scrutinizer score.

oroulet:
The goal is not to improve the score, I am just thinking that it probably highlight an issue ;-)

PR #37, opened 2019-04-20
Refactored import structure

cbergmiller:
I removed the `__all__ ` magic and refactored the import structure. This is intented as a fix for #36 
This contains **breaking changes** since only a few names are available for import from `asyncua` now: `Node`, `uamethod`, `Client`, and `Server`. This should probably be extended?
I think the old solution was more elegant. Moving things around will now break imports in many modules.


cbergmiller:
Scrutinizer said: 35 issues were fixed

oroulet:
I think it is simple and clean. but if you have a better idea, write it down

cbergmiller:
I agree that simple is better (if it works as intended). We can go with this solution. It may be more familiar for other contributor too. Do you think we have to expose more names to import directly from `asyncua`?

oroulet:
The few best practices I read, seems to agree that we should expose as little as possible. It is better for performance

Issue #25, opened 2019-03-15
scrutinizer report


oroulet:
@cbergmiller looks like scrutinizer code checker does not like you.

![image](https://user-images.githubusercontent.com/2564046/54419278-5650f480-4707-11e9-9ac6-4997562d0d49.png)

 https://scrutinizer-ci.com/g/FreeOpcUa/opcua-asyncio/issues/master?orderField=issueCount&order=desc&honorSelectedPaths=0&page=1

cbergmiller:
That's unfortunate!
Many issues stem from my package refactoring.
e.g.:
`__all__ = (callback.__all__ + connection.__all__ + copy_node_util.__all__ + event_objects.__all__)`
It's valid syntax and a style recommended by David Beazley:
[https://www.dabeaz.com/modulepackage/ModulePackage.pdf](https://www.dabeaz.com/modulepackage/ModulePackage.pdf)

Should we/I refactor this?

oroulet:
That pdf is too long for my brain. What does it say?

is the `__all__ `stuff the recommended way to work?

scrutinizer is not against `__all__`, it just complains that the setting of `__all__` uses variables imported through `from xx import *` so it cannot check. Maybe the goal will be to remove some of the `from xx import *` . A few are really necessary to merge modules, but many are not.

swamper123:
Seems to be solved with the PR. Case closed 🕵️

PR #38, opened 2019-04-24
refactored subscription.modify_monitored_item

cbergmiller:
relates to #25 
Refactoring of common.subscription.modify_monitored_item - Scrutinizer reported that `item_to_change` was not defined for all paths.
New test for `modify_monitored_item`.
Fixes a bug in internal_server.modify_monitored_items (has to be a coroutine).


ronytigo:
How do I use `modify_monitored_item` in the sync wrapper?

Issue #39, opened 2019-04-28
Support single handle or list of handles in Subscription.unsubscribe()


cbergmiller:
Since e.g. `Subscription.subscribe_data_change()` returns a list of handles it would be nice if `Subscription.unsubscribe()` would accept this list in addition to a single handle.
This requires minimal code changes. Typings for the `handle` parameters could further help to clarify the usage of the subscription handles.

PR #43, opened 2019-04-29
Subscription #39

cbergmiller:
Subscription.unsubscribe() now supports a list of handles (I added two tests for that).
Docstrings in Subscription were updated and typings for handles were added.

### DIFF
--- a/asyncua/common/subscription.py
+++ b/asyncua/common/subscription.py
@@ -4,6 +4,7 @@ high level interface to subscriptions
 import asyncio
 import logging
 import collections
+from typing import Union, List, Iterable
 
 from asyncua import ua
 from .events import Event, get_filter_from_event_type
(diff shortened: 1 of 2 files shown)


asyncua/common/subscription.py
```diff
@@ -135,7 +138,8 @@ def _call_datachange(self, datachange: ua.DataChangeNotification):
                     self._handler.datachange_notification(data.node, item.Value.Value.Value, event_data)
                 except Exception:
                     self.logger.exception("Exception calling data change handler")
-            elif hasattr(self._handler, "data_change"):  # deprecated API
+            elif hasattr(self._handler, "data_change"):
+                # deprecated API
                 self.logger.warning("data_change method is deprecated, use datachange_notification")
```
oroulet: @cbergmiller we should remove these deprecated APIs in opcua-asyncio....

cbergmiller: I will look into that!

oroulet:
the tests passed on this PR, but these changes seems to break master: https://travis-ci.org/FreeOpcUa/opcua-asyncio

oroulet:
@cbergmiller

PR #45, opened 2019-05-01
Refactored Subscription

cbergmiller:
**preliminary - do not merge yet**

I worked on the client subscription publish mechanism. There is now less _fire and forget_ but the changes affect the  symmetry between client and server a bit (internally).
The `UaClient` now has a task that cyclically issues PublishRequests. `Subscription` no longer has to call the `publish()` method.
There is much more potential for simplification in the workings of `InternalSession`, `SubscriptionService`, `InternalSubscription`, `MonitoredItemService` ...
But before that i would like to try and draw a UML diagram, it is hard for me to keep that level of complexity in my mind.
This branch needs further testing and discussion. I may try to use it for some productive systems at work for testing.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -7,12 +7,8 @@ from typing import Dict, List
 from asyncua import ua
 from typing import Optional
 from ..ua.ua_binary import struct_from_binary, uatcp_to_binary, struct_to_binary, nodeid_from_binary, header_from_binary
-from ..ua.uaerrors import BadTimeout, BadNoSubscription, BadSessionClosed
+from ..ua.uaerrors import BadTimeout, BadNoSubscription, BadSessionClosed, UaStructParsingError
 from ..common.connection import SecureConnection
-
-
-class ResponseParseError(ValueError):
-    """Error while parsing responses."""
 
 
 class UASocketProtocol(asyncio.Protocol):
@@ -426,7 +422,7 @@ class UaClient:
         self.logger.info("create_subscription success SubscriptionId %s", response.Parameters.SubscriptionId)
         if not self._publish_task or self._publish_task.done():
             # Start the publish cycle if it is not yet running
-            self._publish_task = self.loop.create_task(self._publish_cycle())
+            self._publish_task = self.loop.create_task(self._publish_loop())
         return response.Parameters
 
     async def delete_subscriptions(self, subscription_ids):
@@ -457,10 +453,10 @@ class UaClient:
             response = struct_from_binary(ua.PublishResponse, data)
         except Exception:
             self.logger.exception("Error parsing notification from server")
-            raise ResponseParseError
+            raise UaStructParsingError
         return response
 
-    async def _publish_cycle(self):
+    async def _publish_loop(self):
         """
         Start publish cycle that sends a publish request and waits for the publish response in an endless loop.
         Forward the `PublishResult` to the matching `Subscription` by callback.
@@ -489,7 +485,7 @@ class UaClient:
                 self.logger.info("BadNoSubscription received, ignoring because it's probably valid.")
                 # End task
                 return
-            except ResponseParseError:
+            except UaStructParsingError:
                 ack = None
                 continue
             subscription_id = response.Parameters.SubscriptionId
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -138,12 +138,14 @@ class UASocketProtocol(asyncio.Protocol):
         return True
 
     def _call_callback(self, request_id, body):
-        future = self._callbackmap.pop(request_id, None)
-        if future is None:
+        try:
+            self._callbackmap[request_id].set_result(body)
+        except KeyError:
             raise ua.UaError(
-                f"No request found for requestid: {request_id}, callbacks in list are {self._callbackmap.keys()}"
+                f"No request found for request id: {request_id}, pending are {self._callbackmap.keys()}"
             )
-        future.set_result(body)
+        except asyncio.InvalidStateError:
+            raise ua.UaError(f"Future for request id {request_id} is already done")
 
     def _create_request_header(self, timeout=1000):
         hdr = ua.RequestHeader()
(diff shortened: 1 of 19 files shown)


oroulet:
If this is work in progress start name with 'WIP:' that is the common convention

oroulet:
I have not looked at code yet. But you MUST send publish request after each publish response. Most server complain if they have too large requests queue

cbergmiller:
Thanks for the heads up. I did not change the sequencing of the publish request, it remains the same. I just moved it into an endless loop inside of an task in UaClient. This reduces complexity in my opinion.

asyncua/client/ua_client.py
```diff
@@ -406,117 +409,103 @@ def disconnect_socket(self):
...
+                self.logger.info("BadNoSubscription received, ignoring because it's probably valid.")
+                # End task
+                return
+            # parse publish response
+            try:
+                response = struct_from_binary(ua.PublishResponse, data)
+            except Exception:
+                # INFO: catching the exception here might be obsolete because we already
+                #       catch BadTimeout above. However, it's not really clear what this code
+                #       does so it stays in, doesn't seem to hurt.
```
oroulet: here we catch any parsing error. On the network you cannot trust what you receive anyway. Not sure what you mean it has to do with timeout

asyncua/client/ua_client.py
```diff
@@ -406,117 +409,103 @@ def disconnect_socket(self):
...
-                await self._sub_cond.wait()
-                data = self._sub_data_queue.pop(0)
-                await self._call_publish_callback(data)
-    """
+    def publish(self):
+        """
+        This is just for maintaining symmetry.
+        This client takes care of sending PublishRequest internally.
+        """
+        pass
```
oroulet: should't we make the method really send a publish request? some people may want to do it manually. Otherwise we should probably deleted it

asyncua/common/subscription.py
```diff
@@ -89,27 +84,18 @@ def __init__(self, server, params, handler):
...
 
     async def init(self):
-        response = await self.server.create_subscription(self.parameters, self.publish_callback)
+        response = await self.server.create_subscription(self.parameters, callback=self.publish_callback)
         self.subscription_id = response.SubscriptionId  # move to data class
         self.logger.info('Subscription created %s', self.subscription_id)
-        # Send a publish request so the server has one in its queue
-        # Servers should always be able to handle at least on extra publish request per subscriptions
-        await self.server.publish()
+        #self.server.publish()
```
oroulet: do not leave dead commented code

asyncua/server/internal_server.py
```diff
@@ -43,7 +34,19 @@ def __init__(self, serv, cap=None):
...
 
+def default_user_manager(iserver, isession, username, password):
+    """
+    Default user_manager, does nothing much but check for admin
+    """
+    if iserver.allow_remote_admin and username in ("admin", "Admin"):
+        isession.user = User.Admin
+    return True
+
+
```
oroulet: why do I see this? this seems to be an old diff?

oroulet:
As far as I understood you are replacing some callbacks with async/await. This is a good idea

asyncua/client/ua_client.py
cbergmiller: This comment was added in f88dce0d99886226eb7ef5226c3ddec067a2307a back in 2016
I can replace or remove the comment. Also i am not sure if it is the right thing to send an Publish Response without acknowledgement when the Pbulish Response cannot be parsed.

asyncua/client/ua_client.py
cbergmiller: That is a good point. I will make it possible to send PublishRequest via this public method again.

asyncua/server/internal_server.py
cbergmiller: Probably because i moved `InternalSession` into a new file for easier access.

cbergmiller:
Thanks for your review! Yes i am trying to remove some of the callbacks - at least in cases where this leads to cleaner architecture.

asyncua/client/ua_client.py
```diff
@@ -3,14 +3,18 @@
...
+from typing import Dict, List
 from asyncua import ua
 from typing import Optional
 from ..ua.ua_binary import struct_from_binary, uatcp_to_binary, struct_to_binary, nodeid_from_binary, header_from_binary
 from ..ua.uaerrors import BadTimeout, BadNoSubscription, BadSessionClosed
 from ..common.connection import SecureConnection
 
 
+class ResponseParseError(ValueError):
+    """Error while parsing responses."""
```
oroulet: I am quite sure there is a PArsingError or similar already defined somewhere. it should be shared between server and client. I found it in asyncua/ua/uaerrors/_base.py... the name is strange... maybe do a grep and rename...

asyncua/client/ua_client.py
```diff
@@ -422,7 +426,7 @@ def disconnect_socket(self):
         self.logger.info("create_subscription success SubscriptionId %s", response.Parameters.SubscriptionId)
         if not self._publish_task or self._publish_task.done():
             # Start the publish cycle if it is not yet running
-            self._publish_task = self.loop.create_task(self._publish())
+            self._publish_task = self.loop.create_task(self._publish_cycle())
```
oroulet: publish_loop? or publish_task? publish_cycle sounds a  bit strange..

oroulet: ususally a cycle is path of a loop

cbergmiller:
I consider this finished and would like to merge these changes.
To recap what this includes:
- Documentation and Typings for class methods for better maintainability in the future
- `UaClient` now has a `_publish_task` that waits for the publish response in an loop (this simplifies `Subscription`)
- Added support for servers returning a publish response for subsription ID 0 (meaning there are no more subscriptions) as to spec. Part 4 - Section 5.13.5
- some cleanup and fixed typos

asyncua/client/ua_client.py
```diff
@@ -279,6 +280,8 @@ def disconnect_socket(self):
 
     async def close_session(self, delete_subscriptions):
         self.logger.info("close_session")
+        if self._publish_task and not self._publish_task.done():
+            self._publish_task.cancel()
```
oroulet: should'nt we wait for task? just wondering

asyncua/server/internal_server.py
```diff
@@ -3,38 +3,30 @@
...
 except ImportError:
     logging.getLogger(__name__).warning("cryptography is not installed, use of crypto disabled")
-    use_crypto = False
-
-
-class SessionState(Enum):
-    Created = 0
-    Activated = 1
-    Closed = 2
+    uacrypto = False
```
oroulet: that test should be removed, now we have a hard dependency on crypto

PR #52, opened 2019-05-19
LocalizedText extended to handle Locale

jikalousek:
I implemented the Locale field in LocalizedText class in the same way as it was done for Text. Some affected methods were changed accordingly. Also, I made changes in xmlimporter and xmlparser to support import of ListOfLocalizedText. I implemented a few test cases and then I run testing to validate the changes.


oroulet:
looks OK as far as I can see :+1:

PR #62, opened 2019-06-17
WIP: Add commit instruction to save database transaction

### LINKED ISSUE #60: history data are never saved in sqlite database
Hello,

I try do test history feature on the asyncio-server unsuccessfully.

I use the `server-datavalue-history.py` example script that I modify like this:

```
import asyncio
import sys
sys.path.insert(0, "..")
import time
import math


from asyncua import ua, Server
from asyncua.server.history_sql import HistorySQLite


async def main():

    # setup our server
    server = Server()
    
    # Configure server to use sqlite as history database (default is a simple memory dict)
    server.iserver.history_manager.set_storage(HistorySQLite("my_datavalue_history.sql"))
    
    # initialize server 
    await server.init()

    server.set_endpoint("opc.tcp://0.0.0.0:4840/freeopcua/server/")

    # setup our own namespace, not really necessary but should as spec
    uri = "http://examples.freeopcua.github.io"
    idx = await server.register_namespace(uri)

    # get Objects node, this is where we should put our custom stuff
    objects = server.get_objects_node()

    # populating our address space
    myobj = await objects.add_object(idx, "MyObject")
    myvar = await myobj.add_variable(idx, "MyVariable", ua.Variant(0, ua.VariantType.Double))
    await myvar.set_writable()  # Set MyVariable to be writable by clients
    print(myvar)

    # starting!
    await server.start()

    # enable data change history for this particular node, must be called after start since it uses subscription
    await server.historize_node_data_change(myvar, period=None, count=100)

    try:
        count = 0
        while True:
            time.sleep(1)
            count += 0.1
            await myvar.set_value(math.sin(count))
            print(count)

    finally:
        # close connection, remove subscriptions, etc
        await server.stop()

if __name__ == "__main__":
    loop.set_debug(True)
    loop.run_until_complete(main())
```

The data are never saved in sqlite3 database.

Has someone already had this problem?

okyame:
close #60

### DIFF
--- a/asyncua/server/history_sql.py
+++ b/asyncua/server/history_sql.py
@@ -37,7 +37,8 @@ class HistorySQLite(HistoryStorageInterface):
         self.logger.info('Historizing SQL connection closed')
 
     async def _execute_sql(self, sql: str, params: Iterable = None):
-        return await self._loop.run_in_executor(None, self._conn.execute, sql, params or ())
+        await self._loop.run_in_executor(None, self._conn.execute, sql, params or ())
+        await self._loop.run_in_executor(None, self._conn.commit)
 
     async def new_historized_node(self, node_id, period, count=0):
         table = self._get_table_name(node_id)


okyame:
Doesn't work when there is a sql exception

okyame:
Someone know how to re-throw an exception ?

```
    async def _execute_sql(self, sql: str, params: Iterable = None):
        try:
            await self._loop.run_in_executor(None, self._conn.execute, sql, params or ())
            await self._loop.run_in_executor(None, self._conn.commit)
        except sqlite3.Error as e:
            raise 
```

oroulet:
in your example you rethrow. What else do you want.
btw using an executor for that is not nice, there are many asyncio db modules. but this is another subject...

okyame:
My example doesn't work. Only this work:

```
async def _execute_sql(self, sql: str, params: Iterable = None):
        return await self._loop.run_in_executor(None, self._conn.execute, sql, params or ())
```
So In order to factorise  the code, I look for a function that will avoid to repeat "commit()" instruction in others functions.

okyame:
What about use this example to foactorise "UPDATE" And "CREATE" instructions

```
async def execute_sql_delete(self, condition: str, args: Iterable, table: str, node_id):
        try:
            await self._execute_sql(f'DELETE FROM "{table}" WHERE {condition}', args)
        except sqlite3.Error as e:
            self.logger.error("Historizing SQL Delete Old Data Error for %s: %s", node_id, e)
        
         await self._loop.run_in_executor(None, self._conn.commit)
```

oroulet:
Do we really want to commit on every insert? How is it done in python-opcua

okyame:
It is committed on every insert

> How is it done in python-opcua

committed on every insert

okyame:
I write finally this function 

```
async def execute_sql(self, query: str, params: Iterable = None,  error_message: str = None, commit=False):
        try:
            await self._execute_sql(query, params)
            if commit:
                await self._loop.run_in_executor(None, self._conn.commit)
        except sqlite3.Error as e:
            error_message = error_message or "Historizing SQL Execute Error"
            self.logger.error("%s: %s", error_message, e)
```

And call it as this:

```
  async def new_historized_node(self, node_id, period, count=0):
        table = self._get_table_name(node_id)
        self._datachanges_period[node_id] = period, count
        # create a table for the node which will store attributes of the DataValue object
        # note: Value/VariantType TEXT is only for human reading, the actual data is stored in VariantBinary column
        await self.execute_sql(query=f'CREATE TABLE "{table}" (_Id INTEGER PRIMARY KEY NOT NULL,'
                                    ' ServerTimestamp TIMESTAMP,'
                                    ' SourceTimestamp TIMESTAMP,'
                                    ' StatusCode INTEGER,'
                                    ' Value TEXT,'
                                    ' VariantType TEXT,'
                                    ' VariantBinary BLOB)',
                                params=None,
                                error_message=f'Historizing SQL Table Creation Error for "{node_id}"',
                                commit=True,
                                )
```

okyame:
Sometimes I get this error message, and impossible to save anything in database
```
ERROR:asyncua.server.history_sql:Historizing SQL Insert Error for NumericNodeId(ns=2;i=2041): cannot start a transaction within a transaction
```
I think than executor method is not safe.  Do you know other asyncio db modules?
what about  aiosqlite wrapper : https://github.com/jreese/aiosqlite ?

cbergmiller:
I suspect the problem could be that `run_in_executor` uses a thread pool and the _execute_ and _commit_ commands could be handed of to different threads from the pool.
To make things safer, `_execute_sql` should only use one `run_in_executor` call to execute both _execute_ and _commit_.
Also, the `_execute_sql` method should use a [Lock](https://docs.python.org/3/library/asyncio-sync.html#lock) to prevent generation of multiple threads that try to commit to the db concurrently.
I am not against using an asyncio db module but they all have to use threads under the hood since file access (and the sqlite3 module) is blocking I/O.

cbergmiller:
I will draft a new PR with my suggested changes.

oroulet:
I would use something like that https://github.com/jreese/aiosqlite/blob/master/README.md but the person doing the work decides..

oroulet:
The use threads in background but this is completely fine. This is by design and simpler than making our own stuff

oroulet:
outdated PR isnt'it? should I close?

PR #63, opened 2019-06-17
fix server datavalur history example

okyame:


Nanahara:
Hello,

At line 17 : 
`server.iserver.history_manager.set_storage(HistorySQLite("my_datavalue_history.sql"))`

HistorySQLite should be init first by call init() function. I suggest to write this:
`sqlite_history = HistorySQLite("my_datavalue_history.sql")`
`sqlite_history.init()`
`server.iserver.history_manager.set_storage(sqlite_history)`

PR #69, opened 2019-07-09
Use correct signature algorithm in create session response

helmut-jacob:
Previously this was hardcoded to sha1. Fix this by introducing a
new URI field in the SecurityPolicy classes carrying the correct
URIs.

This fixes complaints by some UA clients that match requested
security profiles against the server response and warn the user.


oroulet:
I did not see that you also opened a PR here.. I cherry.picked already from ptyhon-.opcua. Thanks anyway!

PR #75, opened 2019-07-25
Fix for #74 - make InternalSession.add_method_callback sync.

### LINKED ISSUE #74: Missing await in server.link_method
I get `RuntimeWarning: coroutine 'InternalSession.add_method_callback' was never awaited` message when using the link_method() function of server.py
This works for me:

async def link_method(self, node, callback):
        """
        Link a python function to a UA method in the address space; required when a UA method has been imported
        to the address space via XML; the python executable must be linked manually
        :param node: UA method node
        :param callback: python function that the UA method will call
        """
        await self.iserver.isession.add_method_callback(node.nodeid, callback)

cbergmiller:
fixes #74

### DIFF
--- a/asyncua/common/manage_nodes.py
+++ b/asyncua/common/manage_nodes.py
@@ -371,7 +371,7 @@ async def _create_method(parent, nodeid, qname, callback, inputs, outputs):
             datatype=ua.ObjectIds.Argument
         )
     if hasattr(parent.server, "add_method_callback"):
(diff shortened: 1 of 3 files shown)


cbergmiller:
Another Solution would be to make `InternalSession.add_method_callback` synchronous. I doesn't have to be async as far as i can see.

oroulet:
yes add_method_callback jusrt write an internal attrtibut so it should not be async

Issue #79, opened 2019-08-27
Memory increases monotonically from repetitive client connects and disconnects.


dannmartens:
We've observed process memory ballooning to several GBs over time, starting from an initial 80 MB, running an asyncio OPC UA server.

The process is very active doing other things as well, but as long as no OPC UA clients connect, memory stays constant. 

A burn-in test app was created to perform a continuous connect, write value, disconnect cycle and we can observe the memory footprint increasing slowly. 

I ran mem_top against the server with the following results:
`refs:
18566   <class 'dict'> {1: PublishResult(SubscriptionId:78, AvailableSequenceNumbers:[1], MoreNotifications:False, Notifica
10602   <class 'dict'> {'__module__': 'asyncua.ua.object_ids', 'Null': 0, 'Boolean': 1, 'SByte': 2, 'Byte': 3, 'Int16': 4,
9282    <class 'list'> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
9281    <class 'list'> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
9280    <class 'list'> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
9279    <class 'list'> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
9278    <class 'list'> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
9277    <class 'list'> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27,
[...]

PR #80, opened 2019-08-27
Fix #71 (missing publish call in internal server subscriptions)

### LINKED ISSUE #71: Datachange not acknowledged in History
Hello,

I try to test history function and I use `server-datavalue-history.py` from #63 . I see in log information that the datachange notifications are never acknowledged: see attached file: 
[log.txt](https://github.com/FreeOpcUa/opcua-asyncio/files/3397317/log.txt)

We see that `AvailableSequenceNumbers` list is growing up indefinitely with opcua-asyncio  whereas a publish request with acks is generated with python-opcua

Have you an idea about this issue ?

N.B: I run server example without connected client.

cbergmiller:
On internal server subscriptions `InternalSubscription._not_acknowledged_results` kept growing onexternal data changes. This was a bug that was introduced in the #45 Subscription refactoring.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -118,8 +118,8 @@ class InternalSession:
         """COROUTINE"""
         return self.iserver.method_service.call(params)
 
-    async def create_subscription(self, params, callback=None):
-        result = await self.subscription_service.create_subscription(params, callback)
+    async def create_subscription(self, params, callback=None, is_for_client=False):
+        result = await self.subscription_service.create_subscription(params, callback, is_for_client)
         self.subscriptions.append(result.SubscriptionId)
         return result
 
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -118,8 +118,8 @@ class InternalSession:
         """COROUTINE"""
         return self.iserver.method_service.call(params)
 
-    async def create_subscription(self, params, callback=None, is_for_client=False):
-        result = await self.subscription_service.create_subscription(params, callback, is_for_client)
+    async def create_subscription(self, params, callback=None):
+        result = await self.subscription_service.create_subscription(params, callback, external=self.external)
         self.subscriptions.append(result.SubscriptionId)
         return result
 
--- a/asyncua/server/internal_subscription.py
+++ b/asyncua/server/internal_subscription.py
@@ -17,7 +17,7 @@ class InternalSubscription:
     """
 
     def __init__(self, loop: asyncio.AbstractEventLoop, data: ua.CreateSubscriptionResult, aspace: AddressSpace,
-                 callback=None):
+                 callback=None, is_for_client=False):
         self.logger = logging.getLogger(__name__)
         self.loop: asyncio.AbstractEventLoop = loop
         self.data: ua.CreateSubscriptionResult = data
@@ -32,6 +32,7 @@ class InternalSubscription:
         self._keep_alive_count = 0
         self._publish_cycles_count = 0
         self._task = None
+        self._is_for_client = is_for_client
 
     def __str__(self):
         return f"Subscription(id:{self.data.SubscriptionId})"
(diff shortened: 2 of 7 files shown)


cbergmiller:
This is a fix for #71 and possibly #79

asyncua/server/internal_subscription.py
```diff
@@ -97,6 +97,7 @@ def publish_results(self):
         if result is not None:
             self.logger.info('publish_results for %s', self.data.SubscriptionId)
             self.pub_result_callback(result)
+            self.publish([result.NotificationMessage.SequenceNumber])
```
oroulet: Is that code only call when subscription is server side? If it is called from client too then we will receive dobble publish....

cbergmiller: It _should_ be the same way it used to be before #45.
`InternalSubscription.publish` does not do much:
```python
def publish(self, acks: Iterable[int]):
        """
        Reset publish cycle count, acknowledge PublishResults.
        :param acks: Sequence number of the PublishResults to acknowledge
        """
        self.logger.info("publish request with acks %s", acks)
        self._publish_cycles_count = 0
        for nb in acks:
            self._not_acknowledged_results.pop(nb, None)
```
The `pop` already had the `None` default argument - in case the result was already acked.
I will check the client subscription too. If possible i will add a test case to prevent regression bugs in the future.

cbergmiller:
I removed the `publish` call from the `InternalSubscription.publish_results` again since it would break the ack logic of the PublishResults that are sent from Servers to Clients.
I did not want to go back to sending Publish Acks in `Subscription.publish_callback`. Instead i added a flag to `InternalSubscription` that is only set if the Subscription was created for a Client on the Server.
If the Subscription is internal to the Server the Publish Results are now no longer stored for republishing.

asyncua/server/internal_session.py
```diff
@@ -118,8 +118,8 @@ def call(self, params):
         """COROUTINE"""
         return self.iserver.method_service.call(params)
 
-    async def create_subscription(self, params, callback=None):
-        result = await self.subscription_service.create_subscription(params, callback)
+    async def create_subscription(self, params, callback=None, is_for_client=False):
+        result = await self.subscription_service.create_subscription(params, callback, is_for_client)
```
oroulet: in fact the InternalSession object already has a `self.external` property which contains that information

asyncua/server/internal_subscription.py
```diff
@@ -92,10 +93,14 @@ def publish_results(self):
...
         result = None
         if self.has_published_results():
-            self._publish_cycles_count += 1
+            if self._is_for_client:
+                self._publish_cycles_count += 1
             result = self._pop_publish_result()
         if result is not None:
             self.logger.info('publish_results for %s', self.data.SubscriptionId)
+            # The callback can be:
+            #    Subscription.publish_callback -> server internal subscription
```
oroulet: Subscription is also used on client side. So you need to document that Subscription.publish_callback is used for server side internal subcriptions AND client side

oroulet: OK wrong comment. This was a differnt file..

asyncua/server/internal_session.py
cbergmiller: Oops, thanks i will look into that!

asyncua/server/internal_subscription.py
```diff
@@ -17,7 +17,7 @@ class InternalSubscription:
     """
 
     def __init__(self, loop: asyncio.AbstractEventLoop, data: ua.CreateSubscriptionResult, aspace: AddressSpace,
-                 callback=None):
+                 callback=None, is_for_client=False):
```
oroulet: instead of calling that parameter is_for_client. it is probably better to explicitly state what is changes: no_acks or do_not_expect_acks

asyncua/server/internal_subscription.py
```diff
@@ -92,10 +93,14 @@ def publish_results(self):
...
-            self._publish_cycles_count += 1
+            if self._is_for_client:
+                self._publish_cycles_count += 1
             result = self._pop_publish_result()
         if result is not None:
             self.logger.info('publish_results for %s', self.data.SubscriptionId)
+            # The callback can be:
+            #    Subscription.publish_callback -> server internal subscription
+            #    UaProcessor.forward_publish_response -> client subscription
             self.pub_result_callback(result)
```
oroulet: you could call self.publish(result.Acknowledgements) here instead of disabling sending them. Not sure what is best....

cbergmiller: I think this would break the ability to republish not acknowledged results (if so requested by the client). For server internal subscriptions the acknowledgement (and republishing) is not needed and can be disabled (?). I will try to do more testing to be sure.

oroulet: yes should be fine to disable it for internal session

oroulet:
We can merge, can't we?

@dannmartens Could you try that branch and tell us if it changes anything?

dannmartens:
@oroulet I will try it out!

dannmartens:
@oroulet I am going to use the branch 'fix_71' and deploy again, so the test server can run over the week-end.

dannmartens:
Okay, the burn-in test server has been up for more than 48 hours now, receiving writes from a subscribed client without any significant change in the memory consumption. There is just a normal small undulation of allocation and garbage collection, order of magnitude 1 MB to 10 MB.

👌

oroulet:
so maybe that was the only memeroy leak :tada:

PR #89, opened 2019-10-04
remove unnecessary production dependency into tests requirement

memetb:
As per title. Production systems shouldn't need to have `pytest-runner` installed.

(Note: possibly a duplicated effort as branch `asyncua` has a different setup.py file which does not contain this problem)


oroulet:
thanks. it was obvioulsy a little more complicated than initially thought ;-)

PR #90, opened 2019-10-11
Allow Asynchronous Method Calls

chrisjbremner:
I occasionally run methods from my server that involve some I/O or other logic that doesn't execute relatively quickly, and therefore the server will block on these methods for a while if they are called synchronously. I have got around this in the past by running these methods in a thread and then emitting an event when it completes, but I thought it would make sense to implement the ability to run `async` methods directly. My goal is best demonstrated by example:

```python
@uamethod
async def func1(parent, value):
    _logger.info(f"Starting async sleep for {value} seconds")
    await asyncio.sleep(value)
    _logger.info("Ending async sleep")

@uamethod
def func2(parent, value):
    _logger.info(f"Starting sync sleep for {value} seconds")
    time.sleep(value)
    _logger.info("Ending sync sleep")
```

Before, only `func1` was possible, as you couldn't call coroutines from `MethodService.call`. In this PR, I added some logic into the `MethodService.call` method that determines if the method is a coroutine or not, and if it is, it will `await` it.

I created a small script with these functions as well as a variable that increments in the background every 1 sec. From this script's output you can see that the `async` method does not block, whereas the synchronous method does (thus preventing the variable from incrementing):

```
2019-10-11 15:48:43.648 INFO     call request
2019-10-11 15:48:43.649 INFO     Calling: CallMethodRequest(ObjectId:TwoByteNodeId(i=85), MethodId:StringNodeId(ns=2;s=ServerMethod), InputArguments:[Variant(val:3,type:VariantType.Int64)])
2019-10-11 15:48:43.650 INFO     Starting async sleep for 3 seconds
2019-10-11 15:48:43.698 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 0.7
2019-10-11 15:48:44.699 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 0.8
2019-10-11 15:48:45.702 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 0.9
2019-10-11 15:48:46.657 INFO     Ending async sleep
2019-10-11 15:48:46.705 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.0
2019-10-11 15:48:47.700 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.1
2019-10-11 15:48:48.701 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.2
2019-10-11 15:48:49.702 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.3
2019-10-11 15:48:50.695 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.4
2019-10-11 15:48:51.497 INFO     call request
2019-10-11 15:48:51.498 INFO     Calling: CallMethodRequest(ObjectId:TwoByteNodeId(i=85), MethodId:StringNodeId(ns=2;s=ServerMethod2), InputArguments:[Variant(val:3,type:VariantType.Int64)])
2019-10-11 15:48:51.499 INFO     Starting sync sleep for 3 seconds
2019-10-11 15:48:54.500 INFO     Ending sync sleep
2019-10-11 15:48:54.509 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.5
2019-10-11 15:48:55.510 INFO     Set value of Node(NumericNodeId(ns=2;i=2)) to 1.6
```

I have modified this logic such that it works regardless of whether you choose to use the `uamethod` decorator or not.

While testing this, I also found that `server-methods.py` did not get ported over to this from `python-opcua`, so I updated it and added examples of `async` methods

### DIFF
--- a/asyncua/common/methods.py
+++ b/asyncua/common/methods.py
@@ -86,7 +86,7 @@ def uamethod(func):
 
 def _format_call_inputs(parent, *args):
         if isinstance(parent, ua.NodeId):
-        return (parent, *[arg.Value for arg in args])
+            return (parent, *[arg.Value for arg in args])
         else:
             self = parent
             parent = args[0]
--- a/asyncua/common/methods.py
+++ b/asyncua/common/methods.py
@@ -2,6 +2,8 @@
 High level method related functions
 """
 
+from asyncio import iscoroutinefunction
+
 from asyncua import ua
 
 
@@ -67,14 +69,32 @@ def uamethod(func):
     Method decorator to automatically convert
     arguments and output to and from variants
     """
-    def wrapper(parent, *args):
+
+    if iscoroutinefunction(func):
+        async def wrapper(parent, *args):
+            func_args = _format_call_inputs(parent, *args)
+            result = await func(*func_args)
+            return _format_call_outputs(result)
+
+    else:
+        def wrapper(parent, *args):
+            func_args = _format_call_inputs(parent, *args)
+            result = func(*func_args)
+            return _format_call_outputs(result)
+    return wrapper
+
+
+def _format_call_inputs(parent, *args):
         if isinstance(parent, ua.NodeId):
-            result = func(parent, *[arg.Value for arg in args])
+        return (parent, *[arg.Value for arg in args])
         else:
             self = parent
             parent = args[0]
             args = args[1:]
-            result = func(self, parent, *[arg.Value for arg in args])
+        return (self, parent, *[arg.Value for arg in args])
+
+
+def _format_call_outputs(result):
         if result is None:
             return []
         elif isinstance(result, ua.CallMethodResult):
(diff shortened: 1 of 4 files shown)


oroulet:
There is already a similar PR. We need to merge them

oroulet:
@chrisjbremner this looks very good! but there is obviously a typo somewhere. Look at test results: https://travis-ci.org/FreeOpcUa/opcua-asyncio/jobs/596824986
Can you fix it and I will merge asap

chrisjbremner:
Fixed! Looks like I messed something up with my IDE's GitHub integration.

I like some of the logging you have in your PR surrounding these methods. Is the idea that you would merge this now and then incorporate the changes in your PR after? Or should I include those log statements into my PR?

oroulet:
I am merging your stuff first. My PR has some random locking. I need to reduce it as much as I can and then try to debug it..

PR #95, opened 2019-10-21
revmove dispensable loop

Nanahara:
Hello,

I think this loop is dispensable in _read_and_copy_attrs function in copy_node_util file: 
```
for name in names:
        results = await node_type.get_attributes(attrs)
```


oroulet:
looks correct. thanks

PR #98, opened 2019-11-01
Await connection lost tasks when stopping internal server

chrisjbremner:
I was writing some tests for a client and server, and I found that occasionally, when I stopped the client, then the server, and then closed the event loop, there were still some pending tasks that needed to be awaited. After some investigation, I found that the `OPCUAProtocol` creates a few tasks when a connection is lost but since `connection_lost` is a synchronous function, the tasks cannot be awaited. In my use case, where I was closing the client immediately before the server, the connection lost tasks didn't have enough time to complete.

In this PR, I keep track of these tasks, an ensure that they are all awaited when the server is stopped, so that there are no pending tasks left when the event loop is closed.

### DIFF
--- a/asyncua/server/binary_server_asyncio.py
+++ b/asyncua/server/binary_server_asyncio.py
@@ -18,7 +18,7 @@ class OPCUAProtocol(asyncio.Protocol):
     Instantiated for every connection.
     """
 
-    def __init__(self, iserver: InternalServer, policies, clients):
+    def __init__(self, iserver: InternalServer, policies, clients, protocol_tasks):
         self.peer_name = None
         self.transport = None
         self.processor = None
@@ -26,6 +26,7 @@ class OPCUAProtocol(asyncio.Protocol):
         self.iserver: InternalServer = iserver
         self.policies = policies
         self.clients = clients
+        self.protocol_tasks = protocol_tasks
         self.messages = asyncio.Queue()
         self._task = None
 
@@ -49,12 +50,12 @@ class OPCUAProtocol(asyncio.Protocol):
         self.transport.close()
         self.iserver.asyncio_transports.remove(self.transport)
         closing_task = self.iserver.loop.create_task(self.processor.close())
-        self.iserver.tasks_to_await.append(closing_task)
+        self.protocol_tasks.append(closing_task)
         if self in self.clients:
             self.clients.remove(self)
         self.messages.put_nowait((None, None))
         self._task.cancel()
-        self.iserver.tasks_to_await.append(self._task)
+        self.protocol_tasks.append(self._task)
 
     def data_received(self, data):
         self._buffer += data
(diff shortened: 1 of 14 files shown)


chrisjbremner:
@oroulet or @cbergmiller I'd be interested on hearing your thoughts on whether this is a sensible change or not?

For more clarity, the original error I was intermittently seeing was this:
```
task: <Task pending coro=<InternalSubscription._subscription_loop() running at ...\lib\site-packages\asyncua\server\internal_subscription.py:74> wait_for=<Future cancelled created at ...\appdata\local\programs\python\python37-32\Lib\asyncio\base_events.py:391> cb=[<TaskWakeupMethWrapper object at 0x06ECCD90>()] created at ...\lib\site-packages\asyncua\server\internal_subscription.py:51>
Task was destroyed but it is pending!
```

oroulet:
Sorry forgot that one. I agree 100% that we should keep track of tasks. Anything else is bad practice. But I was wondering why making it so complicated. Can't you just keep the task inside BinaryServer and and await it when closing? When pushing it in some list on internal server?

oroulet:
Ok I see we have an issue here. Connection_lost is not a coroutine. I am really wondering what is the correct way to do that...

oroulet:
I never liked the api of the socket server

chrisjbremner:
Yes the fact that `connection_lost` is not a coroutine complicates things. The only other alternative I could think of would be registering callbacks on the tasks that get created inside of `connection_lost` and then waiting for those to be finished inside `InternalServer,stop`, but that felt messier than what I currently have.

oroulet:
Can't we make BinaryServer take care of these tasks? Since it is the one producing them. There are already to many things in innternal_server. Just pass one more argument to protocol and add a stop method to BinaryServer if does not have one yet.

chrisjbremner:
Good idea, was something like https://github.com/FreeOpcUa/opcua-asyncio/pull/98/commits/58c2262d4ae75ee0ab5327f27d3a00a4b598ec41 what you had in mind? Note: I still need to double check that this change still resolves the original issue, which I can do tomorrow.

asyncua/server/binary_server_asyncio.py
```diff
@@ -132,6 +141,11 @@ def _make_protocol(self):
         self.logger.info('Closing asyncio socket server')
         for transport in self.iserver.asyncio_transports:
             transport.close()
+
+        # Wait for all protocol closing tasks to complete
+        await asyncio.gather(*self.protocol_tasks, return_exceptions=True)
```
oroulet: We should log return values, shouldn't we?

chrisjbremner: Sure, no harm in doing so. Done in https://github.com/FreeOpcUa/opcua-asyncio/pull/98/commits/f2ea114fa4acdb461cf7581eba2f04e6559833db

PR #101, opened 2019-11-20
Started to implement Alarms and Conditions

mar-ar:
Hey guys,

I started to implement alarms and conditions. It already works quite well but I have still some problems.

I also tested everything with UaExpert and it works.
image

I consider to improve it to create the condition optionally in the adress space.

My main problem at the moment is that I cant create more than one instance of a condition. This is maybe important for alarms.

Please check the code and give me some feedback and maybe you have some ideas for my problem.


asyncua/server/event_generator.py
```diff
@@ -75,29 +84,20 @@ def __str__(self):
 
     __repr__ = __str__
 
-    def trigger(self, time_attr=None, message=None):
+    def trigger(self, time=None, message=None):
```
oroulet: I do not like using variable names that conflicts with builtin modules like "time" but this is not so important

examples/client-alarms-and-conditions.py
```diff
@@ -0,0 +1,50 @@
...
+        path = ['%s:AlarmObject' % idx]
+        alarm_obj = await objects.get_child(path)
+        alarm = self.client.get_node(ua.NodeId(10637))
+
+        alarm_handle = await sub.subscribe_events(alarm_obj, alarm)
+
+
+async def start():
+    client = OpcUaClient("opc.tcp://localhost:4840")
+    await client.init()
```
oroulet: somewher you should have a Client.Disconnect(). I could not see it. The safest API is to use `async with`

examples/client-alarms-and-conditions.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+if __name__ == '__main__':
+    loop = asyncio.get_event_loop()
+    loop.create_task(start())
+    loop.run_forever()
```
oroulet: maybe just use `asyncio.run` I know it is >=3.7 but there are bugs in asyncio in 3.6 so maybe we should jsut says opcua-asyncio is >=3.7

asyncua/server/event_generator.py
mar-ar: I just took the old code because I saw that it dosnt work. I didn´t want to change something there. Dosen´t belong to my changes. I just didn´t want to look at the local time thing.

examples/server-alarms-and-conditions.py
```diff
@@ -0,0 +1,87 @@
...
+    server = OpcUaServer("opc.tcp://0.0.0.0:4840")
+    await server.init()
+    await server.server.start()
+    await interactive(server)
+
+
+if __name__ == '__main__':
+    loop = asyncio.get_event_loop()
+    loop.create_task(main())
+    loop.run_forever()
```
oroulet: same here use ` async with `and `asyncio.run`

examples/client-alarms-and-conditions.py
mar-ar: yeah you are right

examples/client-alarms-and-conditions.py
mar-ar: I just wanted to write a simple test. Didn´t thought much about it.

oroulet:
Everything I could underdtand looks very good.
the only thing I am wondering about is the removal of emmiting_node. How is this suposed to work? The emitting node option was added in   PR from https://github.com/FreeOpcUa/python-opcua/pull/760 from  @Nebukadneza and there was a reason for it. How de we specify emitting node now?
also I am wondering about that file eventobjects.py... It is huge, can't it be autogenerated? but OK this is for another PR

mar-ar:
I didn´t want to remove something there. I think that my codebase was just old. I look at it and change it.
I did autogenerate eventobjects.py. I just changed the files in schemas.

oroulet:
> I think that my codebase was just old.
OK then you need to review all changes and really only keep those that you want to merge. We cannot break things

mar-ar:
The emmiting_node still works like before.
event_objects.py is autogenerated.
I changed to async with.
Is the try except also ok? So we don´t have to rise the version.
I could need some help with the local time the changes don´t work at my system.

mar-ar:
Now everything schould be ok.
The only thing is that I also want to have more than one instance of an condtition and I at the moment I have no idea how to handle this... 
maybe someone can halp me there.
At the moment you have to use several branches for more than one instance but you have to handle these brachnes on your own.

oroulet:
Show a minimal example having that issue and explain the symptoms

mar-ar:
I just changed somthing in my example:
`# coding: utf-8

from aioconsole import ainput

import asyncio

from asyncua import Server, ua


class OpcUaServer(object):

    def __init__(self, endpoint):
        self.server = Server()
        self.server.set_endpoint(endpoint)
        self.server.set_server_name('Alarms and Conditions Test Server')
        self.server.application_type = ua.ApplicationType.Server
        self.con_gen = None
        self.alarm_gen = None

    async def __aenter__(self):
        await self.init()
        await self.server.start()
        return self

    async def __aexit__(self, exc_type, exc_value, traceback):
        await self.server.stop()

    async def init(self, shelf_file=None):
        await self.server.iserver.init(shelf_file)
        uri = "http://examples.freeopcua.github.io"
        idx = await self.server.register_namespace(uri)
        objects = self.server.get_objects_node()

        con_obj = await objects.add_object(idx, "ConditionObject")
        condition = self.server.get_node(ua.NodeId(2830))
        self.con_gen = await self.server.get_event_generator(condition, con_obj)
        self.con_gen.event.add_property('NodeId', con_obj.nodeid, ua.VariantType.NodeId)

        alarm_obj = await objects.add_object(idx, "AlarmObject")
        alarm = self.server.get_node(ua.NodeId(10637))
        self.alarm_gen = await self.server.get_event_generator(alarm, alarm_obj)

    def generate_condition(self, retain):
        self.con_gen.event.ConditionName = 'Example Condition'
        self.con_gen.event.Message = ua.LocalizedText("Some Message")
        self.con_gen.event.Severity = 500
        self.con_gen.event.BranchId = ua.NodeId(0)
        if retain == 1:
            self.con_gen.event.Retain = True
        else:
            self.con_gen.event.Retain = False
        self.con_gen.trigger()

    def generate_alarm(self, active):
        self.alarm_gen.event.ConditionName = 'Example Alarm1'
        self.alarm_gen.event.Message = ua.LocalizedText("error in module1")
        self.alarm_gen.event.Severity = 500
        self.alarm_gen.event.BranchId = ua.NodeId(0)
        self.alarm_gen.event.AckedState = ua.LocalizedText('Unacknowledged', 'en')
        setattr(self.alarm_gen.event, 'AckedState/Id', False)
        if active == 1:
            self.alarm_gen.event.Retain = True
            self.alarm_gen.event.ActiveState = ua.LocalizedText('Active', 'en')
            setattr(self.alarm_gen.event, 'ActiveState/Id', True)
        else:
            self.alarm_gen.event.Retain = False
            self.alarm_gen.event.ActiveState = ua.LocalizedText('Inactive', 'en')
            setattr(self.alarm_gen.event, 'ActiveState/Id', False)
        self.alarm_gen.trigger()

    def generate_alarm2(self, active):
        self.alarm_gen.event.ConditionName = 'Example Alarm2'
        self.alarm_gen.event.Message = ua.LocalizedText("error in module2")
        self.alarm_gen.event.Severity = 500
        self.alarm_gen.event.BranchId = ua.NodeId(0)
        self.alarm_gen.event.AckedState = ua.LocalizedText('Unacknowledged', 'en')
        setattr(self.alarm_gen.event, 'AckedState/Id', False)
        if active == 1:
            self.alarm_gen.event.Retain = True
            self.alarm_gen.event.ActiveState = ua.LocalizedText('Active', 'en')
            setattr(self.alarm_gen.event, 'ActiveState/Id', True)
        else:
            self.alarm_gen.event.Retain = False
            self.alarm_gen.event.ActiveState = ua.LocalizedText('Inactive', 'en')
            setattr(self.alarm_gen.event, 'ActiveState/Id', False)
        self.alarm_gen.trigger()


async def interactive(server):
    while True:
        # server.generate_condition(1)
        # server.generate_alarm(1)
        line = await ainput(">>> ")
        print('execute:', line)
        if line == 'exit':
            break
        try:
            eval(line)
        except Exception as msg:
            print('Exception:', msg)
            raise Exception


async def main():
    async with OpcUaServer("opc.tcp://0.0.0.0:4840") as server:
        await interactive(server)


if __name__ == '__main__':
    loop.create_task(main())
    try:
        loop.run_forever()
    finally:
        loop.run_until_complete(loop.shutdown_asyncgens())
        loop.close()`

When you have a simple event you can take the event_generator to throw as many events as you want.
But if you have a condition it is not that easy. Because you need a instance for each condition you want to throw (and retain undtil it is solved). Just use the example above and throw server.generate_alarm(1) and server.generate_alarm2(1) and you will see in Uaexper only the second condition: 
image

But I want so see both alarms.

Do you understand what I mean?

mar-ar:
#105

oroulet:
@mar-ar you should not mix up too many things in that PR. Especially since you say there is something broken ;-)
Alo I do not remember about eventing stuff works so it is hard for me to comment

mar-ar:
You are probably right. No there is nothing broken. I just had no Idea how to handle some advanced stuff. I think at the moment everything is working. 
In my opinion you can merge this like this.

oroulet:
@mar-ar anyway I cannot merge anything that breaks so many tests. there must be something wrong with your patch that breaks half of the tests

mar-ar:
You are right. I will look at the tests and fix that. I havent looked at the tests yet.

oroulet:
it might be hard, you can try to fix everything at once, but I advice you to split your changes in several chunks/PR

mar-ar:
All right I will do that.

PR #102, opened 2019-11-23
Resolve collections.Iterable DeprecationWarning for Python 3.8

chrisjbremner:
While running tests, I saw this `DeprecationWarning` come up:

```
...\opcua-asyncio\asyncua\common\subscription.py:210: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    if isinstance(nodes, collections.Iterable):
```

This change [was made in Python 3.3](https://docs.python.org/3.7/library/collections.abc.html#module-collections.abc), and since `opcua-asyncio` only support Python 3.6+, this should be a safe change.

### DIFF
--- a/asyncua/common/subscription.py
+++ b/asyncua/common/subscription.py
@@ -3,7 +3,7 @@ high level interface to subscriptions
 """
 import asyncio
 import logging
-from collections import Iterable as CollectionsIterable
+import collections.abc
 from typing import Union, List, Iterable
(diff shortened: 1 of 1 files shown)


oroulet:
Thanks. I have that one many times on other projects and did not understand how to fix it :+1:

Issue #108, opened 2019-12-03
`get_node` of `sync.Server` is broken


amonsch:
`get_node` of `asyncua.sync.Server` raises an `AttributeError` see this Stacktrace for details:

```python
File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/asyncua/sync.py", line 212, in get_node              
    return Node(self.tloop, server.Server.get_node(self, nodeid))                                                                    
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/asyncua/server/server.py", line 374, in get_node     
    return Node(self.iserver.isession, nodeid)                                                                                       
AttributeError: 'Server' object has no attribute 'iserver'                                                                           
```

amonsch:
I have created a Pull Request, see here: #109

amonsch:
PR was merged

PR #109, opened 2019-12-03
Fix Issue #108

amonsch:
Calling `sync.Server.get_node` raises an `AttributeError`.
This commit fixes the issue and adds a corresponding test.


oroulet:
Thanks, nice to see some testing of that stuff

PR #114, opened 2019-12-17
limit the max amount of connections

mar-ar:
Hey,

I think this is very useful.
I had no other idea where to solve this problem but I think this works fine.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -20,6 +20,8 @@ class InternalSession:
     """
 
     """
+    __max_connections = None
+    __current_connections = 0
     _counter = 10
     _auth_counter = 1000
 
@@ -45,6 +47,9 @@ class InternalSession:
     def __str__(self):
         return f'InternalSession(name:{self.name}, user:{self.user}, id:{self.session_id}, auth_token:{self.auth_token})'
 
+    def set_max_connections(self, quantity):
+        InternalSession.__max_connections = quantity
+
     async def get_endpoints(self, params=None, sockname=None):
         return await self.iserver.get_endpoints(params, sockname)
 
@@ -63,6 +68,9 @@ class InternalSession:
 
     async def close_session(self, delete_subs=True):
         self.logger.info('close session %s', self.name)
+        if InternalSession.__max_connections is not None:
+            if self.state == SessionState.Activated:
+                InternalSession.__current_connections -= 1
         self.state = SessionState.Closed
         await self.delete_subscriptions(self.subscriptions)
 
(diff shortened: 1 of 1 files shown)


oroulet:
the main idea of that PR is good, but there are some implementation details we need to change

asyncua/server/internal_session.py
```diff
@@ -20,6 +20,8 @@ class InternalSession:
     """
 
     """
+    __max_connections = None
+    __current_connections = 0
```
oroulet: no reasond for double_underscore here. Even no reason for underscore at all. Attributes are public in python so if a property has a value which is not too strange just let people set it directly instead of using setter and getter

asyncua/server/internal_session.py
```diff
@@ -45,6 +47,9 @@ def __init__(self, internal_server, aspace: AddressSpace, submgr: SubscriptionSe
     def __str__(self):
         return f'InternalSession(name:{self.name}, user:{self.user}, id:{self.session_id}, auth_token:{self.auth_token})'
 
+    def set_max_connections(self, quantity):
+        InternalSession.__max_connections = quantity
+
```
oroulet: comment over let us get rid of that method

asyncua/server/internal_session.py
```diff
@@ -20,6 +20,8 @@ class InternalSession:
     """
 
     """
+    __max_connections = None
```
oroulet: Probably better to set a high max number of connections, we do not support so many at once anyway, python is single process. Maybe 1000 make sense

asyncua/server/internal_session.py
```diff
@@ -63,6 +68,9 @@ def __str__(self):
 
     async def close_session(self, delete_subs=True):
         self.logger.info('close session %s', self.name)
+        if InternalSession.__max_connections is not None:
+            if self.state == SessionState.Activated:
+                InternalSession.__current_connections -= 1
```
oroulet: be carefull never to get under 0

asyncua/server/internal_session.py
```diff
@@ -71,11 +79,16 @@ def activate_session(self, params):
...
+        if InternalSession.__max_connections is not None:
+            if InternalSession.__current_connections >= InternalSession.__max_connections:
+                raise ServiceError(ua.StatusCodes.BadMaxConnectionsReached)
         self.nonce = create_nonce(32)
         result.ServerNonce = self.nonce
         for _ in params.ClientSoftwareCertificates:
             result.Results.append(ua.StatusCode())
         self.state = SessionState.Activated
+        if InternalSession.__max_connections is not None:
+            InternalSession.__current_connections += 1
```
oroulet: I am sure you can merge these two ìf`into one

asyncua/server/internal_session.py
mar-ar: Yeah I know but I still like the concept of private and protected. I think a lot of python programmers use it like this.

asyncua/server/internal_session.py
mar-ar: Sure we can set a limit. I just thought that there was no limit before so I didnt want to set one now.

asyncua/server/internal_session.py
mar-ar: Without checking the state it will because of this I check the state.

mar-ar:
I make the changes and upload it.

asyncua/server/internal_session.py
oroulet: yes you can use single underscore to protect a variable you do not want people to use, but in that case, you WANT users to know of it, so it is better to make it public and avoid setters ( I am quite sure this is written somewhere as Python best practice)

oroulet:
Looks good but it would be great to have a test for that so we know it really works. Add a test in a new files called test_connections.py for example, create one server with limited number of connections and try to connect several clients

mar-ar:
allright I write a test

PR #115, opened 2019-12-17
enable the ServerTimestamp

mar-ar:
Hey,

here I enabled the ServerTimestamp again.
I think it is useful because you know the server time of the request and then you know at least the timedelta between the writing and the request.

Or why shouldn`t it be useful?

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -2,6 +2,7 @@
 High level node object, to access node attribute
 and browse address space
 """
+from datetime import datetime
 
 import logging
 
@@ -171,7 +172,9 @@ class Node:
         An exception will be generated for other node types.
         DataValue contain a variable value as a variant as well as server and source timestamps
         """
-        return await self.get_attribute(ua.AttributeIds.Value)
+        datavalue = await self.get_attribute(ua.AttributeIds.Value)
+        datavalue.ServerTimestamp = datetime.utcnow()
+        return datavalue
 
     async def set_array_dimensions(self, value):
         """
--- a/asyncua/common/ua_utils.py
+++ b/asyncua/common/ua_utils.py
@@ -23,9 +23,11 @@ def value_to_datavalue(val, varianttype=None):
     elif isinstance(val, ua.Variant):
         datavalue = ua.DataValue(val)
         datavalue.SourceTimestamp = datetime.utcnow()
+        datavalue.ServerTimestamp = datetime.utcnow()
     else:
         datavalue = ua.DataValue(ua.Variant(val, varianttype))
         datavalue.SourceTimestamp = datetime.utcnow()
+        datavalue.ServerTimestamp = datetime.utcnow()
     return datavalue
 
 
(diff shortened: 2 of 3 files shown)


asyncua/common/node.py
```diff
@@ -171,7 +172,9 @@ def __hash__(self):
         An exception will be generated for other node types.
         DataValue contain a variable value as a variant as well as server and source timestamps
         """
-        return await self.get_attribute(ua.AttributeIds.Value)
+        datavalue = await self.get_attribute(ua.AttributeIds.Value)
+        datavalue.ServerTimestamp = datetime.utcnow()
+        return datavalue
```
oroulet: this does not make sense. you are reading an attribute from server and overwritting its timestamp??

asyncua/common/ua_utils.py
```diff
@@ -23,9 +23,11 @@ def value_to_datavalue(val, varianttype=None):
     elif isinstance(val, ua.Variant):
         datavalue = ua.DataValue(val)
         datavalue.SourceTimestamp = datetime.utcnow()
+        datavalue.ServerTimestamp = datetime.utcnow()
     else:
         datavalue = ua.DataValue(ua.Variant(val, varianttype))
         datavalue.SourceTimestamp = datetime.utcnow()
+        datavalue.ServerTimestamp = datetime.utcnow()
```
oroulet: same here. you never want to overwrite value from server

mar-ar:
I just realised that thing when I checked a server with the unified automation SDK.
There the SourceTimestamp is the Timestamp when you wrote the variable the last time.
The ServerTimestamp is always the timestamp when you requested the variable to read.
I will check the specifications if this is corecct.
And if its correct my implementation should be right.

mar-ar:
> The serverTimestamp is used to reflect the time that the Server received a Variable value or knew it to be accurate.

I think then it is correct to overwrite it each time you read it.

oroulet:
The server trimestamp should be updated inside address space when it receives new data. not when reading it. We had that behaviour but I think it was disabled due to performance issue (The on line commented you re-enabled) . So enabling it can be discussed but we should not overwrite it at other places

mar-ar:
No I dont think so. Whats the difference to the SourceTimestamp then?
When you read the variable and it is the last time you knew the value to be accurate.

mar-ar:
image

image

oroulet:
Source is from the underlying system (the sensor, whaterver) and the server is the time the server receive the data from the source

mar-ar:
Hmm idk. I try to explain how I understand it.

SourceTimestamp:

> The sourceTimestamp shall be UTC time and should indicate the time of the last change of the value or statusCode.

Last change of the datasource.

ServerTimestamp:

> The serverTimestamp is used to reflect the time that the Server received a Variable value or knew it to be accurate.

The last time you read the variable you know that it is acurate.

And I know that behavior from an UnifiedAutomation Server.

So I still thik that its right to set the ServerTimestamp when you read the variable. And only set the SourceTimestamp when you write it.

oroulet:
My current understanding is that your interpretation is wrong. ServerTimestamp looks like it has been updated to the last time you read it, but it is not. What most probably happens is that the sync between the underlying system and the opcua server is not based on event but on some brain dead forwarding of data from sensor to server, Thus, ServerTimestamp is constantly updated to the current time (== the time the server received it == the time you read it).
I may be wrong, but I would be surprised.

mar-ar:
Hmm okey you are probably right.
So do we close it or reenable the one line?

PR #118, opened 2019-12-19
Don't cancel missing subscription task

chrisjbremner:
In https://github.com/FreeOpcUa/opcua-asyncio/pull/94, I made it so that you could have a 0ms publishing interval on your subscription. When you do this, the [`InternalSubscription` doesn't create a subscription task](https://github.com/FreeOpcUa/opcua-asyncio/blob/08861799d3413345dcc5b4ae387427739b864bad/asyncua/server/internal_subscription.py#L50-L51). Then, when the `InternalSubscription` is stopped, the `stop` function attempts to cancel the `InternalSubscription`s task, which doesn't exist, leading to an `AttributeError`.

This PR introduces a simple check to see if the task exists before cancelling it, and do nothing if the task doesn't exist.

This was tested with both a server and a client from this package. I subscribed to the server (with a 0ms subscription interval) with the client, then confirmed that the `AttributeError` was not raised when the subscription was stopped.

### DIFF
--- a/asyncua/server/internal_subscription.py
+++ b/asyncua/server/internal_subscription.py
@@ -51,10 +51,13 @@ class InternalSubscription:
             self._task = self.loop.create_task(self._subscription_loop())
 
     async def stop(self):
-        self.logger.info("stopping internal subscription %s", self.data.SubscriptionId)
-        self._task.cancel()
-        await self._task
-        self._task = None
+        if self._task:
+            self.logger.info("stopping internal subscription %s", self.data.SubscriptionId)
+            self._task.cancel()
+            await self._task
+            self._task = None
+        else:
+            self.logger.debug(f"internal subscription has no task to stop")
         self.monitored_item_srv.delete_all_monitored_items()
 
     def _trigger_publish(self):
--- a/asyncua/server/internal_subscription.py
+++ b/asyncua/server/internal_subscription.py
@@ -56,8 +56,6 @@ class InternalSubscription:
             self._task.cancel()
             await self._task
             self._task = None
-        else:
-            self.logger.debug(f"internal subscription has no task to stop")
         self.monitored_item_srv.delete_all_monitored_items()
 
     def _trigger_publish(self):


oroulet:
Hi, there are too many commits and branch issue in that PR. rebase on master then quash everything into one commit. Then fix the small comment

asyncua/server/internal_subscription.py
```diff
@@ -51,10 +51,13 @@ def __str__(self):
...
-        self._task.cancel()
-        await self._task
-        self._task = None
+        if self._task:
+            self.logger.info("stopping internal subscription %s", self.data.SubscriptionId)
+            self._task.cancel()
+            await self._task
+            self._task = None
+        else:
+            self.logger.debug(f"internal subscription has no task to stop")
```
oroulet: I propose to change this code to:

```
if not self.task:
    warning
    return
do your normal stuff here
```

Maybe this is just a habit but I think it is more readble and put all special case code at one place

oroulet: this is also how most of the library is/should be written

chrisjbremner: I this case, though, presumably we want to call `self.monitored_item_srv.delete_all_monitored_items()`, regardless of whether there's a task or not, right?

chrisjbremner:
I rebased and force-pushed and the commits that are already in the `FreeOpcUa:master` branch fell off, leaving just my one commit.

asyncua/server/internal_subscription.py
oroulet: I do not remember the code ... I suppose we do..

cbergmiller: Yes @chrisjbremner , `delete_all_monitored_items` should be called when there is no task (which means "zero delay" publishing).

chrisjbremner: Sounds like what I have is fine then?

oroulet: OK.
But do we need that debug logging? we have already too much debug logging. It is almost impossible to use it anyway

chrisjbremner: Ok, removed.

PR #120, opened 2019-12-20
replace set and get_values method with read and write to get closer t…

oroulet:
…o spec


oroulet:
@cbergmiller opinion on that one? maybe read_value and write_value is more explicit and does not collide with UaClient.read and write but I thought these methods were used so often they had to be short..

cbergmiller:
I would prefer the more explicit `Node.read_value`. On the other hand we can think of `Node.read` as a shorthand version of `read_value` so i don't see a problem.

oroulet:
@cbergmiller I changed to read_value and write_value. I think it is better to be explicit

PR #124, opened 2020-01-07
Limit connections added test

mar-ar:
here is the promised test


tests/test_connections.py
```diff
@@ -0,0 +1,32 @@
...
+
+async def test_max_connections_1(opc):
+    opc.server.iserver.isession.__class__.max_connections = 1
+    port = opc.server.endpoint.port
+    if port == port_num:
+        # if client we already have one connection
+        client_2 = Client(f'opc.tcp://127.0.0.1:{port}')
+        with pytest.raises(BadMaxConnectionsReached):
+            await client_2.connect()
+            await client_2.disconnect()
```
oroulet: disconnect will never be called.
The safest in that case is probably to use with:
```
with  pytest.raises(zxxx):
     with c2 as Client(xxx):
          pass
```

tests/test_connections.py
```diff
@@ -0,0 +1,32 @@
...
+    else:
+        client_1 = Client(f'opc.tcp://127.0.0.1:{port}')
+        await client_1.connect()
+        client_2 = Client(f'opc.tcp://127.0.0.1:{port}')
+        with pytest.raises(BadMaxConnectionsReached):
+            await client_2.connect()
+            await client_2.disconnect()
```
oroulet: same here

oroulet:
@mar-ar looks good! but you need to rebase on master

oroulet:
@mar-ar now it is even worse. there are 18 unnecessary commits  and still  not based on master...

you should have usedonly one command: "git  pull --rebase"but now I  am not sure where you are. 
mayve best to make a new branch  from master and cherry pick the one commit  you want

mar-ar:
Sorry I am used to Mercurial and I think its much more convenient ...
I am not into git yet.
I make it new.

oroulet:
I made themove bzr to git several years ago and I do  no regret it. git is sometimes confisuing but allow many more usefull  things

PR #125, opened 2020-01-07
Add a few methods to sync Node API

chrisjbremner:
I was starting to use the sync API to replace `python-opcua` in an application, but came across a few incompatibilities due to some missing methods in the sync API that did exist in `ua.Node`. I added these methods back to the API and my application is working equivalently to when it was using `python-opcua`. 

While adding these methods to the API, I also noticed that `read_value` had the wrong arguments (it took an extra positional argument that it shouldn't have), so I fixed that as well.

Is the goal to eventually expose the entire API for each object in this sync file?

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -248,12 +248,19 @@ class Node:
         self.aio_obj = aio_node
         self.tloop = tloop
 
-    def __hash__(self):
-        return self.aio_obj.__hash__()
+    def __eq__(self, other):
+        return self.aio_obj == other.aio_obj
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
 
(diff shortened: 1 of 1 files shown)


oroulet:
Yes the goal is to expose everything necessary. As far as I remember most things are exposed automatically. Nice to hear things are working for you. Let us know if you see any differences

chrisjbremner:
Thanks for the quick review! Do you anticipate a release any time soon, or should I use a release from my fork for my purposes?

oroulet:
I can release whenever you want. It should be stable

chrisjbremner:
Alright, in that case, could I request a release? Thanks

oroulet:
done

PR #128, opened 2020-01-12
stream

alisonsalmeida:
Adding the client session history function arguments to better control data packets in a streamed data transmission to one or more optional clients.
The biggest reason was the need that I am having to deploy opcua servers that store data every 1 second and can request the track record of at least one year which will add up to a total of 31.104.000 samples.

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -517,10 +517,10 @@ class Node:
             details.EndTime = ua.get_win_epoch()
         details.NumValuesPerNode = numvalues
         details.ReturnBounds = True
-        result = await self.history_read(details)
+        result = await self.history_read(details, None)
         return result.HistoryData.DataValues
 
-    async def history_read(self, details):
+    async def history_read(self, details, session):
         """
         Read raw history of a node, low-level function
         result code from server is checked and an exception is raised in case of error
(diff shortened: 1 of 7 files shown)


oroulet:
Thanks for the patch. Great with the use of about DB. But how do you use that session arg? It is not used anywhere in examples and I am afraid it will be deleted by mistakes if not used

alisonsalmeida:
Thanks, this is my first contribution to a repository. The session variable is for the server to control data pagination for each connected client, so the server has greater control of which clients are requesting data. My demand is that I have to feed several html pages with the opcua server information that stores the data every second in the bank; The http client might want to request data for at least 30 days, so it's about 2 million samples, so I had the idea to page this data. And since each request can come from different opcua clients, a variable showing which client is requesting data, I think, would be interesting.

alisonsalmeida:
I wrote an example of what I am proposing. But if you have a better solution, I'm open to suggestions

oroulet:
I may misunderstand but this looks strange. What do you call 'page', do you mean 'caching'? 
Caching is already implemented inside DB (and much more efficient than anything you can do in python) so you should not do that in asyncua.

oroulet:
no anser from author, closing

PR #136, opened 2020-01-28
[OpcuaClient] Default ClientSignature.Algorithm reference

ZuZuD:
Default ClientSignature.Algorithm reference to AsymmetricSignatureURI

Similar to https://github.com/FreeOpcUa/python-opcua/pull/945
Small test: https://pastebin.com/SxUSnFi0


oroulet:
thanks

PR #138, opened 2020-01-31
Make value_callback work with coroutines

swamper123:
Add Method to add Coroutines as value_callback
Make all necessary Methods async.


swamper123:
Server-example is running without problems. At the moment I don't know if there are some tests don't support async or it is my code. If someone may have a look on it, it would be great.

oroulet:
Tests tail and there are many obvious errors messages like 'coroutine xxx not awaited' so try to fix that first. But did you do all these changes just too have this value_callback be a coroutines? It might be ok. We just need to make sure it fits not have a big performance issue

oroulet:
Also I did not know someone was using that stuff. Why do you just it?

oroulet:
Also this could have been done in a sync way by pushing the coro to the loop and waiting for its future in a sync method. But once again I am unsure about your goal

cbergmiller:
Hi @swamper123, thanks for your interest and contribution to opcua-asyncio!

I wasn't event aware of the `AttributeValue.value_callback` method. Could you tell us more about your use case? I would advise to go for a server architecture that can also support subscriptions. Could you subscribe/watch your data source to update the attribute values in address space rather than resolving the attribute value asynchronously? We would be happy to advise you in the implementation of an alternative approach.

At the moment i am not sure if the benefits from your PR  are worth the added complexity.
@oroulet maybe we should we deprecate `AttributeValue.value_callback` and issue a warning?

oroulet:
@cbergmiller does that PR add complexity? It just makes many methods async. so I am more afraid of possible performance issues

and yes the AttributeValue.value_callback is not very well though.I remember I made it in C++ so I also put in python,but I have never really see a good use case for it. I breaks subscriptions

cbergmiller:
@oroulet you are right about possible performance issues rather than added complexity.

swamper123:
@oroulet: The Problem is, that this feature triggers a rattail, so you have to start async everything up to the top. When you want anything to run with coroutines, you have to use an `await`. If there is an `await`, you need a `async def`. If the function/methods has an `async def`, you need to use `await` everywhere it is used. If there is an `await` .... and so on. But now to my usecase:

@cbergmiller My build up is a ("stupid-non OPC able") PLC and OPC UA server and I use a script to let them communicate (using asyncio). Instead of requesting all PLC-Values every xxx ms/s, I want to trigger a request process. This is where` value_callback` is used. If somebody wants to read the `Value` of a `Node` , the async request process starts.
Now the problem I have: If I would start a `create_task` or `ensure_future` and throw this thing into a loop, (I believe that) I can't be sure, if the value in `Value` is the new requested one . The coro could still run, but the old `Value` attribute is illegitimate (and maybe taken) and what (or better how) shall I tell my client now what the new result of the reading process is?

I also got nodes which need a async func/method (between server and PLC) to even get a legit value. E.g. the temperature values in the PLC are in °C. Now I got some Nodes with Kelvin and Fahrenheit as well. If I want to read my Node with the °C value, I want to trigger the async value_callback, request, grab the values, calculate the stuff in Kelvin, Fahrenheit and Celsius and put the new values into their nodes as well.
So my thought was, if it would be possible to add coros into the value_callback, I wouldn't have problems with my async communication Script, I could run even time intense async processes without (greater) performance loss incl. other coros timing out and it would be a good feature beside the sync value_callbacks as well.

oroulet:
@swamper123 the technical solution is probably something like:

>> The coro could still run, but the old Value attribute is illegitimate (and maybe taken) and what (or better how) shall I tell my client now what the new result of the reading process is?

`     futur = asyncio.run_coroutine_threadsafe(coro, loop=self.loop)
        return futur.result()  # blocks until the result has arrived
`
This is what I use in sync wrapper. But it would block the entire program until your network request  to your PLC has arrived. so in that sense your solution is better

>> Instead of requesting all PLC-Values every xxx ms/s, I want to trigger a request process.

yes this is why I made that callback stuff. But later I found out that really everything in that PLC/OPC world is based on the idea of polling every xxx ms. It is not that bad. you can have a thread doing this polling a given frequency

Also one more think is that we should make sure we do not have a big race condition here. What happens  if someone deletes that node in the meantime? Is there other things that might happen breaking things @cbergmiller ?

swamper123:
@oroulet Even if I would poll all values from my PLC, I would still have problems with values, which have to be calculated (partly time intensive) by the PLC. I have to start the calculation until somebody wants to have something, which would better be async.
To the reace condition point, wouldn't it throw an error, because the used reference is not used by the server (which could cause other problems, if a node is deleted and another one is created with the same stuff)?
Or do you mean the whole rest which may be asynced?

oroulet:
>> To the reace condition point, wouldn't it throw an error, because the used reference is not used by the server 

The entire point with race conditions is that strange things might happen and we do not have control

swamper123:
> > > To the reace condition point, wouldn't it throw an error, because the used reference is not used by the server
> 
> The entire point with race conditions is that strange things might happen and we do not have control

What race conditions explicit  do you mean? Race conditions caused by using `value_callback`  (async) funcs/methods itself? Race conditions caused by extending more stuff with async (without  the value_callback stuff)? Race conditions in my explicit usecase?

oroulet:
since we are awaiting something while reading, other tasks may do something in the meantime. Like for example deleting that node. If this is the only possibility we need to add a lock there. But maybe other stuff can also happen, not sure, it requires to check all data structures we are accessing during a read operation and making sure nobody can modify them  in the meantime

swamper123:
I got a couple of ideas about that:
- One idea is to add an Attribute `value_callback_handler` to `AttributeValue`, where the handler of the `value_callback` coro could be stored. With that, the `_delete_node` method in address_space.py could be extended by a request,if a callback task exist. If yes, cancel it  (maybe with a logger_warning), and continue the deletion process. Canceling a task, if the handler is available, would be sync as well.
- Another Idea would be, if a `value_callback`task is running and the Node shall be deleted, to use `asyncio.add_done_callback` to let the read-process finish and start the deletion process of the node directly afterwards.
- Another Idea is to add a flag "deletion_progress" and to deny further operations with the Node, until all tasks are done.

oroulet:
No please. Do not fix up different things in one PR. It is already complicated enough with only the callback issue. Make another PR for the address space update

oroulet:
And clean up all these merge conflicts

swamper123:
Clean up in new pull request.

PR #145, opened 2020-02-14
Added missing methods to sync client:

TilmanK:
- Client.set_security
- Client.connect_and_get_server_endpoints
- Node.get_path
- Node.get_node_class
- Node.get_attributes


Ilyazyk:
Could you please add other missing methods, at least: write_values(), read_values().

Issue #146, opened 2020-02-17
Exception when trying to disconnect from the server with sync client.


TilmanK:
The disconnect method of the sync Client is obvioulsy broken. When calling disconnect, I'm getting the following exception:

2020-02-17 10:00:17.626  Thread-10  -                    client - INFO     - disconnect
2020-02-17 10:00:17.626  Thread-10  -                 ua_client - INFO     - Request to close socket received
2020-02-17 10:00:17.626  Thread-10  -                 ua_client - INFO     - Socket has closed connection
Traceback (most recent call last):
  File "C:\Users\krumm-ti\PycharmProjects\LabMonitor\views\mainwindow.py", line 187, in closeEvent
    OpcClient.inst().disconnect()
  File "C:\Users\krumm-ti\PycharmProjects\LabMonitor\opc_client.py", line 170, in disconnect
    self._client.disconnect()
  File "C:\Users\krumm-ti\PycharmProjects\LabMonitor\venv\lib\site-packages\asyncua\sync.py", line 116, in disconnect
    self.tloop.post(self.aio_obj.disconnect())
  File "C:\Users\krumm-ti\PycharmProjects\LabMonitor\venv\lib\site-packages\asyncua\sync.py", line 49, in post
    return futur.result()
  File "C:\Users\krumm-ti\AppData\Local\Programs\Python\Python37-32\lib\concurrent\futures\_base.py", line 423, in result
    raise CancelledError()
concurrent.futures._base.CancelledError

Minimal example:

`from asyncua.sync import Client

client = Client( "your uri here")
client.connect()
client.disconnect()`

Can you please check what causes this? I wasn't able to figure it out by myself.

oroulet:
In fact it looks ok. To stop tasks a cancel exception is raised. But maybe we should catch it so users do not get confused

TilmanK:
Ah ok. Yes, indeed we should to micmic the behaviour of the old sync client. Catching the exception works as expected and the client disconnects.

I'll submit a PR tonight if you want.

oroulet:
My problem is that I am not really sure where it should be catched. The doc sais one should not catch CancelledError. This is true since it might make tasks not killable. But maybe inside disconnect is fine...

TilmanK:
Yes, exactly in the disconnect method of the sync Client. I'd expect the method to behave like the one from opcua sync lib and any difference will make migrations more complicated. 

And the exception doesn't make any sense when using sync IMHO.

[...]

PR #147, opened 2020-02-18
Fixed exception on disconnect call of sync client. #146

TilmanK:


oroulet:
Probably ok. But I do not think we should have a del method. One may want to have several clients using same loop

TilmanK:
The loop is stopped in the __del__ method only if it was created in the __init__ method. AFAIK the __del__ method is exactly the place to handle such things created in __init__.

If a user wants to use the same loop on several Clients, he can pass that loop to the Client's constructor. In this case, the loop isn't stopped.

That's the intention I had when submitting the PR ;)

oroulet:
OK. sorry I forgot that implicit loop creating existed... I am not sure it is a good idea.. but OK we have it

TilmanK:
The implicit generation is more or less quite elegant. Just my 2 cents...

PR #148, opened 2020-02-18
Fix AttributeError when historizing events

chrisjbremner:
I noticed a small bug when trying to historize an event with a `period` value not set to the default value of `None`. While `DataValue` objects store their timestamp in the `SourceTimestamp` attribute, `Event` objects instead use the `Time` attribute. It looks like there was some code for `DataValue`s that was re-used for `Event`s, and this attribute wasn't changed, causing this error:

```python
  File "asyncua\server\history.py", line 152, in save_event
    while len(evts) and now - evts[0].SourceTimestamp > period:
AttributeError: 'Event' object has no attribute 'SourceTimestamp'
```

This was tested by setting a short value for `period`, triggering an event, then triggering another event after the time specified in `period`, and confirming that the first event was removed from the history.

### DIFF
--- a/asyncua/server/history.py
+++ b/asyncua/server/history.py
@@ -149,7 +149,7 @@ class HistoryDict(HistoryStorageInterface):
         period, count = self._events_periods[event.emitting_node]
         now = datetime.utcnow()
         if period:
-            while len(evts) and now - evts[0].SourceTimestamp > period:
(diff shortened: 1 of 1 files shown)


oroulet:
I do not use history so I just trust you ;-) . Anyway it makes sense

PR #153, opened 2020-02-25
Message padding for keys larger than 2048 bit

miomad:
Changed message padding for a big private key as in https://reference.opcfoundation.org/v104/Core/docs/Part6/6.7.2/#6.7.2.5

Added certificates for test_crypto_connect


miomad:
Hi!
The same changes can be applied to the python-opcua. 
Do you backporting common changes to  the python-opcua? Or should I create pull request there?

oroulet:
Great if you can do it! Thanks

PR #154, opened 2020-02-25
Don't remove previous security token when revolving tokens

chrisjbremner:
In https://github.com/FreeOpcUa/python-opcua/pull/915, @jonasgreen88 introduced logic to revolve security tokens whenever a channel is renewed.

In my application, I had a client that used `opcua==0.98.5` and I recently switched to `asyncua==0.8.0`. In that switch, I started seeing an influx of errors that all said `Invalid security token id {n}, expected {n+1} or {0}`. In each case, the received token was always one less than the expected token.

After looking through the change where this was introduced, it seems like whenever the tokens were revolved, [the previous token was removed](https://github.com/FreeOpcUa/opcua-asyncio/compare/master...chrisjbremner:fix_revolve?expand=1#diff-b15c66d6385a64ef5957e8a8e40e354bL280), so that if you then received an "old" token after revolving, this error would be triggered, even if the "old" token was within the 25% expiration window. I didn't see anything in the spec (mainly looked in part 4, 5.5.2.1) that says that "old" tokens should be rejected after a new token is received, so I removed the logic that removes these old tokens prematurely.

In addition, this PR:
* Cleans up some of the conditional logic to un-nest things and make it slightly easier to read
* Applies [`black`](https://github.com/psf/black) to the modified logic with a line length of 99
* Moves `make_remote_symmetric_key` into `revolve_tokens` to simplify conditional logic

I'd love some feedback from @jonasgreen88 as to whether he thinks any of these changes could be detrimental. I have been able to test with a client and a server both using this code, and haven't encountered any issues.

### DIFF
--- a/asyncua/common/connection.py
+++ b/asyncua/common/connection.py
@@ -233,7 +233,6 @@ class SecureConnection:
         self.security_token = self.next_security_token
         self.next_security_token = ua.ChannelSecurityToken()
         self.security_policy.make_local_symmetric_key(self.remote_nonce, self.local_nonce)
-        self.security_policy.make_remote_symmetric_key(self.local_nonce, self.remote_nonce)
 
     def message_to_binary(self, message, message_type=ua.MessageType.SecureMessage, request_id=0):
         """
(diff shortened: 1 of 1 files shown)


oroulet:
I do not know that part of the code but that looks good for me. I am not really keen in low line length and often use 120, but maybe 99 is ok (I often use yapf and not black)

chrisjbremner:
No problem, I changed it to 120in https://github.com/FreeOpcUa/opcua-asyncio/pull/154/commits/dfd5751274b2c5cb269bb8846965db96fbb57bf0

chrisjbremner:
For further context on why this was an issue for me (in case it also becomes an issue for someone else), I was running a server using an older version of `opcua` (`0.98.5`) that did not have the security token rotation that was implemented in https://github.com/FreeOpcUa/python-opcua/pull/915. Then I was using an `asyncua` (`0.8.0`) client running the change that was implemented in https://github.com/FreeOpcUa/opcua-asyncio/commit/c51b2c73f37ea14ea0d9e3d45e1d2bd3a8a22f2d.

In the old server code, whenever the server sent a `PublishResponse`, it used the security token it received from the `PublishRequest`. If the `PublishRequest` and `PublishResponse` had a significant time gap between which the channel was renewed, the `PublishResponse` would arrive at the client with an outdated token.

Would it make sense to add an argument that could be passed into the `SecureConnection` to disable this token check in scenarios where you are using an older version of the server?

PR #157, opened 2020-02-27
added is_connected() method & PEP-8 asyncua/client/client.py

### LINKED ISSUE #155: How can I know if the connection is still active?
Is there a way to know if the connection is still active?

I tried with `send_hello` but when I send that message, the connection drops after a while. (wondering why)

Is there any way? will make sense to have that functionality?

luisfmelo:
Close #155 

Changes:
- added is_connected() method
- PEP-8 asyncua/client/client.py

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -4,16 +4,16 @@ from typing import Union, Coroutine
 from urllib.parse import urlparse
 
 from asyncua import ua
-from .ua_client import UaClient
-from ..common.xmlimporter import XmlImporter
-from ..common.xmlexporter import XmlExporter
-from ..common.node import Node
+from .ua_client import UaClient, UASocketProtocol
 from ..common.manage_nodes import delete_nodes
-from ..common.subscription import Subscription
+from ..common.node import Node
 from ..common.shortcuts import Shortcuts
 from ..common.structures import load_type_definitions, load_enums
-from ..common.utils import create_nonce
+from ..common.subscription import Subscription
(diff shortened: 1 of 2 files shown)


asyncua/client/client.py
```diff
@@ -620,5 +624,11 @@ def load_enums(self) -> Coroutine:
         for result in results:
             result.check()
 
+    async def is_connected(self):
+        """
+        Get OPC-UA client connection status.
+        """
+        return self.uaclient.protocol.state == UASocketProtocol.OPEN
```
oroulet: this might work... but does it always work? I man can't we get errors like "protocol has not attrtbute called state",etc...

chrisjbremner: I also tried this in my own local implementation of `is_connected` and had trouble getting it to be robust. For example, if I disconnected network, the `connection_lost` callback didn't necessarily get called immediately, and the state still remained `OPEN`. I settled on querying the status node of the server as it seemed to be the most reliable method, but would be interested in trying to get this method to be robust.

oroulet: Also from what I remember looking at how prosys and unified automation clients work. They seem to be polling at quite high frequency. Reading the server node once a second or similar. So this seems to be the common way to do it

luisfmelo: What if we call `get_namespaces()`? I know that it is not the cleanest way to do it but it is an option

oroulet: Why not reading the server state node? If this is what all others do.. Then you can also check what is the state of server

PR #158, opened 2020-03-06
Use UASocketProtocol.timeout as default in send_request

miomad:
I use asyncua as a client to connect with a pretty slow server (probably it's all because of hardware)

But UaClient raises TimeoutError after 10 seconds regardless a timeout it's initialized with.

So it's a simple fix of the hardcoded timeout.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -129,12 +129,13 @@ class UASocketProtocol(asyncio.Protocol):
         self.transport.write(msg)
         return future
 
-    async def send_request(self, request, timeout=10, message_type=ua.MessageType.SecureMessage):
+    async def send_request(self, request, timeout=None, message_type=ua.MessageType.SecureMessage):
         """
         Send a request to the server.
         Timeout is the timeout written in ua header.
         Returns response object if no callback is provided.
         """
+        timeout = self.timeout if timeout is None else timeout
         data = await asyncio.wait_for(
             self._send_request(request, timeout, message_type),
             timeout if timeout else None


oroulet:
I had to check that stuff 2 times,. but yes it make sense. It will make the server cancel requests after timeout seconds. It can be the same as self.timeout

asyncua/client/ua_client.py
```diff
@@ -129,12 +129,13 @@ def _send_request(self, request, timeout=1, message_type=ua.MessageType.SecureMe
...
         return future
 
-    async def send_request(self, request, timeout=10, message_type=ua.MessageType.SecureMessage):
+    async def send_request(self, request, timeout=None, message_type=ua.MessageType.SecureMessage):
         """
         Send a request to the server.
         Timeout is the timeout written in ua header.
         Returns response object if no callback is provided.
         """
+        timeout = self.timeout if timeout is None else timeout
```
chrisjbremner: Alternatively (and more readable): `timeout = timeout or self.timeout`

miomad: I disagree. It will ignore `timeout=0` which is valid value meaning no timeout. Used in the `UaClient.publish()` method

I think this is better (but request has been already merged):
```suggestion
        if timeout is None:
            timeout = self.timeout 
```

chrisjbremner: Ah true ok, didn't realize `0` was valid.

oroulet: We can change it if you have a better solution.

miomad: The commit is fine for me. but if you think it's lack in readability, please change it to more clear.

PR #169, opened 2020-03-23
Fix issues with disconnect

TilmanK:
This PR addresses two issues:
1. If close_session is called on a Client, that is already disconnected we'll just return to prevent an exception. This is a use case when the Client loses the connection and therefore disconnects the socket. One would need to call disconnect on the Client to ensure, that i.e. the renew_channel_loop is stopped correctly.
2. Fix an exception that occurs, when the UaClient tries to disconnect its socket, but has not protocol.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -265,6 +265,9 @@ class UaClient:
         if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
             self.logger.warning("disconnect_socket was called but connection is closed")
             return None
+        if not self.protocol:
+            self.logger.warning("disconnect was called while connect is in progress")
+            return None
         return self.protocol.disconnect_socket()
 
     async def send_hello(self, url, max_messagesize=0, max_chunkcount=0):
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from typing import Union, Coroutine
+from typing import Union, Coroutine, Optional
 from urllib.parse import urlparse
 
 from asyncua import ua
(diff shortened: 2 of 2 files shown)


asyncua/client/ua_client.py
```diff
@@ -265,6 +265,9 @@ def disconnect_socket(self):
         if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
             self.logger.warning("disconnect_socket was called but connection is closed")
             return None
+        if not self.protocol:
+            self.logger.warning("disconnect was called while connect is in progress")
+            return None
```
oroulet: And what happens then? If do as disconnect is called too early then we end up connecting and this is going to be an even strsnger error. Isn'it it?

TilmanK: Sorry for the late reply.

The use-case is pretty simple (talking from the sync Client perspective): Create a client that isn't able to connect (works for me simply to disconnect my PC), and after connect was called, simply call disconnect (in my use case to shut down my application).

This results in UaClient.connect_socket() and then UASocketProtocol to never be created. Therefore UaClient.protocol is None.

I understand, that this is a maybe dangerous change, but the alternatives are worse IMHO:

1. I do not call disconnect at all, this will result in my application to never be terminated correctly once connected since there is the ghosting Thread of the opcua lib. I cannot terminate it correctly from the outside.

2. I react on the exception in my own code, by simply catching it. This will result in the same behavior, except that we might miss some follow-up code (that isn't the case now, but might in the future).

3. I remember if I have called connect but the connection has never been made. In this cases I do not call the disconnect method. From what I know the library is doing reconnects on its own hence this will probably be unreliable.

What do you think?

PR #172, opened 2020-04-21
Add indexrange kwarg to Node.set_attribute

stevedh:
This allows for a workaround  for writing (relates to #711) to allow clients to write partial arrays.

I think this seems like a reasonable incremental addition but I am not sure if there is something I am missing here... in any case it does with with the OPC servers I have tested with.

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -252,16 +252,18 @@ class Node:
         """
         return self.set_writable(False)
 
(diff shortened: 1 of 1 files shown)


oroulet:
I hate adding new arguments to these methods. But if DataValue did not have that parameter we have no choices.
But write in comment how the string should look like

oroulet:
@stevedh do you?

PR #173, opened 2020-04-22
WIP: attempt to update to last spec, broken, importer needs to be fixed

oroulet:
Help welcome. We need to fix parser and I have no time currently

the issue (as far as I remember) is that the xml now looks like this.

```
<ArrayDimensions>
    <UInt32>0</UInt32>
  </ArrayDimensions>
```

While before itwas
`<ArrayDimensions>0</ArrayDimensions>`

So the current code is trying to create an object of type UInt32 which does not exist as this should be registered as a number and just write code like : `obj.ArrayDimentsions = 0`  (you get an attribute error since the code is written similar to: `gettatr(ua, myobj)`)
This is touchy to fix since the syntax without the datatype is still used at several other places. 

Also to complicate things, the entire xml address space is now only one file while before it was several small files. This is a big issue on machine with little memory..


oroulet:
fixed in another PR

PR #181, opened 2020-05-07
Add async and await to get_namespace_array methods

### LINKED ISSUE #180: get_namespace_array is not async
Hello together,
just found an Issue here:

https://github.com/FreeOpcUa/opcua-asyncio/blob/5132c4edee2ae8348e5d960a2f890341b899d916/asyncua/server/server.py#L419

I guess `get_namespace_array` shall be a `async def` and return the value of the NamespaceArray Node.

swamper123:
Add async and await to get_namespace_array methods
Add /venv to gitignore
Will fix #180

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -528,7 +528,7 @@ class Client:
         await subscription.init()
         return subscription
 
-    async def get_namespace_array(self) -> Coroutine:
+    async def get_namespace_array(self):
         ns_node = self.get_node(ua.NodeId(ua.ObjectIds.Server_NamespaceArray))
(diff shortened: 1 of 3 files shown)


asyncua/client/client.py
```diff
@@ -528,9 +528,9 @@ async def create_subscription(self, period, handler):
         await subscription.init()
         return subscription
 
-    def get_namespace_array(self) -> Coroutine:
+    async def get_namespace_array(self) -> Coroutine:
```
oroulet, line 531: Remove the coroutine typing comment. It is better to write List if we want to document type.although this is obvious in that case...

Issue #82, opened 2019-09-23
Certificate handling


mitchpi:
Hi Guys, thanks for a great package!

I'm struggling a little with the certificate handling. The server always accepts self signed certificates, shouldn't the certificates have to be moved to a trust list prior to it being trusted? And hence the client will be rejected and not able to view the server nodes?

Cheers. Mitch.

oroulet:
It should... Patch welcome ;-)

mitchpi:
Thanks for the prompt response! I'll investigate, may be a bit beyond me.

JoeyFaulkner:
Hi there! 

also echo mitchpi's sentiment: this is an awesome package! I would like to work on this certificate handling. Do you have any tips on where this needs to be implemented and if there's anything I need to be careful of?

JoeyFaulkner:
@oroulet i think we can close this issue now because of the closed PR?

PR #182, opened 2020-05-07
simple certificate handling

JoeyFaulkner:
As mentioned in #82 , it would be useful to have a way to manage which certificates were allowed to access the OPC UA server. 

In this pull request we create a new object crypto.certificate_handler.CertificateHandler, which simply presides over a list of accepted certificates, and at connection time checks whether the certificate is in this handler, after the standard check for the certificate's validity.

Also the examples/server-with-encryption.py has been updated to be async and is now working.

Also apparently my IDE corrected some files to be pep8 compliant :smile: .

Please let me know if any major or minor changes need to be implemented, I am new to this codebase.


examples/server-with-encryption.py
```diff
@@ -1,42 +1,55 @@
...
-    myvar = myobj.add_variable(idx, "MyVariable", 6.7)
-    myvar.set_writable()    # Set MyVariable to be writable by clients
+    myobj = await objects.add_object(idx, "MyObject")
+    myvar = await myobj.add_variable(idx, "MyVariable", 6.7)
+    await myvar.set_writable()  # Set MyVariable to be writable by clients
 
     # starting!
-    server.start()
+    await server.start()
     try:
```
oroulet, line 40: since you are here, can you replace the try/excpet with `async with`

`async with Server(xxx) as server:
      do something

`

examples/server-with-encryption.py
```diff
@@ -1,42 +1,55 @@
...
-        server.stop()
+        # close connection, remove subcsriptions, etc
+        await server.stop()
+
+
+if __name__ == "__main__":
+    loop = asyncio.get_event_loop()
+    loop.set_debug(True)
+    loop.run_until_complete(main())
+    # setup our server
```
oroulet, line 55: also here, now we should use asyncio.run, instead these 3 lines

oroulet:
Can you rebase and I merge asap. There are some conflicts since I update the examplei the main time

JoeyFaulkner:
> Can you rebase and I merge asap. There are some conflicts since I update the examplei the main time

rebased, are the certificate store changes alright?

examples/server-with-encryption.py
JoeyFaulkner: when i try to do this with client-with-encryption.py, it will not connect because i havent set a security policy, what is the best approach here?

for a concrete example, that's

```
async with Server(url=url) as server: # ERROR, security policy None not acceptable
    await server.set_security_policy(blahblahblah)
```

oroulet:
I still cannot merge here. This is the second time I see the same issue in a few days. Also generally use rebase and NOT merge (pull -r origin master , for example)

oroulet:
OK but I can squash so good

swamper123:
In my case: 
https://github.com/FreeOpcUa/opcua-asyncio/blob/432701845167b2627c46549b4dfdbb46004d8d50/tests/test_crypto_connect.py#L156
and:
https://github.com/FreeOpcUa/opcua-asyncio/blob/432701845167b2627c46549b4dfdbb46004d8d50/tests/test_crypto_connect.py#L171
Always fails. Even if asyncio is raising a `TimeoutError`.
My failure rattail goes down here:
https://github.com/FreeOpcUa/opcua-asyncio/blob/88bbc425eb2137917795640196fbde8712898681/asyncua/client/ua_client.py#L203
before the TimeoutError is thrown.

Is it just me or others as well?

PR #183, opened 2020-05-07
Allow for asynchronous subscription callbacks

chrisjbremner:
I am working a server implementation that would greatly benefit from being able to create subscriptions that have asynchronous callbacks. In this PR, I change the callback behavior for subscriptions so that you can use either synchronous or asynchronous definitions:

For example this would be valid:
```python
class MySubHandler:
    async def datachange_notification(self, node, val, data):
        await asyncio.sleep(1)
        print("got datachange")
    
    def event_notification(self, event):
        print("got event")
```

In addition, I made it so that the async callbacks run as tasks and the sync callbacks run in the event loop so that other processing is not blocked.

Would this make sense, or would you propose just using `asyncio.create_task(my_async_coro)` inside the callback instead?

### DIFF
--- a/asyncua/server/internal_subscription.py
+++ b/asyncua/server/internal_subscription.py
@@ -111,7 +111,7 @@ class InternalSubscription:
             #    Subscription.publish_callback -> server internal subscription
             #    UaProcessor.forward_publish_response -> client subscription
             if asyncio.iscoroutinefunction(self.pub_result_callback):
-                asyncio.create_task(self.pub_result_callback(result))
+                self.loop.create_task(self.pub_result_callback(result))
             else:
                 self.loop.call_soon(self.pub_result_callback, result)
 
--- a/asyncua/server/internal_subscription.py
+++ b/asyncua/server/internal_subscription.py
@@ -58,13 +58,13 @@ class InternalSubscription:
             self._task = None
         self.monitored_item_srv.delete_all_monitored_items()
 
-    def _trigger_publish(self):
+    async def _trigger_publish(self):
         """
         Trigger immediate publication (if requested by the PublishingInterval).
         """
         if not self._task and self.data.RevisedPublishingInterval <= 0.0:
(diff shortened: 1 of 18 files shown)


oroulet:
I thought this was already done. I think it is ok. It gives more flexibility to users. I would even telling the possibility for sync handlers

oroulet:
@cbergmiller opinion?

cbergmiller:
Looks like an improvement. I can't see any obvious problems in the changes.

asyncua/server/internal_subscription.py
```diff
@@ -110,7 +110,10 @@ def publish_results(self):
             # The callback can be:
             #    Subscription.publish_callback -> server internal subscription
             #    UaProcessor.forward_publish_response -> client subscription
-            self.pub_result_callback(result)
+            if asyncio.iscoroutinefunction(self.pub_result_callback):
+                self.loop.create_task(self.pub_result_callback(result))
```
oroulet, line 114: ops I did not see that one. create tasks and forgetting them is not a good idea. Can't you make publish_callback() async? I do not think it is called form many places

chrisjbremner: `publish_callback` is already async, are you proposing I do `await self.pub_result_callback(result)` here? If so, we have to make `publish_results`. Making that change results in quite a long cascade of updates from sync functions to async, which wouldn't be that bad, but it eventually requires a change to `trigger_event` to async, which presumably would be a breaking change for anyone who relies on that being a sync function.

oroulet: There is another PR doing that in the pipeline anyway so maybe just go for it

oroulet:
Looks good. But you need to rebase on master so we can merge

oroulet:
github does not let me merge yet. try again to rebase on master `git pull -r origin master`

oroulet:
nevermind I made it using command line

PR #198, opened 2020-06-03
Add async and await statement for HistoryManager.stop

swamper123:
Because a Coroutine was expected, I added an async and await statement, even if this feature seems not to be supported at the moment.


swamper123:
#206

PR #204, opened 2020-06-03
Add async and await for _add_node method in XMLImporter class

swamper123:
I am not sure why: https://github.com/FreeOpcUa/opcua-asyncio/blob/79072d761e3e1e4fcb00ec324bb6c3f4d13f7dbd/asyncua/server/internal_session.py#L111
is async, because it just contains a sync method.

Should have been there something more async?

Anyway, the commited method now it is clearly async.


swamper123:
#206

PR #205, opened 2020-06-03
Add async and await for call Method in InternalSession class

swamper123:


oroulet:
Ops supert i never ment you should create so many pull requests. But in you last request you had both adding async and other things. You can add all related async things in one pull request

oroulet:
Sorry I meant. Not super ;-)

swamper123:
*sigh, grabs keyboard again* ;-)
Than it was a misunderstood, sorry.
I will make a single one later today.

swamper123:
#206

PR #206, opened 2020-06-03
Async multiple methods

swamper123:
Async call method in internal_session
Async history_read method in internal_session
Async get_endpoints method in server
Async delete_nodes method in server and client
Async load_enums in server and client
Async load_type_definitions method in server and Client
Async import_xml in server and Client

Async stop method in HistoryManager class
- Because a Coroutine was expected in this class, I added an async and await statement, even if this feature seems not to be supported at the moment.

I am not sure why:
https://github.com/FreeOpcUa/opcua-asyncio/blob/79072d761e3e1e4fcb00ec324bb6c3f4d13f7dbd/asyncua/server/internal_session.py#L111

is async, because it just contains a sync method.
Should have been there something more async?

Anyway, the commited methods are now clearly async as expected.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -547,15 +547,15 @@ class Client:
         _logger.info("get_namespace_index %s %r", type(uries), uries)
         return uries.index(uri)
 
-    def delete_nodes(self, nodes, recursive=False) -> Coroutine:
-        return delete_nodes(self.uaclient, nodes, recursive)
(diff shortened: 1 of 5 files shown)


swamper123:
So, any stomachache with that?

cbergmiller:
I intentionally omitted `async` where possible when i converted the code to asnycio because it comes with a performance cost. Might be negligible in this case.
Since the tests pass i would say: go for it

swamper123:
@oroulet Have you any concerns about these readability fixes?

PR #209, opened 2020-06-08
User management with certificates

JoeyFaulkner:
This pull request is to create a system of user management/user tracking/user-based permissions on the opc server, tied to their certificate. Each client's certificate is tied to a `User` object, which has a name and a `UserRole` of Admin, User or Anonymous (formerly `User`). The user's role is what is checked for the functions where before the `User` was checked.
Other changes:
- Add name of the user alongside their action in logs, e.g.: `INFO:asyncua.server.uaprocessor:Read request (user certificates/peer-certificate-example-1.der)`
- Create object to decide on permissions (`PermissionRuleset`) and explicitly check permission for every action a client takes.
- Change error when peer certificate is not in certificate handler to `BadUserAccessDenied`
- add testing for permissions 
- add documentation on how to use user management functions including the certificate handler.


asyncua/server/internal_session.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, internal_server, aspace: AddressSpace, submgr: SubscriptionSe
         self.subscription_service: SubscriptionService = submgr
         self.name = name
         self.user = user
+        print(self.user)
```
swamper123: Can you replace that with a logger.info or logger.debug output?

asyncua/server/binary_server_asyncio.py
```diff
@@ -19,11 +19,13 @@ class OPCUAProtocol(asyncio.Protocol):
     """
 
     def __init__(self, iserver: InternalServer, policies, clients, protocol_tasks):
+        # InternalServer should bring a certificate
         self.peer_name = None
         self.transport = None
         self.processor = None
         self._buffer = b''
         self.iserver: InternalServer = iserver
+        print(self.iserver.certificate)
```
swamper123: Here a logger.info or logger.debug as well

asyncua/server/binary_server_asyncio.py
```diff
@@ -37,7 +39,7 @@ def __str__(self):
 
     def connection_made(self, transport):
         self.peer_name = transport.get_extra_info('peername')
-        logger.info('New connection from %s', self.peer_name)
+        logger.error('New connection from %s', self.peer_name)
```
swamper123: Why is that changed to logger.error?

asyncua/server/uaprocessor.py
```diff
@@ -132,11 +132,17 @@ async def process_message(self, seqhdr, body):
             return True
 
     async def _process_message(self, typeid, requesthdr, seqhdr, body):
+        user = self._connection.security_policy.user
+        if user is None:
```
swamper123, line 146: Just an Idea to add either a  __repr__ in User to get the Username or in UserRole maybe?

asyncua/server/address_space.py
```diff
@@ -57,6 +57,7 @@ async def write(self, params, user=User(role=UserRole.Admin)):
         #self.logger.debug("write %s as user %s", params, user)
         res = []
         for writevalue in params.NodesToWrite:
+            print(user)
```
swamper123: Same as with the other prints. Replace them with a logger expression.

asyncua/server/internal_session.py
```diff
@@ -100,7 +100,13 @@ def history_read(self, params) -> Coroutine:
         return self.iserver.history_manager.read_history(params)
 
     async def write(self, params):
-        return await self.iserver.attribute_service.write(params, self.user)
+        if self.user is None:
+            user = User()
+            subscription_result = await self.iserver.attribute_service.write(params, user=user)
+        else:
+            subscription_result = await self.iserver.attribute_service.write(params, user=self.user)
```
oroulet, line 107: put the effective code outside of if/else. just set user variable under if/else

asyncua/server/uaprocessor.py
oroulet: yes do not put logg formating stuff here. find something smarter. like `__str__` or `_repr__`

asyncua/server/uaprocessor.py
```diff
@@ -153,22 +171,22 @@ async def _process_message(self, typeid, requesthdr, seqhdr, body):
...
             if self.session:
                 deletesubs = ua.ua_binary.Primitives.Boolean.unpack(body)
                 await self.session.close_session(deletesubs)
             else:
-                _logger.info("Request to close non-existing session")
+                _logger.info("Request to close non-existing session {}".format(extra_log_str))
 
             response = ua.CloseSessionResponse()
-            _logger.info("sending close session response")
+            _logger.info("sending close session response {}".format(extra_log_str))
```
oroulet, line 182: also never use `.format(xx)`for logging. use `%s` in string and extra arg. It is not efficient. Run pylint on code for example to get all warnings about it

asyncua/server/uaprocessor.py
```diff
@@ -124,6 +128,12 @@ async def process_message(self, seqhdr, body):
             _logger.error("sending service fault response: %s (%s)", status.doc, status.name)
             self.send_response(requesthdr.RequestHandle, seqhdr, response)
             return True
+        except ua.uaerrors.BadUserAccessDenied as e:
+            user = self._connection.security_policy.user
+            _logger.warning(f"{user} attempted to do something they are not permitted to do")
```
oroulet, line 133: same here. never use f-string in logging

asyncua/server/users.py
```diff
@@ -5,10 +5,16 @@
...
     """
     Admin = 0
     Anonymous = 1
     User = 3
+
+
+class User:
+    def __init__(self, role=UserRole.Anonymous, name=None):
+        self.role = role
+        self.name = name
```
oroulet, line 20: here add `__str__` and `__repr__` methods

oroulet: in fact not. make that class use @dataclass instead of handwritten

docs/user_management.md
```diff
@@ -0,0 +1,57 @@
+# User Management
+## Overview
+Currently user management on OPC servers here is done exclusively through certificates.
```
oroulet, line 3: write OPC UA not only OPC

JoeyFaulkner:
@oroulet @swamper123 thanks for the helpful comments! will implement them today :)

oroulet:
@JoeyFaulkner I am not finished. If we touch that user management we should make it pluggable, so we can easily implement other schemes. Is that in your code? I will try to have a look later

JoeyFaulkner:
> @JoeyFaulkner I am not finished. If we touch that user management we should make it pluggable, so we can easily implement other schemes. Is that in your code? I will try to have a look later

It is possible to make a custom `certificate_manager` and `permission_ruleset` which will allow for different schemes, however the checks within the functions for `UserRole.Admin` which were there before this pull request are still there, and require that the user object has a role attribute. It is definitely worth discussing if we should change this :)

tests/test_permissions.py
```diff
@@ -0,0 +1,127 @@
...
+    cert_handler = CertificateHandler()
+    key, cert = request.param
+    await cert_handler.trust_certificate(admin_peer_certificate, user_role=UserRole.Admin)
+    await cert_handler.trust_certificate(user_peer_certificate, user_role=UserRole.User)
+    await cert_handler.trust_certificate(anonymous_peer_certificate, user_role=UserRole.Anonymous)
+    await srv.init()
+    srv.set_endpoint(uri_crypto_cert)
+    srv.set_security_policy([ua.SecurityPolicyType.Basic256Sha256_SignAndEncrypt],
+                            certificate_handler=cert_handler,
+                            permission_ruleset=SimpleRoleRuleset())
```
oroulet, line 67: I might be wrong but that API looks a bit strange to me. As far as I understand you are adding a user management functionality where users are checked using certificated. This is one of several possibilities,some people will not want it but will want to use password for examples

Iwould have expected an API ala
```

@abc
class UserManager:
      def validate_user(self, user, password=None, certificate=None) => Bool:
           raise NotImplementError
      
class UserManageByCertificate:
      def validate_user(self, user, password=None, certificate=None):       
               # add all you certificate code
               return True/False

class InternalServer:
     def __init__(self, x,x,x,x,, user_mananger=None):
          xxxxx



```

JoeyFaulkner: yeah, you are right that this supports only certificate based authentication. The method you've proposed does seem better and more general! let me do some work on it and see if i have any issues.

JoeyFaulkner: one question is how would a username/password be passed to the security policy?

oroulet: Does it need to be passed there? Anyway password arrives in create_session in Internal server. You should have access to everything there

JoeyFaulkner:
Finally passed tests, would be interested to see what you think in this new architecture

asyncua/server/user_managers.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class PermissiveUserManager:
+    def get_user(self, iserver, username=None, password=None, certificate=None):
+        """
+        Default user_manager, does nothing much but check for admin
+        """
+        if username and iserver.allow_remote_admin and username in ("admin", "Admin"):
+            return User(role=UserRole.Admin)
+        else:
```
oroulet, line 17: No need for else here. Just return

oroulet:
I think it looks good 👍. I will try to have a closer look later

asyncua/crypto/permission_rules.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+
+    def check_validity(self, user, action_type, body):
+        raise NotImplementedError
+
+
+class SimpleRoleRuleset(PermissionRuleset):
+    """
+    Standard simple role-based ruleset.
+    Admins alone can write, admins and users can read, and anonymous users can't do anything.
```
swamper123, line 52: Just a question about what you mean with: "anonymous users can't do anythin". Shall "Anonymous" just be a placeholder for an invalid User?
Does this mean they shouldn't connect to the server as well or should they be able the browse root, but nothing more?

swamper123: But on the first sight, it looks good. Nice job °u°) 👍

JoeyFaulkner: its a good question, im not sure what the best course of action is here. One thing to note is that if the certificate handler is passed to the security policy such as https://github.com/FreeOpcUa/opcua-asyncio/pull/209/files#diff-762e1cb0d7270d41d25526c62134823cR27 here, then untrusted certificates wont be able to connect at all. I guess what i was doing here was just to retain the functionality of the original Roles.

asyncua/server/user_managers.py
JoeyFaulkner: I agree that it could be without the else, however in my opinion it reads a bit clearer with the else. I can change it if necessary

oroulet:
@JoeyFaulkner good job. I could accept the PR as-is. but I am wondering about something else.
Why is that certificate handler class only passed to set_security method... 
Is the certificate linked to the security type? do we want a different certificate for every security type?

Doesn't it make sense to have CertificateManager it as a full member of InternalServer? so it can easily be reused everywhere?

oroulet:
btw you need to rebase on master before I can merge

JoeyFaulkner:
> @JoeyFaulkner good job. I could accept the PR as-is. but I am wondering about something else.
> Why is that certificate handler class only passed to set_security method...
> Is the certificate linked to the security type? do we want a different certificate for every security type?
> 
> Doesn't it make sense to have CertificateManager it as a full member of InternalServer? so it can easily be reused everywhere?

I'm not sure, that sounds like it would make sense too, I can implement this if its wanted

oroulet:
the main question is to decide if the CertificateManager is something which has to be different for every security method. I do not think it is, but I never really looked closely at these crypto thigns

JoeyFaulkner:
yeah, its a good question! i also am not particularly experienced in this area. I guess when I started I just read "security policy" and assumed this was a security thing. But i can see your point entirely!

oroulet:
what are the cerificates used by CertificateHandler used for ? only authentication?
then maybe it should be called more precisely something like AuthenticationCertHandler (or Manager btw)

JoeyFaulkner:
so currently, the CertificateHandler decides which certificates can even make a connection with the OPCUA server, it will close any connection doesnt. But you can also feed this same certificate handler to a CertificateUserManager which will associate each certificate with a User object. So yeah I guess you're right that they're only used for authentication, that makes sense as a change in naming

JoeyFaulkner:
that sounds like a good plan!

JoeyFaulkner:
i've implemented this @oroulet , let me know what you think! :)

oroulet:
Looks good. No more comments from me 👍

PR #211, opened 2020-06-14
READY: update: generation from spec-nodeset, node.py, ua_binary.py

AndreasHeine:
according to PR#210

node.py:
-str(node) will return nodeid ("ns=1;i=1005")
ua_binary.py:
-if u pack_array with not a list as input it will log a warning and `return Primitives.Int32.pack(-1) #to prevent crashing while runtime`

asyncua/schemas:
-added generation date
-changed something in the generate_address_space.py which causes issues in ua_binary.py like above!
-regenerate all files with the latest nodesets


AndreasHeine:
wait with the merge! i am on the f-string stuff in asyncua like i wrote in python-opcua # 1071

oroulet:
If the pr is not ready flag it with "WIP:      "

oroulet:
thanks that code had not been updated for a while!

Issue #106, opened 2019-11-29
Running a test OPC-UA server


gkotian:
!! N00b alert !!

I'm just getting into the OPC world, and one of the very first things I'm doing is to try and run a test OPC-UA server using this repo.

I created a docker image with the given [Dockerfile](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/Dockerfile) using:
```
docker build -t test-opc-server .
```

Then I opened a container instance with:
```
docker run --rm -it test-opc-server bash
```

Finally I ran `uaserver` within the container, and got this:
```
root@1bd3fc641716:/# uaserver 
Traceback (most recent call last):
  File "/usr/local/bin/uaserver", line 8, in <module>
    sys.exit(uaserver())
  File "/usr/local/lib/python3.8/site-packages/asyncua/tools.py", line 546, in uaserver
    sync.start_thread_loop()
AttributeError: module 'asyncua.sync' has no attribute 'start_thread_loop'
root@1bd3fc641716:/#
```

What am I missing?

driftregion:
This still a problem in the latest release 0.8.3, and seems unrelated to Docker.
It appears that `sync.start_thread_loop` and `sync.stop_thread_loop` have been removed.
I can get the server to run by removing calls to those functions in `tools.py`

swamper123:
Since there have been some changes since then, is it still an issue @driftregion @gkotian ?

gkotian:
Changing priorities have meant that I'm not working on OPC-related tasks at the moment. Closing this issue as obsolete would be fine by me. I will open a new one in the future if I run into trouble again.

PR #212, opened 2020-06-17
add basic test for CLI tools

driftregion:
This parameterized test catches https://github.com/FreeOpcUa/opcua-asyncio/issues/106

### DIFF
--- a/asyncua/tools.py
+++ b/asyncua/tools.py
@@ -5,6 +5,7 @@ import argparse
 from datetime import datetime, timedelta
 import math
 import time
+import concurrent.futures._base
 
 try:
     from IPython import embed
@@ -15,7 +16,7 @@ except ImportError:
         code.interact(local=dict(globals(), **locals()))
 
 from asyncua import ua
-from asyncua import Client
+from asyncua import Client, Server
 from asyncua import Node, uamethod
 from asyncua import sync
 from asyncua.ua.uaerrors import UaStatusCodeError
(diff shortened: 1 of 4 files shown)


oroulet:
I think there are tests for tools , but maybe not for all. so this is fine. But it looks like you need to make it pass :-)

oroulet:
What is broken?

swamper123:
> tests/test_tools.py::test_that_tool_can_be_invoked_without_internal_error[uasubscribe]
> 
>   /home/travis/build/FreeOpcUa/opcua-asyncio/tests/test_tools.py:15: RuntimeWarning: coroutine 'uasubscribe' was never awaited
> 
>     tool()

behind `uasubscribe ` is a Coroutine, so it crashes here. Maybe using `asycio.run()` could solve the problem, because it doesn't matter if it runs a coro or a function.

https://github.com/FreeOpcUa/opcua-asyncio/blob/3353bc874c4511dab7187bf8d7614309331cf903/asyncua/tools.py#L388

asyncua/tools.py
```diff
@@ -614,26 +616,30 @@ async def _uadiscover():
...
-
-    for i, ep in enumerate(await client.connect_and_get_server_endpoints(), start=1):
-        print('Endpoint {0}:'.format(i))
-        for (n, v) in endpoint_to_strings(ep):
-            print('  {0}: {1}'.format(n, v))
-        print('')
+        for i, ep in enumerate(await client.connect_and_get_server_endpoints(), start=1):
+            print('Endpoint {0}:'.format(i))
+            for (n, v) in endpoint_to_strings(ep):
+                print('  {0}: {1}'.format(n, v))
```
oroulet, line 638: you can use f-string instead of format

driftregion:
Hi, I'm having a much better development experience with https://github.com/FreeOpcUa/python-opcua.
I'm closing this to focus on issues there.

oroulet:
@driftregion what issues do you have with opcua.-python that you donot have with python-opcua? python-opcua isin maintenance mode so I encourage everyone to focus on opcua-asyncio for new things

driftregion:
thanks for asking

It's my workflow that's the issue.  ipdb is such a powerful tool, I rarely
need to search documentation anymore when I can drop an ipdb breakpoint in
the program and dig around until I find what I'm looking for.

last I experienced, ipdb is unpleasant to use with async libs including
asyncio.  I.e. one can break inside coroutines, but can't await other
coroutines, which makes figuring out large complex IO libraries
exceptionally difficult.

can you share what your workflow is like in this regard?

swamper123:
@driftregion It seems that IPython is on this problem. Async/Await is supported with Version IPython 7.0 and since then it still has some features in experimental.
Anyway they are working on it in the ipython repo and with the last update they even had a fix for ipdb.

My 5 cents for the workflow question: I am using the PyCharm debugger, which is a great tool in my opinion. It shows everything about a variable, works without any trouble with async stuff and the IDE is pretty neat, too. I haven't worked so much with IPython, so I can't compare these two with one another.

oroulet:
At least I know ipython/embed has quite good suport for coroutines. Not sure that works all the time though...  `%autowait on`

swamper123:
I had a look on [the latest test reports e.g.](https://travis-ci.org/github/FreeOpcUa/opcua-asyncio/jobs/724503867#L741). It seems that `test_that_tool_can_be_invoked_without_internal_error` throws errors in the subprocess, but tests still pass. 
I guess this isn't wanted behaviour?

Issue #208, opened 2020-06-04
How catch error while renewing session Task?


xvax:
Hello! 
Dear @oroulet could you please help to understand how it possible to catch this exception?
https://github.com/FreeOpcUa/opcua-asyncio/blob/79072d761e3e1e4fcb00ec324bb6c3f4d13f7dbd/asyncua/client/client.py#L388-L390

I know only method like this
`self.client._renew_channel_task.add_done_callback(self.future_callback_error_logger)`
where in callback call `.result` on Future and handle exceptions, but this is access to protected attribute and feels not right.

First of all issue happens like this

> asyncio Task exception was never retrieved
future: <Task finished name='Task-30774' coro=<Client._renew_channel_loop() done, defined at /usr/local/lib/python3.8/site-packages/asyncua/client/client.py:361> exception=TimeoutError() created at /usr/local/lib/python3.8/site-packages/asyncua/client/client.py:358>
source_traceback: Object created at (most recent call last):
  File "run.py", line 651, in <module>
    p.start()
  File "/usr/local/lib/python3.8/multiprocessing/
...
  File "/usr/local/lib/python3.8/site-packages/asyncua/client/client.py", line 218, in connect
    await self.create_session()
  File "/usr/local/lib/python3.8/site-packages/asyncua/client/client.py", line 358, in create_session
    self._renew_channel_task = self.loop.create_task(self._renew_channel_loop())
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/asyncua/client/client.py", line 373, in _renew_channel_loop
    await self.open_secure_channel(renew=True)
  File "/usr/local/lib/python3.8/site-packages/asyncua/client/client.py", line 267, in open_secure_channel
    result = await self.uaclient.open_secure_channel(params)
  File "/usr/local/lib/python3.8/site-packages/asyncua/client/ua_client.py", line 271, in open_secure_channel
    return await self.protocol.open_secure_channel(params)
  File "/usr/local/lib/python3.8/site-packages/asyncua/client/ua_client.py", line 200, in open_secure_channel
    result = await asyncio.wait_for(
  File "/usr/local/lib/python3.8/asyncio/tasks.py", line 490, in wait_for
    raise exceptions.TimeoutError()
asyncio.exceptions.TimeoutError

swamper123:
I would say that a try/except block is missing here.
[...]

PR #215, opened 2020-06-18
added exception handling to open_secure_channel and _renew_channel_loop

bobaflash:
Maybe I have a solution for issue #208

A minimal example to handle errors could be something like this:
```
import asyncio

from asyncua import Client


class OpcSubscriptionHandler(object):

    def datachange_notification(self, node_value, node):
        print(f'{node.subscription_data.node.nodeid}: {node_value} ')

    def event_notification(self, event):
        print("Python: New event", event)


def execption_handler(loop, context):
    print('Do something when error occur!')
    print(context)
    loop.stop()


async def main():
    client = Client(url='opc.tcp://10.60.123.123:4841', timeout=1, loop=loop)
    client.session_timeout = 3000
    await client.connect()
    subscription = await client.create_subscription(500, OpcSubscriptionHandler)
    node = client.get_node("ns=6;s=::ResReader:Cooling_Water.WaterTemp_Out")
    await subscription.subscribe_data_change(node)



if __name__ == '__main__':
    loop.set_exception_handler(execption_handler)
    loop.create_task(main())
    loop.run_forever()

```
A use case for this is to handle failure if a Ethernet cable get unplugged.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -385,8 +385,12 @@ class Client:
                 _logger.debug("server state is: %s ", val)
         except asyncio.CancelledError:
             pass
-        except:
+        except Exception as err:
             _logger.exception("Error while renewing session")
+            self.loop.call_exception_handler({
+                'message':'Error while renewing session',
+                'exception': err
+            })
             raise
 
     def server_policy_id(self, token_type, default):
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -197,17 +197,23 @@ class UASocketProtocol(asyncio.Protocol):
         return await asyncio.wait_for(ack, self.timeout)
 
     async def open_secure_channel(self, params):
-        self.logger.info("open_secure_channel")
-        request = ua.OpenSecureChannelRequest()
-        request.Parameters = params
-        result = await asyncio.wait_for(
-            self._send_request(request, message_type=ua.MessageType.SecureOpen),
-            self.timeout
-        )
-        response = struct_from_binary(ua.OpenSecureChannelResponse, result)
-        response.ResponseHeader.ServiceResult.check()
-        self._connection.set_channel(response.Parameters, params.RequestType, params.ClientNonce)
-        return response.Parameters
+        try:
+            self.logger.info("open_secure_channel")
+            request = ua.OpenSecureChannelRequest()
+            request.Parameters = params
+            result = await asyncio.wait_for(
+                self._send_request(request, message_type=ua.MessageType.SecureOpen),
+                self.timeout
+            )
+            response = struct_from_binary(ua.OpenSecureChannelResponse, result)
+            response.ResponseHeader.ServiceResult.check()
+            self._connection.set_channel(response.Parameters, params.RequestType, params.ClientNonce)
+            return response.Parameters
+        except Exception as err:
+            self.loop.call_exception_handler({
+                'message': 'Error while open secure channel',
+                'exception': err
+            })
 
     async def close_secure_channel(self):
         """


swamper123:
An exception handler for such a case could be good, especially when asyncio has this feature to do so.

How the handler shall behave would be the programmers choice (maybe I would write a comment here, because this may be a regular question how to deal with it).

oroulet:
The problem is that you are only catching a small subset of what might happen... Renew_channel/session is not called often...

oroulet:
Maybe you could make something similar with a second task reading server state every 5s and calling a callback if any error. But do not modify ua_client for that

swamper123:
@oroulet is right, just recognised that the session is just renewed after ~1hr.
Either polling the ServerStatus or subscribing would be a way to deal with it.

Here [a link ](https://reference.opcfoundation.org/v104/Core/docs/Part4/6.7/)how the foundation wants to deal with such reconnections.

bobaflash:
I'm using a subscription... How can I access the Server status flag which is named in the link above?

swamper123:
Afaik the ServerStatus State can be found as node value in ns=0;i=2259.
When you got a Client GUI (like the lovely [FreeOpcUa Clien-Gui](https://github.com/FreeOpcUa/opcua-client-gui))t, you'll find it in Server->ServerStatus->State.
The possible State-Values can be found [here.](https://reference.opcfoundation.org/v104/Core/DataTypes/ServerState/).

AndreasHeine:
do not use subscription for that, most products with uaservers does not change serverstatus... polling is the best way to do it! i use always the servicelevel which doubles as check for connecting to a redundant server instance.

bobaflash:
I solved it in an other way...
I was looking for a solution to regconize when an ethernet wire got unplugged or broken. Actually I can't catch some kind of exceptions. So to get this information and provide it to other modules, I start an separate task where I perform a ping to the SPS. According to the answer I set a flag "connected". With this information I can control the behavior of my application. E.g. it can inform other applications about the missing data flow

swamper123:
@bobaflash Yout idea is clever and I guess will work well with your application.
The thing is, that OPC UA is not for PLCs alone, so you may have a PC as server. While a PLC is just doing it's thing while running, a PC could be connected, but the server is not running.
I am not aware of different PLC types, but maybe it is even possible for PLCs to turn the server on/off, but are still reachable via Ethernet.
So you are right about the question "is there a connection possible in a kind of way?", but not about "is the server still up? Shall I reconnect?".

oroulet:
Instead of ping. Read the server state nice (as done in KeepAlive) and you know if the opcua server is up and in what state

bobaflash:
@oroulet Should I use an extra connection for this?
What is your opinion?

This could be an minimal working example:

```
from asyncua import Client
import asyncio

async def connect():
    while True:
        try:
            await client.connect()
            stat = client.get_node("ns=0;i=2259")
            print(await stat.get_value())
            await client.disconnect()
            await asyncio.sleep(1)
        except:
            print('Broken')



if __name__ == '__main__':
    client = Client(url='opc.tcp://10.60.123.123:4841', timeout=1, loop=loop)
    loop.create_task(connect())
    loop.run_forever()

```

bobaflash:
I'm not sure if it's a good idea to open and close the connection every second.

swamper123:
@bobaflash Well...this could be another way, but if you already have a connection, you can just read the node value of server state and catch an exception if this fails or server_state is not zero.

oroulet:
No. no new connection. The doal is to know if connection works. Just do as  in your PR but instead of using the KeepAlive task. Make your own in Client class that starts after connection and stops when disconnecting. And call you handler if there is an error.

AndreasHeine:
maybe its helpful that the way i build up my [client](https://github.com/AndreasHeine/opcua-sub-to-websocket/blob/39beee1c466d66cb00742e37298b1b11ddf24053/async_client.py#L68) just an idea!

bobaflash:
Finally i solved it in this way:

```
from asyncua import Client
from asyncua.ua.uaerrors import BadTooManySessions
import asyncio
from datetime import datetime


class OpcuaClient:

    def __init__(self):
        self.disconnected = asyncio.Event()
        self.loop = asyncio.get_event_loop()
        self.client = Client(url='opc.tcp://10.60.123.123:4841', timeout=1, loop=self.loop)

        self.opcua_task = self.loop.create_task(self.connect())
        self.keep_alive_task = self.loop.create_task(self.keep_alive())
        self.loop.run_forever()

    async def connect(self):

        while True:
            print(self.disconnected.is_set())
            await self.disconnected.wait()
            try:
                print('try to connect')
                await asyncio.wait_for(self.client.connect(),timeout=1.0)
                print('connected')
                self.disconnected.clear()
            except asyncio.TimeoutError:
                await self.connect()
            except OSError:
                await asyncio.sleep(3)
                await self.connect()
            except BadTooManySessions:
                await asyncio.sleep(5)
                print('Too many sessions on PLC')
                print(datetime.utcnow())

    async def keep_alive(self):
        while True:
            await asyncio.sleep(1)
            if not self.disconnected.is_set():
                try:
                    stat = self.client.get_node("ns=0;i=2259")
                    print(await stat.get_value())
                except:
                    self.disconnected.set()
                    print('connection lost')


if __name__ == '__main__':
    test_client = OpcuaClient()
```

PR #216, opened 2020-06-18
Add get_properties method to sync version of Node

### LINKED ISSUE #213: Compatibility of sync module with python-opcua
Switched from your `python-opcua` since it is in maintenance mode, planning to go async as soon as possible, but for now I can't yet and use `asyncua.sync` module, which gives me a problem:

`AttributeError: 'Node' object has no attribute 'get_properties'`

Looking at the code I found possible solution: add code below to `asyncua.sync.Node` class

```python
@syncmethod
def get_properties(self):
    pass
```

Can open a Pull Request for this. What do you think?

HarrySky:
Fixes #213

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -291,6 +291,10 @@ class Node:
         pass
 
     @syncmethod
+    def get_properties(self):
+        pass
+
+    @syncmethod
     def get_children_descriptions(
         self,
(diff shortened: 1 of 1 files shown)


HarrySky:
@oroulet is it possible to release new version in the near future? Would love to stop patching `site-packages` :smile:

PR #217, opened 2020-06-19
add ability to find endpoints for unauthenticated user

JoeyFaulkner:
I made the mistake earlier thinking that GetEndpoints should be only available to authenticated users. However clients such as UAExpert do a GetEndpoints before authenticating. I think therefore GetEndpoint is necessary to be allowed to unauthenticated users.


oroulet:
Correct.

PR #219, opened 2020-06-19
Improve string performances

swamper123:
Replace .format strings with f-strings
Remove unused import statements
Make code more PEP8 conform


swamper123:
_No autogenerated or "do not touch"-file was harmed in this PR._

AndreasHeine:
in the generation files are is lot of mix/match f-string and .format... maybe i should change that too... would be consistend then!

swamper123:
There have been more files partly, but I either don't know how to replace them in a good way or wasn't sure if I should have replace them, so I left them as they have been.

PR #229, opened 2020-07-07
write callbacks and user metadata attaching to server callbacks

JoeyFaulkner:
I'm not sure if this is the best way to go about this, so am happy and not offended if this pull request gets closed and you tell me there is a better way. 

For a task I'm working on at the moment, I need to have functionality where a function is called every time a write is performed on the opcua server and it needs to know which user carried out the write. On top of this, the function itself needs to be async. So I implemented the possibility to make these callbacks asynchronous (but not a requirement)

Please let me know if there is a better way or any changes that need to be made! :)


tests/test_callback.py
```diff
@@ -0,0 +1,87 @@
...
+port_num = 48560
+
+
+def create_monitored_items(event, dispatcher):
+    print("Monitored Item")
+
+    for idx in range(len(event.response_params)):
+        if (event.response_params[idx].StatusCode.is_good()):
+            nodeId = event.request_params.ItemsToCreate[idx].ItemToMonitor.NodeId
+            print("Node {0} was created".format(nodeId))
```
swamper123, line 18: While you used f-strings further down and because they are way faster than older formats, we should use them as standard. Have a look in your other lines as well.

AndreasHeine:
binding a callback to a node write would be a important feature for audit usage! 

https://documentation.unified-automation.com/uasdkhp/1.0.0/html/_doc__opc_ua__base_model__event_types.html#Doc_OpcUa_AuditUpdateEventType

https://documentation.unified-automation.com/uasdkhp/1.0.0/html/_doc__opc_ua__base_model__event_types.html#Doc_OpcUa_AuditWriteUpdateEventType

https://reference.opcfoundation.org/v104/Core/docs/Part2/4.14.1/

AndreasHeine:
maybe its stupid... =>it should be possible to make a node you want to write just read-only and provide a opcua method to write to the node value then you can log the interaction.

swamper123:
@AndreasHeine This would be a more simple way, but I don't think that this would be a good way (or not the way where I would say "merge it"). 
If a Node is marked as read-only then nobody should be able to touch it from "outside", neither directly nor via methods or other ways. Just "intern processes" should touch it.

oroulet:
I did not even know that file existed... What is that?
it is in common directory but there is a doc string saying: "server side implementation of callback events"...

oroulet:
and since to be only used in server code...

oroulet:
I try to check all PR but I am always surprised to see what passed my eyes...

oroulet:
the code is also outdated with a lot of inheriting of "object"

oroulet:
OK. looks like some very partial instrumentation.
Might be OK, although the name server_callback_dispatcher is not very obvious. maybe auditing_service is better... any better proposition?

swamper123:
@oroulet Your proposal is ok. Sounds better. But still could there be some concerns about race conditions?

oroulet:
@JoeyFaulkner are you still working on that one?

JoeyFaulkner:
> @JoeyFaulkner are you still working on that one?

yes, sorry about the delay i got caught up with other things. Just writing some documentation now for something else and will try and implement all of your changes this afternoon :)

JoeyFaulkner:
> OK. looks like some very partial instrumentation.
> Might be OK, although the name server_callback_dispatcher is not very obvious. maybe auditing_service is better... any better proposition?

@oroulet , I think im not a fan of `auditing_service`, as the callbacks by the server can actually perform functions and change the server and not just store logs, for example, and i wouldnt think that would be possible with an auditing service. is `server_callback_handler` better? or keep it as is?

oroulet:
Can they really change the behavior?
What about InternalCallbacksService then?

JoeyFaulkner:
That sounds better! will rename it that

JoeyFaulkner:
anything else @oroulet?

asyncua/server/internal_server.py
```diff
@@ -271,13 +271,13 @@ def subscribe_server_callback(self, event, handle):
         """
         Create a subscription from event to handle
         """
-        self.server_callback_dispatcher.addListener(event, handle)
+        self.server_callback_service.addListener(event, handle)
```
oroulet, line 274: why is the word server repeated? Can't we just call it self.callback_service?

asyncua/common/callback.py
```diff
@@ -55,9 +58,16 @@ def dispatch(self, eventName, event=None):
...
+            await self.call_listener(event, listener)
+
         return event
 
+    async def call_listener(self, event, listener):
+        if asyncio.iscoroutinefunction(listener):
+            await listener(event, self)
+        else:
+            listener(event, self)
+
```
oroulet, line 70: can't we just require these callbacks to be async? probably fine for users and a bit less code

tests/test_callback.py
```diff
@@ -0,0 +1,87 @@
+from asyncua import Server, Client
```
oroulet, line 1: and maybe call that file test_callback_service. The name callbackis very generric so it is hard to know what this is about

oroulet:
@JoeyFaulkner can you fix these small things so we can merge

JoeyFaulkner:
yep, will do so today!

asyncua/common/callback.py
JoeyFaulkner: this would be a breaking change potentially for some users though. Is that a problem?

JoeyFaulkner: @oroulet what do you think? if we keep it like this you can just merge

oroulet:
@JoeyFaulkner Sorry I had to think again about the goal of that entire thing:
I just saw a bug request where users wanted to do some tests before accepting a write request. Could this beeng done with that CallbackService? but it looks like you call the callback AFTER the write is performed. Should we call it before?maybe have two differents so people can modify a write request or refuse it. but maybe this is too costly...

oroulet:
but maybe this should be implemented in AddressSpace. .. There is already a callback stuff for read... should probably be one for write...

swamper123:
@oroulet What method do you especially mean? `read_attirbute_value`?

JoeyFaulkner:
hey, just FYI, im on holiday this week so won't be able to look at this again until next monday.

JoeyFaulkner:
@oroulet I'm a little confused on the current status of this pull request, what does it mean to deal with this in Address space exactly? The callbacks i am interested in actually would work better if the function was called before the write, so happy to change that, but it would make the writes different from the subscribes etc.

oroulet:
Can't you just name the write event with am obvious name so users understand this is processes before the write?
One event for before and one for after? Same with reads?

JoeyFaulkner:
ok will do!

JoeyFaulkner:
is this sort of what you're looking for @oroulet ? :)

oroulet:
not bad! I have no better ideas for now. Thanks for your work

oroulet:
shit it looks like you need to rebase the merge request so I can merge it

JoeyFaulkner:
done! :)

oroulet:
@JoeyFaulkner I still cannot merge. you need to rebase on master: `git pull -r origin master`

JoeyFaulkner:
this is quite time consuming, what's the benefit of this over a merge?

oroulet:
@JoeyFaulkner sorry I found out I could merge anyway here

oroulet:
@JoeyFaulkner but usually using pull rebase instead of merge make a much cleaner history so it is easier to find bugs.

JoeyFaulkner:
good to know! thanks

JoeyFaulkner:
thanks for your patience with this PR, aswell

PR #237, opened 2020-07-15
Don't return AccessDenied on FindServers requests

### LINKED ISSUE #236: adding an asyncua server in ignition fails
Trying to add an asyncua server from [ignition](https://en.wikipedia.org/wiki/Ignition_SCADA) fails with the following error message, even when using the permissive user manager : 

```
 UaServiceFaultException: status=Bad_UserAccessDenied, message=User does not have permission to perform the requested operation.
```

Trying to investigate this, it seems that Ignition sends a message with the following header and body : 

```
Header(type:b'MSG', chunk_type:b'F', body_size:92, channel:18)
```

```
b'\x01\x00\xa6\x01\x00\x00 \xb8n\xb9\xe5Z\xd6\x01\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\xff\xff`\xea\x00\x00\x00\x00\x00#\x00\x00\x00opc.tcp://host.docker.internal:4840\x00\x00\x00\x00\x00\x00\x00\x00'
```

(`opc.tcp://host.docker.internal:4840` being the address of the server)

and parsing the message falls inside [this condition in `UaProcessor._process_message`](https://github.com/FreeOpcUa/opcua-asyncio/blame/8c0fe3442d71c1ae90afbeb8026737506b0b21bc/asyncua/server/uaprocessor.py#L153-L154) : 

```python
        elif self.session is None:
            raise ua.uaerrors.BadUserAccessDenied
```

This condition seems to have been added in https://github.com/FreeOpcUa/opcua-asyncio/commit/47920a7fe022a8682c833ff99124273943bf3982 , and the bug does not occur in v0.8.3 of this package (neither in the synchronous [python-opcua](https://github.com/FreeOpcUa/python-opcua))

lovasoa:
Fixes #236

### DIFF
--- a/asyncua/server/uaprocessor.py
+++ b/asyncua/server/uaprocessor.py
@@ -147,10 +147,12 @@ class UaProcessor:
     async def _process_message(self, typeid, requesthdr, seqhdr, body):
         if typeid in [ua.NodeId(ua.ObjectIds.CreateSessionRequest_Encoding_DefaultBinary),
                       ua.NodeId(ua.ObjectIds.ActivateSessionRequest_Encoding_DefaultBinary),
+                      ua.NodeId(ua.ObjectIds.FindServersRequest_Encoding_DefaultBinary),
                       ua.NodeId(ua.ObjectIds.GetEndpointsRequest_Encoding_DefaultBinary)]:
             # The connection is first created without a user being attached, and then during activation the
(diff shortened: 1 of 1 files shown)


lovasoa:
Related to #217
@JoeyFaulkner

PR #238, opened 2020-07-18
supprt DataTypeDefinition + small fixes

oroulet:


oroulet:
ops there is a big bug. fixing tomorrow..

asyncua/common/structures104.py
```diff
@@ -0,0 +1,197 @@
...
+from enum import IntEnum
+import logging
+import re
+
+from asyncua import ua
+
+
+logger = logging.getLogger(__name__)
+
+def clean_name(name):
```
swamper123, line 13: Here should be a logger warning in case some characters are removed/replaced, because of conflicts with Python.

asyncua/common/structures104.py
```diff
@@ -0,0 +1,197 @@
...
+            raise RuntimeError(f"Unknown field datatype for field: {field} in structure:{name}")
+        uatype = prefix + ua.ObjectIdNames[field.DataType.Identifier]
+        if field.ValueRank >= 1 and uatype == 'Char':
+            uatype = 'String'
+        uatypes.append((field, uatype))
+        code += f"        ('{field.Name}', '{prefix + uatype}'),\n"
+    code += "    ]\n"
+    code += f"""
+    def __str__(self):
+        vals = [name + ": " + str(val) for name, val in self.__dict__.items()]
```
swamper123, line 93: Use f-string to concatenate strings.

asyncua/common/structures104.py
```diff
@@ -0,0 +1,197 @@
...
+async def load_data_type_definitions(server, base_node=None):
+    if base_node is None:
+        base_node = server.nodes.base_structure_type
+    for desc in await base_node.get_children_descriptions(refs=ua.ObjectIds.HasSubtype):
+        if desc.BrowseName.Name == "FilterOperand":
+            #FIXME: find out why that one is not in ua namespace...
+            continue
+        if hasattr(ua, desc.BrowseName.Name):
+            continue
+        logger.warning("Registring structure %s %s", desc.NodeId, desc.BrowseName)
```
swamper123, line 149: Should the logger level be warning? It looks like wanted behaviour.

asyncua/common/structures104.py
```diff
@@ -0,0 +1,197 @@
...
+    exec(code, env)
+    return env
+
+
+async def load_data_type_definitions(server, base_node=None):
+    if base_node is None:
+        base_node = server.nodes.base_structure_type
+    for desc in await base_node.get_children_descriptions(refs=ua.ObjectIds.HasSubtype):
+        if desc.BrowseName.Name == "FilterOperand":
+            #FIXME: find out why that one is not in ua namespace...
```
swamper123, line 145: Log this as a warning.

oroulet:
works now on some test server.. was more complicated tha I hoped

asyncua/common/structures104.py
oroulet: not sure... must be a bug on our side...

asyncua/common/structures104.py
oroulet: only temporary. AlthoughI think many users might be interested

asyncua/common/structures104.py
oroulet: how in that case? show me something ;-)

asyncua/common/structures104.py
oroulet: good point

asyncua/common/structures104.py
swamper123: `vals = [f"{name}:  {str(val)}" for name, val in self.__dict__.items()] ` should work, shouldn't it?

oroulet: I just tested and it does not work... must be some limitation in f-string...

oroulet: yes it works but it was a bit hard with f-string inside f-string

oroulet:
Still some work. But merging to get more testing from others

PR #239, opened 2020-07-20
Update node.py

AndreasHeine:
Edit: down below!

PR #242, opened 2020-07-20
Update instantiate_util.py

AndreasHeine:
Result of PR #239

```python
import os, asyncio
from asyncua import Server, ua, uamethod
from asyncua.common.instantiate_util import instantiate

NAMESPACE = "http://test.de/UA"

async def main():
    # Serversetup
    server = Server()
    await server.init()
    server.set_endpoint("opc.tcp://127.0.0.1:4840")
    idx = await server.register_namespace(NAMESPACE)

    objects = server.get_objects_node()

    base_object_type = server.get_node(ua.ObjectIds.BaseObjectType)

    my_object_type = await base_object_type.add_object_type(idx, "myObjectType")
    mandatory = await my_object_type.add_variable(idx, "Mandatory", 0.0, ua.VariantType.Float)
    await mandatory.set_modelling_rule(True)
    optional = await my_object_type.add_variable(idx, "Optional", 0.0, ua.VariantType.Float)
    await optional.set_modelling_rule(False)

    await objects.add_object(idx, "TestObj", my_object_type)

    async with server:
        while 1:
            await asyncio.sleep(1)

# Start Server
if __name__ == "__main__":
    asyncio.run(main())
```

Unbenannt

found a bug in instantiate_util.py
OLD:
```python
                    if not instantiate_optional and refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
                        logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                            addnode.BrowseName)
                        continue
```
NEW (replaces the and with an or):
```python
                    if not instantiate_optional or refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
                        logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                            addnode.BrowseName)
                        continue
```

Unbenannt

### DIFF
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -95,7 +95,7 @@ async def _instantiate_node(server,
                     if not refs:
                         # spec says to ignore nodes without modelling rules
                         logger.info("Instantiate: Skip node without modelling rule %s as part of %s",
-                        c_rdesc.BrowseName, addnode.BrowseName)
+                                    c_rdesc.BrowseName, addnode.BrowseName)
                         continue
                         # exclude nodes with optional ModellingRule if requested
                     if refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -123,7 +123,8 @@ async def _instantiate_node(server,
                             res.AddedNodeId,
                             c_rdesc,
                             nodeid=ua.NodeId(namespaceidx=res.AddedNodeId.NamespaceIndex),
-                            bname=c_rdesc.BrowseName,instantiate_optional=instantiate_optional
+                            bname=c_rdesc.BrowseName,
+                            instantiate_optional=instantiate_optional
                         )
                     added_nodes.extend(nodeids)
     return added_nodes
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -12,7 +12,7 @@ from .node_factory import make_node
 logger = logging.getLogger(__name__)
 
 
-async def instantiate(parent, node_type, nodeid=None, bname=None, dname=None, idx=0, instantiate_optional=False):
+async def instantiate(parent, node_type, nodeid=None, bname=None, dname=None, idx=0, instantiate_optional=True):
     """
     instantiate a node type under a parent node.
     nodeid and browse name of new node can be specified, or just namespace index
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -98,7 +98,7 @@ async def _instantiate_node(server,
                             c_rdesc.BrowseName, addnode.BrowseName)
                         continue
                         # exclude nodes with optional ModellingRule if requested
-                    if refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
+                    if refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
                         # instatiate optionals
                         if not instantiate_optional:
                             logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -98,7 +98,8 @@ async def _instantiate_node(server,
                             c_rdesc.BrowseName, addnode.BrowseName)
                         continue
                         # exclude nodes with optional ModellingRule if requested
-                    if refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):#not instantiate_optional or refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
+                    if refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
+                        # instatiate optionals
                         if not instantiate_optional:
                             logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                                 addnode.BrowseName)
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -49,7 +49,7 @@ async def _instantiate_node(server,
                             bname,
                             dname=None,
                             recursive=True,
-                            instantiate_optional=False):
+                            instantiate_optional=True):
     """
     instantiate a node type under parent
     """
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -95,7 +95,7 @@ async def _instantiate_node(server,
                     if not refs:
                         # spec says to ignore nodes without modelling rules
                         logger.info("Instantiate: Skip node without modelling rule %s as part of %s",
-                            c_rdesc.BrowseName, addnode.BrowseName)
+                        c_rdesc.BrowseName, addnode.BrowseName)
                         continue
                         # exclude nodes with optional ModellingRule if requested
                     if refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -98,7 +98,7 @@ async def _instantiate_node(server,
                                     c_rdesc.BrowseName, addnode.BrowseName)
                         continue
                         # exclude nodes with optional ModellingRule if requested
-                    if not instantiate_optional and refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
+                    if not instantiate_optional or refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
                         logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                                     addnode.BrowseName)
                         continue
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -99,9 +99,7 @@ async def _instantiate_node(server,
                         continue
                         # exclude nodes with optional ModellingRule if requested
                     if refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):#not instantiate_optional or refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
-                        if instantiate_optional:
-                            pass
-                        else:
+                        if not instantiate_optional:
                             logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                                 addnode.BrowseName)
                             continue
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -113,7 +113,8 @@ async def _instantiate_node(server,
                             res.AddedNodeId,
                             c_rdesc,
                             nodeid=ua.NodeId(identifier=inst_nodeid, namespaceidx=res.AddedNodeId.NamespaceIndex),
-                            bname=c_rdesc.BrowseName
+                            bname=c_rdesc.BrowseName,
+                            instantiate_optional=instantiate_optional
                         )
                     else:
                         nodeids = await _instantiate_node(
@@ -122,7 +123,7 @@ async def _instantiate_node(server,
                             res.AddedNodeId,
                             c_rdesc,
                             nodeid=ua.NodeId(namespaceidx=res.AddedNodeId.NamespaceIndex),
-                            bname=c_rdesc.BrowseName
+                            bname=c_rdesc.BrowseName,instantiate_optional=instantiate_optional
                         )
                     added_nodes.extend(nodeids)
     return added_nodes
--- a/asyncua/common/instantiate_util.py
+++ b/asyncua/common/instantiate_util.py
@@ -12,7 +12,7 @@ from .node_factory import make_node
 logger = logging.getLogger(__name__)
 
 
-async def instantiate(parent, node_type, nodeid=None, bname=None, dname=None, idx=0, instantiate_optional=True):
+async def instantiate(parent, node_type, nodeid=None, bname=None, dname=None, idx=0, instantiate_optional=False):
     """
     instantiate a node type under a parent node.
     nodeid and browse name of new node can be specified, or just namespace index
@@ -49,7 +49,7 @@ async def _instantiate_node(server,
                             bname,
                             dname=None,
                             recursive=True,
-                            instantiate_optional=True):
+                            instantiate_optional=False):
     """
     instantiate a node type under parent
     """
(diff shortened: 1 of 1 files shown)


AndreasHeine:
```python
import os, asyncio
from asyncua import Server, ua, uamethod
from asyncua.common.instantiate_util import instantiate

NAMESPACE = "http://test.de/UA"

async def main():
    # Serversetup
    server = Server()
    await server.init()
    server.set_endpoint("opc.tcp://127.0.0.1:4840")
    idx = await server.register_namespace(NAMESPACE)

    objects = server.get_objects_node()

    base_object_type = server.get_node(ua.ObjectIds.BaseObjectType)

    my_object_type = await base_object_type.add_object_type(idx, "myObjectType")
    mandatory = await my_object_type.add_variable(idx, "Mandatory", 0.0, ua.VariantType.Float)
    await mandatory.set_modelling_rule(True)
    optional = await my_object_type.add_variable(idx, "Optional", 0.0, ua.VariantType.Float)
    await optional.set_modelling_rule(False)
    none = await my_object_type.add_variable(idx, "None", 0.0, ua.VariantType.Float)
    await none.set_modelling_rule(None)

    await objects.add_object(idx, "TestObj", my_object_type)

    
    await instantiate(objects, my_object_type, ua.NodeId.from_string(f"ns={idx};s=Test2"), instantiate_optional=False)

    async with server:
        while 1:
            await asyncio.sleep(1)

# Start Server
if __name__ == "__main__":
    asyncio.run(main())
```

Unbenannt2

need a second pair of eyes on it! @oroulet @swamper123 
dont realy understand for what instantiate_optional is for...

swamper123:
@AndreasHeine I am still trying to figuring it out as well, but I think this is about these [Modelling rules]?(https://reference.opcfoundation.org/v104/Core/docs/Part3/6.4.4/#6.4.4.2) Correct me if I'm wrong here.

AndreasHeine:
If i understand it right with "instantiate_optional" you should be able to instantiate a type incl. optionals !? but its not working i would expect it...

AndreasHeine:
this appears to be right:
```python
async def instantiate(parent, node_type, nodeid=None, bname=None, dname=None, idx=0, instantiate_optional=False):
    """
    instantiate a node type under a parent node.
    nodeid and browse name of new node can be specified, or just namespace index
    If they exists children of the node type, such as components, variables and
    properties are also instantiated
    """
    rdesc = await _rdesc_from_node(parent, node_type)
    rdesc.TypeDefinition = node_type.nodeid

    if nodeid is None:
        nodeid = ua.NodeId(namespaceidx=idx)  # will trigger automatic node generation in namespace idx
    if bname is None:
        bname = rdesc.BrowseName
    elif isinstance(bname, str):
        bname = ua.QualifiedName.from_string(bname)

    nodeids = await _instantiate_node(
        parent.server,
        make_node(parent.server, rdesc.NodeId),
        parent.nodeid,
        rdesc,
        nodeid,
        bname,
        dname=dname,
        instantiate_optional=instantiate_optional)
    return [make_node(parent.server, nid) for nid in nodeids]


async def _instantiate_node(server,
                            node_type,
                            parentid,
                            rdesc,
                            nodeid,
                            bname,
                            dname=None,
                            recursive=True,
                            instantiate_optional=False):
    """
    instantiate a node type under parent
    """
    addnode = ua.AddNodesItem()
    addnode.RequestedNewNodeId = nodeid
    addnode.BrowseName = bname
    addnode.ParentNodeId = parentid
    addnode.ReferenceTypeId = rdesc.ReferenceTypeId
    addnode.TypeDefinition = rdesc.TypeDefinition

    if rdesc.NodeClass in (ua.NodeClass.Object, ua.NodeClass.ObjectType):
        addnode.NodeClass = ua.NodeClass.Object
        await _read_and_copy_attrs(node_type, ua.ObjectAttributes(), addnode)

    elif rdesc.NodeClass in (ua.NodeClass.Variable, ua.NodeClass.VariableType):
        addnode.NodeClass = ua.NodeClass.Variable
        await _read_and_copy_attrs(node_type, ua.VariableAttributes(), addnode)
    elif rdesc.NodeClass in (ua.NodeClass.Method,):
        addnode.NodeClass = ua.NodeClass.Method
        await _read_and_copy_attrs(node_type, ua.MethodAttributes(), addnode)
    elif rdesc.NodeClass in (ua.NodeClass.DataType,):
        addnode.NodeClass = ua.NodeClass.DataType
        await _read_and_copy_attrs(node_type, ua.DataTypeAttributes(), addnode)
    else:
        logger.error("Instantiate: Node class not supported: %s", rdesc.NodeClass)
        raise RuntimeError("Instantiate: Node class not supported")
    if dname is not None:
        addnode.NodeAttributes.DisplayName = dname

    res = (await server.add_nodes([addnode]))[0]
    added_nodes = [res.AddedNodeId]

    if recursive:
        parents = await get_node_supertypes(node_type, includeitself=True)
        node = make_node(server, res.AddedNodeId)
        for parent in parents:
            descs = await parent.get_children_descriptions(includesubtypes=False)
            for c_rdesc in descs:
                # skip items that already exists, prefer the 'lowest' one in object hierarchy
                if not await is_child_present(node, c_rdesc.BrowseName):
                    c_node_type = make_node(server, c_rdesc.NodeId)
                    refs = await c_node_type.get_referenced_nodes(refs=ua.ObjectIds.HasModellingRule)
                    if not refs:
                        # spec says to ignore nodes without modelling rules
                        logger.info("Instantiate: Skip node without modelling rule %s as part of %s",
                            c_rdesc.BrowseName, addnode.BrowseName)
                        continue
                        # exclude nodes with optional ModellingRule if requested
                    if refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):#not instantiate_optional or refs[0].nodeid == ua.NodeId(ua.ObjectIds.ModellingRule_Optional):
                        if instantiate_optional:
                            pass
                        else:
                            logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                                addnode.BrowseName)
                            continue
                    # if root node being instantiated has a String NodeId, create the children with a String NodeId
                    if res.AddedNodeId.NodeIdType is ua.NodeIdType.String:
                        inst_nodeid = res.AddedNodeId.Identifier + "." + c_rdesc.BrowseName.Name
                        nodeids = await _instantiate_node(
                            server,
                            c_node_type,
                            res.AddedNodeId,
                            c_rdesc,
                            nodeid=ua.NodeId(identifier=inst_nodeid, namespaceidx=res.AddedNodeId.NamespaceIndex),
                            bname=c_rdesc.BrowseName
                        )
                    else:
                        nodeids = await _instantiate_node(
                            server,
                            c_node_type,
                            res.AddedNodeId,
                            c_rdesc,
                            nodeid=ua.NodeId(namespaceidx=res.AddedNodeId.NamespaceIndex),
                            bname=c_rdesc.BrowseName
                        )
                    added_nodes.extend(nodeids)
    return added_nodes

```

oroulet:
I have not looked at that code for a long time!
But there was a patch newly supposed to fix it for deep children...

AndreasHeine:
now its like i expect it!
```python
 await instantiate(objects, my_object_type, ua.NodeId.from_string(f"ns={idx};s=Test2"),bname="2:Device0001", instantiate_optional=True)
    await instantiate(objects, my_object_type, ua.NodeId.from_string(f"ns={idx};s=Test3"),bname="2:Device0001", instantiate_optional=False)
```
first instance with optionals and second without!
Unbenannt3

AndreasHeine:
lets see if i broke something ;)

AndreasHeine:
beautifying...

AndreasHeine:
@oroulet 

> But there was a patch newly supposed to fix it for deep children...

link?

swamper123:
@AndreasHeine I think @oroulet meant this one in python-opcua:

https://github.com/FreeOpcUa/python-opcua/pull/1093

AndreasHeine:
i guess now it's like it supposed to be!

swamper123:
I think a test about creating different combinations of the same type and check if they work/exist could be nice as well.
A good reference could be [Figure 20](https://reference.opcfoundation.org/v104/Core/docs/Part3/6.4.4/) for such a case.

AndreasHeine:
you mean a inheriting custom types?

swamper123:
Yes this was the word I was missing. ^^

AndreasHeine:
maybe  we need to combine FreeOpcUa/python-opcua#1093 with FreeOpcUa/opcua-asyncio#242 to get all together 

@Nebukadneza please have a look on this!

```python
if not instantiate_optional and refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
```
maybe it has to be like:
```python
                    if refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
                        # instatiate optionals
                        if not instantiate_optional:
                            logger.info("Instantiate: Skip optional node %s as part of %s", c_rdesc.BrowseName,
                                addnode.BrowseName)
                            continue
```

swamper123:
@AndreasHeine What this placeholder is doing in the code, in the specs or both?

AndreasHeine:
to be honest in general... 🤷

swamper123:
[This is the Ref. for OptionalPlaceholder](https://reference.opcfoundation.org/v104/Core/docs/Part3/6.4.4/#6.4.4.5#6.4.4.5.5) in the specs.

AndreasHeine:
@oroulet any suggestion on that topic?

swamper123:
I guess I understand it know and give it a try to explain. _**If something is unclear or wrong, please correct me**_:



You got a `DeviceType`. You can instanciate different Devices out of it `(Device A,B,C`).

- Device A has just `Parameter A`
- Device B has `Parameter B` and `Parameter C` 
- Device C has `Parameter A` and `Parameter B`

So how would you try to represent these possebilities now in `DeviceType`?You can't just say "Device B just gets Parameter A and the content is None/zero/whatever".
Here this Placeholder thing may be useful. You say, that there can be different Parameters, depending on the instance.
And to make that clear, you add to `DeviceParameter` a reference to `OptionalPlaceholder`.

This OptionalPlaceholder denies a BrowserName of the Parametertype, because we don't know what Parameter/s it will be and says that here can be different versions of instances as well.

- - - - - 

The same with Figure 24 - the methods (maybe even a better example):



Depending on how you instanciate your device, the implementation of the DeviceMethods are different.
- Device A hasn't got a method `callable`.
- Device B has this method `callable`, but doesn't use arguments 
- Device C has this method `callable`, but with possible arguments 

So these PlaceHolders shall mean depending on the instance you may have a method, a method without arguments or a method with arguments. And to avoid that Device A may has anything to do with a method, this palceholder is used.
It is optional if there is a method or not and it is optional what arguments are necessary.

- - - 
To add abit of contrast, let's have a quick look at the mandatoryPlaceholder:

If we would replace in our second example `OptionalPlaceholder `with a `MandatoryPlaceholder`, Device A wouldn't be allowed, because it is mandatory that `callable` exist, but it doesn't specify in which kind of way it has to exist.

AndreasHeine:
your explaination is good but i dont understand the usecase why should it make sence to have a object-type as a blueprint if you can configer it as you want !? its one of this confusing parts of spec...

swamper123:
I would say it makes sense when everyone agrees "Yes this attributes represent a car. It needs 4 wheels, a brake, should have one or more  doors, a gas pedal and a motor.", but the rest is vendor stuff. What fuel shall it use, how many doors shall it have etc. Why should you recreate a car everytime from scratch, if you can just use such a blueprint? You will be on the safer side.

Such "blueprint"-objects may be an important thing for vendors as well, if they use a companion spec. Yes you can still create your own objects, in the end you can reach the same goal. Either via building my own object from scratch or use a blueprint and throw some parameters at it.
But for every Device you want to make, in every possible way, you have to touch each parameter once, to create a DeviceNr.1 Object. With a "blueprint" you can safe a bit more time.
But we should discuss that another time anywhere else, if this thing makes sense or not. 😃

swamper123:
@AndreasHeine "Mandatory"/"Optional"/"Maybe existing multiple times or once in some cases, maybe not"/"I don't know what and how many there are, but it is there!"

_ExposesItsArray_ is a modelling rule as well, but I think we haven't implemented that either.

AndreasHeine:
guess we have to merge this as first step (to get the smallest working feature) and we need to have a discussion on that topic for me its a "egg-laying wool milk sow" (maybe bad translation but...)

oroulet:
I know nothing on that topic but pr looks good. I can merge if you want

AndreasHeine:
go ahead!

Nebukadneza:
Hi there,

just a quick ping — our team is currently deep into Fully-Inherited-Instance-Declaration land, and the ping by @AndreasHeine is probably a fair point. If it’s feasible to wait until tomorrow with merging, we could have another look, and maybe help to sched some more light — as the topic is indeed a very complex one.

Thanks and sorry for the delay,
-Dario

AndreasHeine:
no problem!

AndreasHeine:
@Nebukadneza anything new / any suggestions / ideas?

Nebukadneza:
Hi, and sorry for the delay,

> Maybe @Nebukadneza is working on a testcase?

Not yet, but the topic is inherently complex, so it’s on my mind-roadmap to contribute a few tests down the road. For the record: I do feel bad for contributing the fixes without an accompanying test. Sorry for that!

> maybe it has to be like:
> 
> ```python
>                     if refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
>                         if not instantiate_optional:
>                             continue
> ```

To be honest, I currently don’t see the improvement. Maybe I’m a little blind right now ^_^. The combination of these 2 `if`s is, IMO, precisely equivalent to an `and`, and says something like: »if we are told no to instantiate optionals, and the modelling rule is one of the 2 optional varieties, `continue` this loop iteration«. Correct me if I’m wrong, but i feel this is what we want here?

> Edit:
> dont know what this placeholder is doing yet... try to find out!

@swamper123 provided a nice explanation, that to me reads correct — thanks!
One thing I would like to add is that the 4 ModellingRules can be read has:
* Optional: MAY have exactly one
* OptionalPlaceholder: MAY have zero or more
* Mandatory: MUST have one
* MandatoryPlaceholder: MUST have one or more

So when skipping optional parts, it’s (AFAIK) correct to only skip over the 2 Optional types, but instantiate exactly one of the Mandatory ones. If there’s a MandatoryPlaceholder, the behavior is a bit fuzzy — we should instantiate one, but the name would need to be users-supplied. Any extra ones, it’s the users responsibility to instantiate by her/himself.


>  But we should discuss that another time anywhere else, if this thing makes sense or not. smiley

While I agree, maybe I can offer one more possible use-case for @AndreasHeine ^_^. In fact, it’s named Modelling**RULE** for a reason: It can not only be used to (semi-)automatically create instances from »blueprints«, but it can also serve as a kind of type-checker for a instance. Since it defines how an object   of type `SomeType` has to look like, it makes it possible to automatically check if Objects coming from users/other systems/… are valid or not.



> guess we have to merge this as first step

After reading discussion and code, it looks reasonable to me. Just 3 remarks:
* Pulling the one `if` into 2 isn’t necessarily required — it may be valid as a question of style, though
* By changing `instantiate_optional=True` to `instantiate_optional=False` in the toplevel function, we are making a breaking API change which could bite users, and needs a warning in release notes. I’m not sure why that change is necessary.
* While you’re touching this file here anyways, maybe you would want to forward-port https://github.com/FreeOpcUa/python-opcua/pull/1088/files too? This fixes the behavior regarding `instantiate_optional` on the recursive levels below direct top-level.

Thanks, and sorry for the long ramblings :sweat_smile: :sweat_drops: ,
-Dario

AndreasHeine:
thanks for the light 😀 now i need some time to boil it down... just as a ping back!

AndreasHeine:
> Pulling the one if into 2 isn’t necessarily required — it may be valid as a question of style, though

of course it a question of style its kind of long after adding ModellingRule_OptionalPlaceholder...

```python
if refs[0].nodeid in (ua.NodeId(ua.ObjectIds.ModellingRule_Optional), ua.NodeId(ua.ObjectIds.ModellingRule_OptionalPlaceholder)):
```

> By changing instantiate_optional=True to instantiate_optional=False in the toplevel function, we are making a breaking API change which could bite users, and needs a warning in release notes. I’m not sure why that change is necessary.

this brought me thinking... it boils down to the question why it was chosen to be default true? i could live with true as default but then we/i should add "instatiate_optional" to "add_object" and "create_object" because they internaly use "instantiate" and dont have that parameter. maybe i stumble over that "optional" word to much but for me "optionals are an option" sorry for that phrase...^^

> While you’re touching this file here anyways, maybe you would want to forward-port https://github.com/FreeOpcUa/python-opcua/pull/1088/files too? This fixes the behavior regarding instantiate_optional on the recursive levels below direct top-level.

Done!

oroulet:
I did not follow the discussion at all but the code looks OK so I can merge if you want

AndreasHeine:
I give up regarding the default value (it would be a breaking change and maybe its not wise to do so thx to @Nebukadneza) there is still a problem for the placeholder and the instantiate behavior (it is pretty abstract problem because its not realy easy to tell the difference between instantiate a specific optional and only this one or not, without some more information as input ... read this part of spec multiple times)... 
i added stuff which is in python-opcua and missing in asyncua! (at least progress, small but, present)
in a later PR i will pass instantiate_optional as an "optional parameter" in the high level api so you can choose in add_obect to not instantiate optionals just to get more flexability

so that's basically it! feel free to merge.

AndreasHeine:
@oroulet !? Ok or not?

PR #244, opened 2020-07-20
Fix smaller issues

swamper123:
Add pretty argument to docstrings
Fix warning in Travis tests about invalid escape sequences
Add new line to be PEP8 conform
Remove unused import
Remove doubled expressions
remove redundant parenthese

### DIFF
--- a/asyncua/common/xmlexporter.py
+++ b/asyncua/common/xmlexporter.py
@@ -115,7 +115,7 @@ class XmlExporter:
         Write the XML etree in the exporter object to a file
         Args:
             xmlpath: string representing the path/file name
-
+            pretty: add spaces and newlines, to be more readable
         Returns:
         """
         # try to write the XML etree to a file
--- a/schemas/generate_model.py
+++ b/schemas/generate_model.py
@@ -448,7 +448,7 @@ def add_basetype_members(model):
         emptystruct = False
         if len(struct.fields) == 0:
             emptystruct = True
-        if struct.basetype in ('ExtensionObject'):
+        if struct.basetype in 'ExtensionObject':
             struct.basetype = None
             continue
         base = model.get_struct(struct.basetype)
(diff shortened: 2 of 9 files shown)


schemas/generate_model.py
```diff
@@ -448,7 +448,7 @@ def add_basetype_members(model):
         emptystruct = False
         if len(struct.fields) == 0:
             emptystruct = True
-        if struct.basetype in ('ExtensionObject'):
+        if struct.basetype in 'ExtensionObject':
```
oroulet, line 451: this is probabky not correct. It is goign to compare to every char in string

schemas/generate_protocol_python.py
```diff
@@ -151,7 +151,7 @@ def generate_struct_code(self, obj):
                 self.write("self.Encoding = 1")
             elif field.uatype == obj.name:  # help!!! selv referencing class
                 self.write("self.{} = None".format(field.name))
-            elif not obj.name in ("ExtensionObject") and field.name == "TypeId":  # and ( obj.name.endswith("Request") or obj.name.endswith("Response")):
+            elif not obj.name in "ExtensionObject" and field.name == "TypeId":  # and ( obj.name.endswith("Request") or obj.name.endswith("Response")):
```
oroulet, line 154: same here. Either use "==" or "in some_list"

swamper123: Same here. I will add an coma, so that it is more visible that other types could be added in future.

asyncua/common/xmlexporter.py
```diff
@@ -115,7 +115,7 @@ async def write_xml(self, xmlpath, pretty=True):
         Write the XML etree in the exporter object to a file
         Args:
             xmlpath: string representing the path/file name
-
+            pretty: add spaces and newlines, to be more readable
```
oroulet, line 118: Why make the string start with r? I never remember...

swamper123: By adding `'r'` as prefix, you tell the interpreter that this string shall not be interpreted as Pythonstring, where backslash characters may cause trouble then. It would be an invalid escape sequence, which is throwing a DeprecationWarning atm, but in later Versions it throws an SyntaxError.
E.g. r"\n" is something else than "\n".

PR #248, opened 2020-07-23
Raise ConnectionError in send_request if state is not OPEN

### LINKED ISSUE #241: UASocketProtocol transport should not be set to None silently on disconnect
When disconnect happen `UASocketProtocol.transport` is silently set to `None` and I get `AttributeError`s instead of something more appropriate like `IOError`:

```python
AttributeError: 'NoneType' object has no attribute 'write'
```

I need to check if I am connected like that: `client.uaclient.protocol.transport is not None`.
Suggestion: throw `IOError` (or `ConnectionError`) when trying to use transport with broken connection.

HarrySky:
Fixes #241 . Simple `try/catch` block that will raise `ConnectionError` in `send_request` if connection was lost

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -137,10 +137,17 @@ class UASocketProtocol(asyncio.Protocol):
         Returns response object if no callback is provided.
         """
         timeout = self.timeout if timeout is None else timeout
-        data = await asyncio.wait_for(
-            self._send_request(request, timeout, message_type),
-            timeout if timeout else None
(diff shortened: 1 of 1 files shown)


oroulet:
when is that necessary?

HarrySky:
@oroulet , see issue #241 , right now it raises `AttributeError: 'NoneType' object has no attribute 'write'`, which is not the ideal error for connection issues.

oroulet:
Ok

PR #261, opened 2020-08-17
fix missing async in server-callback.py

AndreasHeine:


AndreasHeine:
how does this ever work? ;)

oroulet:
wating for the other callback PR to be merged. Then this one must be updated too

AndreasHeine:
ok

oroulet:
@AndreasHeine nowit is merged. you need to update some names I think

AndreasHeine:
https://github.com/FreeOpcUa/opcua-asyncio/pull/229/files

@oroulet doesn't look like the api has changed, i would add the three new callbacktypes!

PR #271, opened 2020-08-24
bugfix for memory leak

JoeyFaulkner:
implemented @oroulet comments on the other pull request, as much as i understood them https://github.com/JoeyFaulkner/opcua-asyncio/pull/3

### DIFF
--- a/asyncua/server/binary_server_asyncio.py
+++ b/asyncua/server/binary_server_asyncio.py
@@ -18,7 +18,7 @@ class OPCUAProtocol(asyncio.Protocol):
     Instantiated for every connection.
     """
 
-    def __init__(self, iserver: InternalServer, policies, clients, protocol_tasks):
+    def __init__(self, iserver: InternalServer, policies, clients, closing_tasks):
         self.peer_name = None
         self.transport = None
         self.processor = None
@@ -26,7 +26,7 @@ class OPCUAProtocol(asyncio.Protocol):
         self.iserver: InternalServer = iserver
         self.policies = policies
         self.clients = clients
-        self.protocol_tasks = protocol_tasks
+        self.closing_tasks = closing_tasks
         self.messages = asyncio.Queue()
         self._task = None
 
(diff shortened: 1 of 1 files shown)


JoeyFaulkner:
ok i think this error is out of my understanding https://travis-ci.org/github/FreeOpcUa/opcua-asyncio/jobs/720599077#L919

asyncua/server/binary_server_asyncio.py
```diff
@@ -136,19 +135,29 @@ async def start(self):
...
             await self._server.wait_closed()
+
+    async def _await_closing_tasks(self):
+        await asyncio.sleep(10)
+        while self.closing_tasks:
+            task = self.closing_tasks.pop()
+            try:
+                await task
+            except Exception:
+                logger.exception("Unexpected crash in BinaryServer._await_closing_tasks")
```
swamper123, line 163: There is an empty line missing here.

swamper123:
This test "Run[s] an example OPC-UA server. By importing xml definition and using uawrite command line, it is even possible to expose real data using this server" within 2 seconds. After that, a SIGINT is send to the server and it has to shutdown in 2 seconds.

And there something is not working well...and I can't say exactly what. I would add a broad except block beside OSError and KeyboardInterrupt into this, perhaps you may catch an error with a better description.

(Perhaps these sleeps could be async as well, but I don't think that these are causing problems.)

asyncua/server/binary_server_asyncio.py
```diff
@@ -136,19 +135,29 @@ async def start(self):
...
             if isinstance(result, Exception):
                 self.logger.error(f"An error ocurred while closing a transport: {result}")
-        self.protocol_tasks = []
+        self.closing_tasks = []
 
         if self._server:
             self.iserver.loop.call_soon(self._server.close)
             await self._server.wait_closed()
+
+    async def _await_closing_tasks(self):
```
swamper123, line 156: If I understand this right, this method is just called at the start of a server? Or should it be call repeatadly?

swamper123 (changes requested):

oroulet:
Travis seemsto be down. What is the issue? It looks like master on my PC is broken too. I get a strange (RuntimeError: There is no current event loop....) No ideas where it comes from..

swamper123:
grafik

The process returns with 2 in the test.

oroulet:
travis is broken today. We need to wait...

oroulet:
see the other pull request

PR #272, opened 2020-08-24
Joey faulkner bugfix/server memory leak

oroulet:
@JoeyFaulkner

### DIFF
--- a/asyncua/server/binary_server_asyncio.py
+++ b/asyncua/server/binary_server_asyncio.py
@@ -160,6 +160,9 @@ class BinaryServer:
             task = self.closing_tasks.pop()
             try:
                 await task
+            except asyncio.CancelledError:
+                # this means a stop request has been sent, it should not be catched
+                raise
             except Exception:
                 logger.exception("Unexpected crash in BinaryServer._await_closing_tasks")
         if recursive:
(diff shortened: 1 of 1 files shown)


swamper123:
@oroulet asyncio.create_task is a python3.7+ method. You need to use ensure_future for 3.6.

oroulet:
or drop support for 3.6 ;-) i hatethat name ensure_future. it is impossible to understand

asyncua/server/binary_server_asyncio.py
```diff
@@ -136,19 +136,32 @@ async def start(self):
             self.hostname = sockname[0]
             self.port = sockname[1]
         self.logger.info('Listening on %s:%s', self.hostname, self.port)
+        self.cleanup_task = asyncio.create_task(self._await_closing_tasks())
```
swamper123, line 139: Aren't you running await_closing_tasks() once as well?
With asyncio.create_task(), you are scheduling this Task in the eventloop. 
In case there aren't any closing_tasks, this will return None and stay None as Future result.

swamper123:
@oroulet I bet, I would be the last one in this repo missing python 3.6 :^)

asyncua/server/binary_server_asyncio.py
oroulet: _await_closing_task nevers returns. if you think it does, let me know

oroulet:
@JoeyFaulkner does that patch solves your issue? it seems to work

JoeyFaulkner:
will use this and check it out

JoeyFaulkner:
seems fine!

PR #278, opened 2020-09-02
accidentally wrote PostWrite instead of PreWrite

JoeyFaulkner:


JoeyFaulkner:
merging this because its tiny and a bug, let me know if this is bad practice and i wont do it again :)

swamper123:
I would say it is ok to fix such smaller stuff faults, especially if they are self made. :^)
So IMO please do it again, in case it happens. 😄

oroulet:
Fine for me. I do that kind of thing sometimes

PR #283, opened 2020-09-08
[ Fix ] test_tool quick fix

ZuZuD:
Fix the test ```test_that_server_can_be_invoked_without_internal_error``` failing on TravisCI since tests are executed from the repository root directory, see #212. 

## Before
```
> pytest tests/test_tools.py
================================================ test session starts =================================================
platform linux -- Python 3.7.5+, pytest-5.3.2, py-1.8.1, pluggy-0.13.1
rootdir: /home/prigentj/github/opcua-asyncio, inifile: pytest.ini
plugins: asyncio-0.10.0, cov-2.8.1, mock-2.0.0
collected 10 items

tests/test_tools.py .........F                                                                                 [100%]

====================================================== FAILURES ======================================================
_______________________________ test_that_server_can_be_invoked_without_internal_error _______________________________

    def test_that_server_can_be_invoked_without_internal_error():
>       proc = subprocess.Popen(['uaserver'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)

tests/test_tools.py:22:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/fbcode/platform007/lib/python3.7/subprocess.py:800: in __init__
    restore_signals, start_new_session)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <subprocess.Popen object at 0x7f018f8f5c90>, args = ['uaserver'], executable = b'uaserver', preexec_fn = None
close_fds = True, pass_fds = (), cwd = None, env = None, startupinfo = None, creationflags = 0, shell = False
p2cread = -1, p2cwrite = -1, c2pread = 12, c2pwrite = 13, errread = 14, errwrite = 15, restore_signals = True
start_new_session = False

[...]
>               raise child_exception_type(errno_num, err_msg, err_filename)
E               FileNotFoundError: [Errno 2] No such file or directory: 'uaserver': 'uaserver'

/usr/local/fbcode/platform007/lib/python3.7/subprocess.py:1551: FileNotFoundError
[...]

PR #284, opened 2020-09-09
[ Cli tools ] Fix uaserver and tests

ZuZuD:
This commit adds the following:

* A server fixture which runs in a dedicated thread to handle OPC-UA requests properly (existing server fixture just accept TCP connections but can't handle hello since they are not running).

* We then use it to enhance the CLI tests and set more consistent exit codes across the cli (following the comment on #283).

* Fix the ```uaserver``` cli tool, its event_loop was stuck in non-async sleep preventing it from accepting requests.

### DIFF
--- a/asyncua/tools.py
+++ b/asyncua/tools.py
@@ -29,7 +29,7 @@ if sys.version_info.minor >= 7:
         return asyncio.run(coro)
 else:
     def run(coro):
-        loop = asyncio.get_event_loop()
+        loop = asyncio.new_event_loop()
         return loop.run_until_complete(coro)
 
 
--- a/asyncua/tools.py
+++ b/asyncua/tools.py
@@ -146,8 +146,12 @@ async def _uaread():
             print(attr.Value)
         else:
             print(attr)
+    except Exception as e:
+        print(e)
+        sys.exit(1)
     finally:
         await client.disconnect()
+    sys.exit(0)
 
 
 def _args_to_array(val, array):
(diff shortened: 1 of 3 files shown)


asyncua/tools.py
```diff
@@ -577,18 +585,19 @@ def multiply(parent, x, y):
...
         await server.init()
         async with server:
+
             if args.shell:
                 embed()
             elif args.populate:
                 count = 0
                 while True:
-                    time.sleep(1)
+                    await asyncio.sleep(1)
```
oroulet, line 594: that one  was ugly ;-)

asyncua/tools.py
```diff
@@ -29,7 +29,7 @@ def run(coro):
         return asyncio.run(coro)
 else:
     def run(coro):
-        loop = asyncio.get_event_loop()
+        loop = asyncio.new_event_loop()
```
oroulet, line 32: whatis the idea here?

tests/conftest.py
```diff
@@ -7,6 +7,7 @@
 
 from .test_common import add_server_methods
 from .util_enum_struct import add_server_custom_enum_struct
+from threading import Thread
```
swamper123, line 10: Why do you want to create a server with a thread? I would recommened to run a second process, so that there is a real parallel execution possible or using the Executor possebilities asyncio offers (to be on a threadsafe side).

asyncua/tools.py
ZuZuD: To mimic ```asyncio.run()```  which create a new event loop (see [cpython/asyncio](https://github.com/python/cpython/blob/3.8/Lib/asyncio/runners.py#L8-L51)) for py < 3.7. Required for the testing.

tests/conftest.py
ZuZuD: The intent was to keep the server tied to the fixture to stop it properly via finalizer (done here not via a thread variable ```do_run````).

ZuZuD:
Some fine-tuning for pypy3, tests are better looking now.

Issue #286, opened 2020-09-15
git fetch fails because of branch name


cbergmiller:
I can't fetch this repository any more and i think it is due to the use of the `>` character in a recently introduced branch.
Anyone else having similar trouble?

```
git fetch
error: cannot lock ref 'refs/remotes/origin/Fix-history_sql.py-emitting-node-is-not-the-source-->-sql-error': Unable to create 'PycharmProjects/opcua-asyncio/.git/refs/remotes/origin/Fix-history_sql.py-emitting-node-is-not-the-source-->-sql-error.lock': Invalid argument
From https://github.com/FreeOpcUa/opcua-asyncio
 ! [new branch]      Fix-history_sql.py-emitting-node-is-not-the-source-->-sql-error -> origin/Fix-history_sql.py-emitting-node-is-not-the-source-->-sql-error  (unable to update local ref)
```

swamper123:
![grafik](https://user-images.githubusercontent.com/24408657/93185683-04df5680-f73e-11ea-8514-2c844bc988ca.png)
This realy seems to be an error if you are using console commands (tested on Windows and Linux, same result via cmd). Perhaps UI stuff may handle it in another way.

AndreasHeine:
@swamper123 done!

swamper123:
@AndreasHeine I think it is still there. Have you pushed it?
![grafik](https://user-images.githubusercontent.com/24408657/93299615-9eb60a80-f7f5-11ea-916a-a35ab1bb188f.png)

AndreasHeine:
... some strange things happend must delete the branch!

cbergmiller:
Works again for me, thanks!

AndreasHeine:
sorry about that!

swamper123:
No worries, can happen. :)

PR #287, opened 2020-09-16
Update history_sql.py

AndreasHeine:
recreated after branch deleted: because of #286


oroulet:
should I merge?

AndreasHeine:
yes, same as in python-opcua...

PR #291, opened 2020-09-21
Check all zero ID instead of single zero in ID.

### LINKED ISSUE #290: GUID with single zero byte misidentified as null identifier
https://github.com/FreeOpcUa/opcua-asyncio/blob/75adf4f885a71ef3b8b49711dc12a6ed8dab87d8/asyncua/ua/uatypes.py#L332

It matches a single byte instead of all of them being zero, e. g. `0032b2d5-c219-4524-8df5-5490a2109e9b`.
The consequence is that instead of the supplied GUID a new integer identifier is generated.

I can't say for sure that this is not an illegal node identifier, but

> Each value is formatted as a hexadecimal number with padded zeros.
(https://reference.opcfoundation.org/v104/Core/docs/Part6/5.1.3/)

seems to suggest multiple (especially leading) zeros can exist in a GUID.

I also found this:

> A Node in the AddressSpace shall not have a null as its NodeId.
(https://reference.opcfoundation.org/v104/Core/docs/Part3/8.2.4/)

but I think this references an all-zeros identifier, not containing a single zero byte.
Other than that there seems to be no restrictions.

My suggestion (which I'll gladly submit if you agree) is to replace the regex check with a check for an all zero bytes array.

```python
if self.NodeIdType == NodeIdType.Guid and self.Identifier.bytes == b'\00'*16:
```

Randelung:
Fixes #290.

### DIFF
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -329,7 +329,7 @@ class NodeId(object):
     def has_null_identifier(self):
         if not self.Identifier:
             return True
-        if self.NodeIdType == NodeIdType.Guid and re.match(b'\00+', self.Identifier.bytes):
+        if self.NodeIdType == NodeIdType.Guid and self.Identifier.bytes == b'\00'*16:
             return True
         return False
 
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -307,9 +307,9 @@ class NodeId(object):
             else:
                 raise UaError("NodeId: Could not guess type of NodeId, set NodeIdType")
         else:
-            self.check_type_compatibility()
+            self.check_identifier_type_compatibility()
 
-    def check_type_compatibility(self):
+    def check_identifier_type_compatibility(self):
         '''
         Check whether the given identifier can be interpreted as the given node identifier type.
         '''
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -308,7 +308,7 @@ class NodeId(object):
                 raise UaError("NodeId: Could not guess type of NodeId, set NodeIdType")
         else:
             if not (isinstance(self.Identifier, int) and self.NodeIdType in [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte] or
-                    isinstance(self.Identifier, str) and self.NodeIdType is NodeIdType.String or
+                    isinstance(self.Identifier, str) and self.NodeIdType in [NodeIdType.String, NodeIdType.ByteString] or
                     isinstance(self.Identifier, bytes) and self.NodeIdType in [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte] or
                     isinstance(self.Identifier, uuid.UUID) and self.NodeIdType is NodeIdType.Guid):
                 raise UaError("NodeId: Incompatible identifier and nodeidtype")
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -321,9 +321,8 @@ class NodeId(object):
         ]
         for identifier, valid_node_types in valid_type_combinations:
             if isinstance(self.Identifier, identifier) and self.NodeIdType in valid_node_types:
-                break
-        else:
-            raise UaError(f"NodeId of type {self.NodeIdType} has an incompatible identifier {self.Identifier} of type {type(self.Identifier)}")
+                return
+        raise UaError(f"NodeId of type {self.NodeIdType} has an incompatible identifier {self.Identifier} of type {type(self.Identifier)}")
 
     def __eq__(self, node):
         return isinstance(node, NodeId) and self.NamespaceIndex == node.NamespaceIndex and self.Identifier == node.Identifier
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -267,7 +267,7 @@ class NodeId(object):
     Args:
         identifier: The identifier might be an int, a string, bytes or a Guid
         namespaceidx(int): The index of the namespace
-        nodeidtype(NodeIdType): The type of the nodeid if it cannor be guess or you want something
+        nodeidtype(NodeIdType): The type of the nodeid if it cannot be guess or you want something
         special like twobyte nodeid or fourbytenodeid
 
 
@@ -306,6 +306,12 @@ class NodeId(object):
                 self.NodeIdType = NodeIdType.Guid
             else:
                 raise UaError("NodeId: Could not guess type of NodeId, set NodeIdType")
+        else:
+            if not (isinstance(self.Identifier, int) and self.NodeIdType in [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte] or
+                    isinstance(self.Identifier, str) and self.NodeIdType is NodeIdType.String or
+                    isinstance(self.Identifier, bytes) and self.NodeIdType in [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte] or
+                    isinstance(self.Identifier, uuid.UUID) and self.NodeIdType is NodeIdType.Guid):
+                raise UaError("NodeId: Incompatible identifier and nodeidtype")
 
     def __eq__(self, node):
         return isinstance(node, NodeId) and self.NamespaceIndex == node.NamespaceIndex and self.Identifier == node.Identifier
@@ -329,7 +335,7 @@ class NodeId(object):
     def has_null_identifier(self):
         if not self.Identifier:
             return True
-        if self.NodeIdType == NodeIdType.Guid and self.Identifier.bytes == b'\00'*16:
+        if self.NodeIdType is NodeIdType.Guid and self.Identifier.int == 0:
             return True
         return False
 
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -307,11 +307,17 @@ class NodeId(object):
             else:
                 raise UaError("NodeId: Could not guess type of NodeId, set NodeIdType")
         else:
-            if not (isinstance(self.Identifier, int) and self.NodeIdType in [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte] or
-                    isinstance(self.Identifier, str) and self.NodeIdType in [NodeIdType.String, NodeIdType.ByteString] or
-                    isinstance(self.Identifier, bytes) and self.NodeIdType in [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte] or
-                    isinstance(self.Identifier, uuid.UUID) and self.NodeIdType is NodeIdType.Guid):
-                raise UaError("NodeId: Incompatible identifier and nodeidtype")
+            valid_type_combinations = [
+                (int, [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte]),
+                (str, [NodeIdType.String, NodeIdType.ByteString]),
+                (bytes, [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte]),
+                (uuid.UUID, [NodeIdType.Guid])
+            ]
+            for identifier, valid_node_types in valid_type_combinations:
+                if isinstance(self.Identifier, identifier) and self.NodeIdType in valid_node_types:
+                    break
+            else:
+                raise UaError(f"NodeId of type {self.NodeIdType} has an incompatible identifier {self.Identifier} of type {type(self.Identifier)}")
 
     def __eq__(self, node):
         return isinstance(node, NodeId) and self.NamespaceIndex == node.NamespaceIndex and self.Identifier == node.Identifier
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -307,17 +307,23 @@ class NodeId(object):
             else:
                 raise UaError("NodeId: Could not guess type of NodeId, set NodeIdType")
         else:
-            valid_type_combinations = [
-                (int, [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte]),
-                (str, [NodeIdType.String, NodeIdType.ByteString]),
-                (bytes, [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte]),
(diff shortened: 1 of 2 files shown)


Randelung:
I noticed that the test uses a NodeId with a byte identifier but specifies GUID as node ID type. Checking for null or zero identifier throws an exception since `<byte array>.bytes` doesn't exist. The same would be the case for all identifier types when mixed with GUID as node ID type since an actual check for identifier data type isn't being performed.

It's not important for the test, but it would probably be smart to either check identifier data type during `__init__` (instead of blindly relying on the type parameter) or during the null or zero tests.

swamper123:
There isn't a compact tool I know to check a Node Identifier, but you can check what Node.NodeId.NodeIdType is:
 https://github.com/FreeOpcUa/opcua-asyncio/blob/75adf4f885a71ef3b8b49711dc12a6ed8dab87d8/asyncua/ua/uatypes.py#L254 

In case there is no GUID-Node, a GUID-Node has to be made perhaps ?!

Randelung:
I think I didn't explain well enough, let me try again.

`NodeId` objects have `is_null` and `has_null_identifier` checks. It's possible to specify the `NodeIdType` in the `NodeId` instance, which will jump over the ID type guessing in `__init__` and just believe the parameter. It's therefore possible to lie to the object about its node identifier data type (you can pass a `byte` array as identifier and `ua.NodeIdType.Guid` as `nodeidtype`).

The latter method is being used in `_add_node`:
https://github.com/FreeOpcUa/opcua-asyncio/blob/3636b610ae17a4c318730399150431afb8a81d5e/asyncua/server/address_space.py#L223
It will fail if the NodeId is being lied to like this.

I'm not sure if it needs to be possible to specify `nodeidtype` during initialization for a use case, but the `bytes` access in `has_null_identifier` should be guarded. Otherwise node creation will fail:

```python
$ python3 -m asyncio
>>> import asyncio
>>> import asyncua
>>> server = asyncua.Server()
>>> await server.nodes.objects.add_variable(asyncua.ua.NodeId(b'\01', nodeidtype=asyncua.ua.NodeIdType.Guid), 'node', 0)
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 388, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/common/node.py", line 665, in add_variable
    return await create_variable(self, nodeid, bname, val, varianttype, datatype)
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/common/manage_nodes.py", line 106, in create_variable
    await _create_variable(parent.server, parent.nodeid, nodeid, qname, var, datatype=datatype, isproperty=False)
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/common/manage_nodes.py", line 274, in _create_variable
    results = await server.add_nodes([addnode])
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/server/internal_session.py", line 128, in add_nodes
    return self.iserver.node_mgt_service.add_nodes(params, self.user)
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/server/address_space.py", line 206, in add_nodes
    results.append(self._add_node(item, user))
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/server/address_space.py", line 223, in _add_node
    if item.RequestedNewNodeId.has_null_identifier():
  File "/home/randelung/.local/lib/python3.8/site-packages/asyncua/ua/uatypes.py", line 339, in has_null_identifier
    if self.NodeIdType == NodeIdType.Guid and self.Identifier.bytes == b'\00'*16:
AttributeError: 'bytes' object has no attribute 'bytes'
```

So I would suggest one or more of several things to do:
- Add a type guard to the `has_null_identifier` method
- Replace the `nodeId.NodeIdType` check with a type guard since the `NodeIdType` is (apparently by design) unreliable
- Check whether the `nodeidtype` parameter for `NodeId` is useful or not and remove it if applicable to remove the error case before it even arises
- Create an extra ticket discussing what to do

I realize this comment section is a bad place to go into detail, but here we are.

asyncua/ua/uatypes.py
```diff
@@ -329,7 +329,7 @@ def is_null(self):
     def has_null_identifier(self):
         if not self.Identifier:
             return True
-        if self.NodeIdType == NodeIdType.Guid and re.match(b'\00+', self.Identifier.bytes):
+        if self.NodeIdType == NodeIdType.Guid and self.Identifier.bytes == b'\00'*16:
```
oroulet, line 332: I know next to nothing about guid. Are you sure they are always 16 bytes?Isn't there a way to check generically that ALL bytes are 0? just wondering..

oroulet: all([i==0 for i in bytes])?

Randelung: https://reference.opcfoundation.org/v104/Core/docs/Part6/5.1.3/ Yeah, GUID are always 16 bytes.
The uuid.UUID type that is used also reinforces this. https://docs.python.org/3/library/uuid.html

Since we're using the `UUID.bytes` method, how about https://docs.python.org/3/library/uuid.html#uuid.UUID.int instead?
```python
if type(self.Identifier) == uuid.UUID and self.Identifier.int == 0:
```

Randelung:
```suggestion
        if type(self.Identifier) == uuid.UUID and self.Identifier.int == 0:
```

oroulet: do we need to test for type?
the identifier MUST be an uuid.UUID, isn't it? 
In your proposition over the answer might be wrong if the type is not waht expected... Maybe we should test for type at the line over and raise omUaError if it is not the correct type
```
if not insinstance(self.identifier, uuid.UIID):
    raise RuntimeError(XXX) or UaError not sure what is bes 
if self.NodeIdType == NodeIdType.Guid and type(self.Identifier) == uuid.UUID and self.Identifier.int == 0:
```

oroulet: also in python isinstance is better than type(xx) == 11

Randelung: I don't know if you see this thread in a different context than the pull request I submitted, but I outlined the issue that requires the type check here: https://github.com/FreeOpcUa/opcua-asyncio/pull/291#issuecomment-696204533
The gist of it is that the identifier is not necessarily a UUID when the `NodeIdType` is `Guid`.
Whether that is supposed to be an allowed situation is beyond my current knowledge of this library. :D
There is a possibility that the `NodeIdType` parameter isn't even needed in the first place if the parameter and identifier type need to always be synchronous. At least I think the parameter probably shouldn't be settable arbitrarily, i. e. the checks you proposed should be in `__init__` to see if the parameter matches the identifier type, not in this method.

```suggestion
        if self.NodeIdType == NodeIdType.Guid and isinstance(self.Identifier, uuid.UUID) and self.Identifier.int == 0:
```

oroulet: OK I checked the code. Identifier must be None or UUID so you can just remove the type test

oroulet: then I can merge

Randelung: I mean I have an explicit case where that is not true. The test_unit.py test module actually uses this exact setup:
https://github.com/FreeOpcUa/opcua-asyncio/blob/9f3dd531397db854f2e1d2a59ff04f4231cab418/tests/test_unit.py#L408
So it is very possible for the identifier type and parameter to not agree, and if the resulting node ID it's used in a `create_node` the server dies - as outlined in the comment I linked.

In `__init__`, a type guess is being done only if the NodeIdType parameter isn't explicitly passed, everything else is just accepted blindly. You can pass any combination of identifier and NodeIdType, as long as it's not `None`.

That's why I think either check for disagreement at `__init__` (in which case it might show the NodeIdType param to be useless since identifier types are already unique) and get rid of the type check here or allow discrepancies but then definitely do a type check.

oroulet: Having a type check here and ignoring the case when it is not the expected type is definitely wrong.

And yes we can add type check at init but then we should check for all possible types not only the uuid case

Randelung:
I've updated the PR to include type checks in `__init__` and not in `has_null_identifier`. I tried to infer what types would match which NodeIdType; if I missed some, feel free to suggest.

!!! THIS PR WILL BREAK SOME TESTS !!!

One example is the test case linked above that uses a bytes identifier and NodeIdType.Guid.

I was able to start my server project and server-client projects, though, so sanity check is fine (admittedly only using automatic numeric and GUIDs)

asyncua/ua/uatypes.py
```diff
@@ -306,6 +306,12 @@ def __init__(self, identifier=None, namespaceidx=0, nodeidtype=None):
                 self.NodeIdType = NodeIdType.Guid
             else:
                 raise UaError("NodeId: Could not guess type of NodeId, set NodeIdType")
+        else:
+            if not (isinstance(self.Identifier, int) and self.NodeIdType in [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte] or
+                    isinstance(self.Identifier, str) and self.NodeIdType in [NodeIdType.String, NodeIdType.ByteString] or
+                    isinstance(self.Identifier, bytes) and self.NodeIdType in [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte] or
+                    isinstance(self.Identifier, uuid.UUID) and self.NodeIdType is NodeIdType.Guid):
+                raise UaError("NodeId: Incompatible identifier and nodeidtype")
```
oroulet, line 314: Sorry to come with one more comment but for readablity I would prefer to split that looooong test in several ones the other way around:
and we must return a more precise error message otherwise that stuff will be impossible to debug
if self.NodeIdType in [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte] and not isintance(self.identifier, int):
     raise UaError(f"NodeId of type {self.NodeTypeId} has an incompatible identifier {self.Identifier} of type {type(self.identifier}")
elif self.NodeTypeId in [xxxxxx]:
      raise xxxxx
etc...

Randelung: No worries, that's why we're here!
I'm afraid your suggestion is broken since e. g. a TwoByte type can be represented by an integer or a byte[] or a ByteString, so the check needs to be inverted to find a single correct match instead of fulfilling all of the checks.
How about a more generic approach altogether? Something like
```python
validTypeCombinations = [
    (int, [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte]),
    (str, [NodeIdType.String, NodeIdType.ByteString]),
    (bytes, [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte]),
    (uuid.UUID, [NodeIdType.Guid])
]
for combination in validTypeCombinations:
    if isinstance(self.identifier, combination[0]) and self.NodeIdType in combination[1]):
        break
else:
    raise UaError(f"NodeId of type {self.NodeIdType} has an incompatible identifier {self.Identifier} of type {type(self.Identifier)")
```
A little DRYer, a little easier to extend, and I think easier to read than explicitly checking all possible combinations of identifier type and NodeIdType.

oroulet: looks good !

oroulet: but

```
for identifier, nodestypes in valid_type_combinations:
   if xxxx
```

use lower_case AND split tuple instead of indexing

Randelung: Oh yeah, I keep forgetting about that. :D

asyncua/ua/uatypes.py
```diff
@@ -306,6 +306,18 @@ def __init__(self, identifier=None, namespaceidx=0, nodeidtype=None):
...
+                (int, [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte]),
+                (str, [NodeIdType.String, NodeIdType.ByteString]),
+                (bytes, [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte]),
+                (uuid.UUID, [NodeIdType.Guid])
+            ]
+            for identifier, valid_node_types in valid_type_combinations:
+                if isinstance(self.Identifier, identifier) and self.NodeIdType in valid_node_types:
+                    break
+            else:
+                raise UaError(f"NodeId of type {self.NodeIdType} has an incompatible identifier {self.Identifier} of type {type(self.Identifier)}")
```
oroulet, line 320: lokos good but sorry I had one more. Could you move that check in a method called check() for example?
```

def check(self):
   valid_type_combinations = [
                (int, [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte]),
                (str, [NodeIdType.String, NodeIdType.ByteString]),
                (bytes, [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte]),
                (uuid.UUID, [NodeIdType.Guid])
            ]
            for identifier, valid_node_types in valid_type_combinations:
                if isinstance(self.Identifier, identifier) and self.NodeIdType in valid_node_types:
                    return
            raise UaError(f"NodeId of type {self.NodeIdType} has an incompatible identifier {self.Identifier} of type {type(self.Identifier)}")
```

thus we avoid having such a long constructor method and let users check them if they want

oroulet: and call check in constructor of course

oroulet: now I am finished ;-)

asyncua/ua/uatypes.py
```diff
@@ -306,6 +306,24 @@ def __init__(self, identifier=None, namespaceidx=0, nodeidtype=None):
...
+        '''
+        valid_type_combinations = [
+            (int, [NodeIdType.Numeric, NodeIdType.TwoByte, NodeIdType.FourByte]),
+            (str, [NodeIdType.String, NodeIdType.ByteString]),
+            (bytes, [NodeIdType.ByteString, NodeIdType.TwoByte, NodeIdType.FourByte]),
+            (uuid.UUID, [NodeIdType.Guid])
+        ]
+        for identifier, valid_node_types in valid_type_combinations:
+            if isinstance(self.Identifier, identifier) and self.NodeIdType in valid_node_types:
+                break
```
oroulet, line 324: rerurn is best here

asyncua/ua/uatypes.py
```diff
@@ -306,6 +306,24 @@ def __init__(self, identifier=None, namespaceidx=0, nodeidtype=None):
...
+        else:
```
oroulet, line 325: no need for else

PR #300, opened 2020-09-24
Lazy load lxml

joncutting:
Lazy load lxml to reduce frozen app size (>2MB) and save time at startup and for applications that don't require lxml processing. Should also speed up unit testing. Caveat - I have not run unit tests on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
   - pip install pytest-mock
   - pip install cryptography
   - pip install dataclasses
+  - pip install lazy_import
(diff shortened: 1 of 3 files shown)


oroulet:
Not sure what to think about that one... Quite a lot of code for a very special case

oroulet:
Also it breaks all tests....

oroulet:
The best would be to get rid of lxml ... Not much better that the build in stuff

swamper123:
Since lxml was removed with #346 today, this PR is deprecated.

Issue #296, opened 2020-09-23
Memory leak in Master


mar-ar:
Hey,

I found a memory leak in the current master branch. My server was running on the last release 0.8.4 and there was no Probem.
Then I updated the library to the current master and realized a memory leak. First I thought that I have a leak in my code but now I found the error in the library.

To detect the leak I logged like this: (simplified)
import objgraph
objgraph.show_growth(limit=20, file=myFile)

Then I found that every 10 seconds there is one more frame and coroutine
frame          991        +1
coroutine      989        +1

Afterwards I searched the guilty pull request with the result #272 
I just deleted line 139 and everything worked fine again.

Now we need to find out if the intended fix still works. 
I dont know why we need this recursion there. 

Can someone help me to fix that again?

best regards

mar-ar:
I just thought about it and I think I found the fault:

We need to replace the last line.
`
await self._await_closing_tasks()
`
into
`
self.iserver.loop.create_task(self._await_closing_tasks())
`

joncutting:
I think I just started seeing this today after switching from 0.8.4 to master.

When interrupting my program with Ctrl-C I get this:

2020-09-24 09:58:37,264 - asyncio - ERROR: unhandled exception during asyncio.run() shutdown
task: <Task finished name='Task-1' coro=<main() done, defined at C:/mystuff/myserver.py:575> exception=RecursionError('maximum recursion depth exceeded while calling a Python object')>
Traceback (most recent call last):
  File "C:\Python38\lib\asyncio\runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "C:\Python38\lib\asyncio\base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "C:\Python38\lib\asyncio\windows_events.py", line 316, in run_forever
    super().run_forever()
  File "C:\Python38\lib\asyncio\base_events.py", line 570, in run_forever
    self._run_once()
  File "C:\Python38\lib\asyncio\base_events.py", line 1823, in _run_once
    event_list = self._selector.select(timeout)
  File "C:\Python38\lib\asyncio\windows_events.py", line 430, in select
    self._poll(timeout)
  File "C:\Python38\lib\asyncio\windows_events.py", line 779, in _poll
    status = _overlapped.GetQueuedCompletionStatus(self._iocp, ms)
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #302, opened 2020-09-28
fix 296

mar-ar:
#296

### DIFF
--- a/asyncua/server/binary_server_asyncio.py
+++ b/asyncua/server/binary_server_asyncio.py
@@ -167,4 +167,4 @@ class BinaryServer:
                 logger.exception("Unexpected crash in BinaryServer._await_closing_tasks")
         if recursive:
             await asyncio.sleep(10)
-            await self._await_closing_tasks()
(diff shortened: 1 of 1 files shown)


asyncua/server/binary_server_asyncio.py
```diff
@@ -167,4 +167,4 @@ async def _await_closing_tasks(self, recursive=True):
                 logger.exception("Unexpected crash in BinaryServer._await_closing_tasks")
         if recursive:
             await asyncio.sleep(10)
-            await self._await_closing_tasks()
+            self.iserver.loop.create_task(self._await_closing_tasks())
```
oroulet, line 170: OK that does not work.... We added a method to wait for lost not awaited task but then you move that method to task so it is not awaited anymore... this is going in a loop....
Something is wrong here.I have not had the time to look at what your issue is...

mar-ar: Oh you are right.
I also didnt took a closer look at it, but there is somewhere a mistake.

mar-ar:
At least we get a recursion Error and the loop is interrupted.

Issue #305, opened 2020-10-02
Multiple Nodes can share same BrowseName with same parent


swamper123:
If I create two Nodes which share the same parent and have the same Browsename, the server is still running without throwing `Bad_BrowseNameDuplicated`.

Here "MyVariable" share the same parent "MyObject".

![grafik](https://user-images.githubusercontent.com/24408657/94897904-dcbb5b80-0490-11eb-8953-df73842205f2.png)

This should not be allowed.

EDIT: Even if their type may be different, the BrowseName can still be similar.

swamper123:
Case solved and closed 🕵️

PR #306, opened 2020-10-02
Add check if BrowseName already exists in ParentNode

swamper123:
Fix for #305 

I haven't dealed so much with UaErrors, so I don't know what may be best practise in such a case.

This is a solution where the error is logged and the second Node won't be created.

Critics are welcomed. :)

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -246,6 +246,20 @@ class NodeManagementService:
             result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
             return result
 
+        if bool(self._aspace._nodes) and type(item.ParentNodeId) in [ua.NumericNodeId, ua.NodeId]:
+            try:
+                if item.BrowseName.Name in [ref.BrowseName.Name for ref in self._aspace._nodes[item.ParentNodeId].references]:
+                    self.logger.warning("AddNodesItem: Requested Browsename %s already exists in Parent Node", item.BrowseName.Name)
+                    result.StatusCode = ua.StatusCode(ua.StatusCodes.BadBrowseNameDuplicated)
+                    return result
+            except KeyError as e:
+                if e.args[0].Identifier == 0 and e.args[0].NamespaceIndex == 0:
+                    pass
+                else:
+                    self.logger.warning(f"{e} - NodeParent does not exist in Server")
+                pass
+
+
         nodedata = NodeData(item.RequestedNewNodeId)
 
         self._add_node_attributes(nodedata, item, add_timestamps=check)
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -246,17 +246,16 @@ class NodeManagementService:
             result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
             return result
 
-        if bool(self._aspace._nodes) and type(item.ParentNodeId) in [ua.NumericNodeId, ua.NodeId]:
+        if bool(self._aspace._nodes) and type(item.ParentNodeId) in [ua.NumericNodeId, ua.NodeId] and item.ParentNodeId.Identifier != 0 and item.ParentNodeId.NamespaceIndex != 0:
             try:
                 if item.BrowseName.Name in [ref.BrowseName.Name for ref in self._aspace._nodes[item.ParentNodeId].references]:
                     self.logger.warning("AddNodesItem: Requested Browsename %s already exists in Parent Node", item.BrowseName.Name)
                     result.StatusCode = ua.StatusCode(ua.StatusCodes.BadBrowseNameDuplicated)
                     return result
             except KeyError as e:
-                if e.args[0].Identifier == 0 and e.args[0].NamespaceIndex == 0:
-                    pass
-                else:
-                    self.logger.warning(f"{e} - NodeParent does not exist in Server")
+                self.logger.warning(f"{e} - NodeParent does not exist in Server")
+            except BaseException as e:
+                self.logger.warning(f"Unknown Exception thrown: {e}")
                 pass
 
 
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -246,15 +246,18 @@ class NodeManagementService:
             result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
             return result
 
-        if bool(self._aspace._nodes) and (type(item.ParentNodeId) == ua.NodeId or
-                                          ua.NodeId in type(item.ParentNodeId).__bases__):
+        if type(item.ParentNodeId) == ua.NodeId or ua.NodeId in type(item.ParentNodeId).__bases__:
             try:
-                if item.BrowseName.Name in [ref.BrowseName.Name
-                                            for ref in self._aspace._nodes[item.ParentNodeId].references]:
-                    self.logger.warning("AddNodesItem: Requested Browsename %s already exists in Parent Node",
-                                        item.BrowseName.Name)
-                    result.StatusCode = ua.StatusCode(ua.StatusCodes.BadBrowseNameDuplicated)
-                    return result
+                for ref in self._aspace._nodes[item.ParentNodeId].references:
+                    # Check if the Parent has a "HasChild" Reference (or subtype of it) with the Node
+                    if ref.ReferenceTypeId.Identifier in [ua.ObjectIds.HasChild, ua.ObjectIds.HasComponent,
+                                               ua.ObjectIds.HasProperty, ua.ObjectIds.HasSubtype,
+                                               ua.ObjectIds.HasOrderedComponent] and \
+                            item.BrowseName.Name == ref.BrowseName.Name:
+                        self.logger.warning("AddNodesItem: Requested Browsename %s already exists in Parent Node",
+                                            item.BrowseName.Name)
+                        result.StatusCode = ua.StatusCode(ua.StatusCodes.BadBrowseNameDuplicated)
+                        return result
             except KeyError as e:
                 self.logger.warning(f"{e} - NodeParent does not exist in Server")
             except BaseException as e:
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -246,7 +246,12 @@ class NodeManagementService:
             result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
             return result
 
-        try:
+        if item.ParentNodeId not in self._aspace._nodes:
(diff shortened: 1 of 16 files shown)


swamper123:
While testing my Code, I got this thing here:

grafik

Which leads to a failing test, because an 'UaClient' object has no attribute 'aspace'.

What I am wondering about is, how a Client shall be able to use `add_folder()` and similar `add_*()` methods? I thought these are thing, only a server should be able to do?!

swamper123:
Today I learned: There are [Services defined in the specs in Part 4 5.7.2.](https://reference.opcfoundation.org/v104/Core/docs/Part4/5.7.2/) where Clients are allowed to create Nodes. So this behaviour is fine and I have to figure out how and where I can implement this feature in any kind of way. :/

swamper123:
I fixed some tests, where the Nodes with the same parent have the same BrowseNames.

It is not easy to deal that sort of thing in a clean/nice looking way and I haven't found a better place to catch all methods where a Node shall be created and it is possible to check the parent etc, in an atomic way.

swamper123:
So the most of the test servers have been reused in the testmodules. 
This behaviour is normally good and fine, but broke the PR in this case, because no Test was tidying up behind itself and Browsenames have been reused multiple times.
So either all Tests have to get back in the state as they start in every test or the server has to been created for every test (which is done here). Con side is, that tests take longer and it seems that it even stucks here and there.

I am not sure why there are sometimes parent folders missing (e.g. the Node "ServerNamespaces") or a Node with the browsename "EnumVal" was created (perhaps an xml issue? or a Nodeset thing?) again inside the same parent (which is illegal).

swamper123:
Well I am stucked here like the test itself.
I can't recreate the tests on my local device for Python 3.6 at the moment, so sorry for these many commits. I don't know where or why it is stucked since there isn't any usefull feedback of the test or code itself.

The other Python versions doesn't need much time for that one. It doesn't seem to be an interfere thing here...

swamper123:
I am not able to reproduce the error/issue with the same Python version. All tests run fine:

grafik

Does anybody has a clue what it can be?

AndreasHeine:
do we still support 3.6? dont think so!
mabe we can remove the 3.6 test?

kind of unclear in the readme...
> OPC UA / IEC 62541 Client and Server for Python >= 3.6 and pypy . http://freeopcua.github.io/, https://github.com/FreeOpcUa/opcua-asyncio

---------------------

> Motivation
> 
> opcua-asyncio is an asyncio-based asynchronous OPC UA client based on python-opcua, removing hacks for support of python < 3.7. Asynchronous programming allows for simpler code (e.g. less need for locks) and potentially performance gains.

swamper123:
Perhaps we may update that into "We realy do not support <= 3.6 anymore" anyway in another PR?

What I found on Travis:
> My build script is killed without any error #
> 
> Sometimes, you’ll see a build script causing an error and the message in the log will be something like Killed.
> This is usually caused by the script or one of the programs it runs exhausting the memory available in the build sandbox, which is currently 3GB. Plus, there are two cores available, bursted.
> Depending on the tool in use, this can be caused by a few things:
>     Ruby test suite consuming too much memory
>     Tests running in parallel using too many processes or threads (e.g. using the parallel_test gem)
>     g++ needing too much memory to compile files, for instance with a lot of templates included.

So perhaps the combo between Python Version and OS was not good enough and caused memory problems?! 
The tests are running on ubuntu xenial (16.04) which is pretty outdated todays... I will try to get travis running with different os systems.

swamper123:
"It's a small space for me ... but a giant reason for travis to not being triggered" *grumble*
So Python 3.6 realy consumes to much RAM here. The return of "error 137" is the OOM (out of memory) error for some Dockers.

I bet this is caused, because every server is build up and teared down after each test. This would mean, that Python 3.6 or pytest (incl. plugins) (or both) are causing memory leaks here and it seems to explode everytime `test_custom_structures.py::test_opc_type_dict_append_struct_1` is executed.

So my Test could have been passed, because I got a bit more RAM on my local device. 

I would still recommened:
- setup and tear down as server after each test, because interfereing tests can cause other troubles
- drop Python 3.6 (see #312 )
- and this PR will be rebased if #312 is okay for the rest

swamper123:
So the state of this PR is, that I tried to remove all nodes that are created in each test, so that they are not interfere each other.
But this doesn't solve a possible leak issue, while running tests. I thought that with "scope=functional" the server would be destroyed everytime, but even with a forced `gc.collect()` the problem still appears that Travis kinda freeze to death (10min, no respons). 
I will still have a look at that issue in future, since this problem can cause more trouble in the future.

swamper123:
(This PR won't be touched anymore and should be ready)

asyncua/server/address_space.py
```diff
@@ -246,6 +246,27 @@ def _add_node(self, item, user, check=True):
...
+                if ref.ReferenceTypeId.Identifier in [ua.ObjectIds.HasChild, ua.ObjectIds.HasComponent,
+                                           ua.ObjectIds.HasProperty, ua.ObjectIds.HasSubtype,
+                                           ua.ObjectIds.HasOrderedComponent] and ref.IsForward:
+                    if item.BrowseName.Name == ref.BrowseName.Name:
+                        self.logger.warning(f"AddNodesItem: Requested Browsename {item.BrowseName.Name}"
+                                            f" already exists in Parent Node. ParentID:{item.ParentNodeId} --- "
+                                            f"ItemId:{item.RequestedNewNodeId}")
+                        result.StatusCode = ua.StatusCode(ua.StatusCodes.BadBrowseNameDuplicated)
+                        return result
+        except KeyError as e:
```
oroulet, line 261: instead of try/except it is faster (and recommanded) to check if key is in dict BEFORE.
something like:

if not item.ParentNodeId in self._aspace._nodes:
    return errror
for ref in xxxxx

asyncua/server/address_space.py
```diff
@@ -246,6 +246,27 @@ def _add_node(self, item, user, check=True):
...
+            if item.ParentNodeId.Identifier == 0 and item.ParentNodeId.NamespaceIndex == 0:
+                self.logger.debug(f"{e} - NodeParent seems to be Root Node")
+            else:
+                self.logger.debug(f"{e} - NodeParent does not exist in Server")
+        except BaseException as e:
+            self.logger.warning(f"Unknown Exception thrown: {e}")
+            pass
```
oroulet, line 268: that one should not be necessary. There is not reason the code over can fail without being a huge bug we need to fix

oroulet:
Sorry that PR had been opened for so long that I completely forgot it. Do not hesitate to ping me when I do not answer.
I think it looks good now. Fix the small stuff over and ping me to merge it

swamper123:
No worries about that, it is in the nature of open source projects. As well I know, that you got other stuff to do as well. I made the changes, hope that travis is not flodded in Backlogs (well it's Monday, so it should go fast today). 
Next time, I will ping you regular. ;)

asyncua/server/address_space.py
```diff
@@ -246,6 +246,24 @@ def _add_node(self, item, user, check=True):
             result.StatusCode = ua.StatusCode(ua.StatusCodes.BadParentNodeIdInvalid)
             return result
 
+        if item.ParentNodeId not in self._aspace._nodes:
+            if item.ParentNodeId.Identifier == 0 and item.ParentNodeId.NamespaceIndex == 0:
+                self.logger.debug("NodeParent of %s seems to be Root Node", item.ParentNodeId)
+            else:
+                self.logger.debug("NodeParent of %s does not exist in Server", item.ParentNodeId)
+        else:
+            for ref in self._aspace._nodes[item.ParentNodeId].references:
```
oroulet, line 255: OK sorry. one more comment:
In fact the entire if clause here seems to be for debugging only. We have an issue already that we have far too much debug logging. So much that it is useless. So I propose to simply remove that code. Just wirte: `if item.ParentNodeId in self._aspace: do the check`
Also I think you do not need to acesss _nodes diurectly. you can probably write: `if item.ParentNodeId in self._aspace: for node in self._aspace[xxx].references`

oroulet: that will be commit 42 ;-) not bad for something that is a relative simple change

swamper123: I am not sure how we should handle missing parents in that case. Should I do it like in the other cases and just return a `BadParentNodeIdInvalid` in case a Parent does not exist (and ignore the root-node case)?

swamper123: > Also I think you do not need to acesss _nodes diurectly. you can probably write: if item.ParentNodeId in self._aspace: for node in self._aspace[xxx].references

The tests seem to pass, but I am still sceptical, since our target is/should only in _nodes, even in later versions. Iterating over the whole Objects feels not right (correct me if I am wrong).

swamper123: > that will be commit 42 ;-) not bad for something that is a relative simple change

🙄 I know...but if I would substract the issues with the tests, it would have been less as well.

oroulet: _nodes is a private variable (starts with underscore) It is clearly not made to be accessed directly.
And also the AddressSpace is made for that:

```

    def __getitem__(self, nodeid):
        return self._nodes.__getitem__(nodeid)

    def get(self, nodeid):
        return self._nodes.get(nodeid, None)

    def __setitem__(self, nodeid, value):
        return self._nodes.__setitem__(nodeid, value)

    def __contains__(self, nodeid):
        return self._nodes.__contains__(nodeid)

```

oroulet: the goal of that patch is not detetc missing parent. I think there are lines before to do that. So at that point in code, you assume that missing parent is something solve.
In theory we should refuse to create nodes with missing parent but there are nodes with missing parents in standard address space so for now we allow it...

swamper123: > _nodes is a private variable

Protected one! Private would be 2 underscores. ;^) Netherless, the handling would be mostly the same. But okay, in that case, I will change that, if it was made for it.

The missing parent stuff was a follow problem in some cases. If you wanted to check the child refs of the parent, but the parent doesn't exist, it wasn't possible to make that check. The lines above that PR catch some cases, but it seems not all. But that will be another problem.

Funfact: There are logs for nodes with missing parents in standard addressspace, but they don't matter. The logs are created in the first run, where parents may be missed, but in the second run, they seem to be added without errors. Otherwise a RuntimeError would be thrown. 

See https://github.com/FreeOpcUa/opcua-asyncio/blob/0261d3ed974b18fc7443914b4dd7a5d49bbef858/asyncua/server/standard_address_space/standard_address_space.py#L31 and the \_\_exit\_\_ line which is used here: 
https://github.com/FreeOpcUa/opcua-asyncio/blob/0261d3ed974b18fc7443914b4dd7a5d49bbef858/asyncua/server/standard_address_space/standard_address_space.py#L42

oroulet: yes that missing parent issue is something important that should be addressed at some point. It might need to reorders even better nodes while importating

PR #312, opened 2020-10-14
Drop Python 3.6 support in travis tests

swamper123:
Drop Python 3.6 support in travis tests
Upgrade test cases from ubuntu xenial to bionic
Update README


oroulet:
while you are at it you could add tests for python 3.9 if possible.
Also yes I agree about removing 3.6 suppott completely

swamper123:
It seems, that python 3.9 is still not available for travis-ci, just the 3.9-dev branch at the moment. I guess this will still be alright, until the official support is released.

PR #317, opened 2020-10-20
fix a potential race condition

oroulet:


asyncua/common/connection.py
```diff
@@ -264,6 +269,9 @@ def _check_sym_header(self, security_hdr):
             self.revolve_tokens()
             return
 
+        if self._renewal_started and security_hdr.TokenId == self.security_token.TokenId + 1:
```
Randelung, line 272: This just disables the check, yes? Potential security issue. Not a fix (the underlying race condition still exists), but a workaround.
I can think of at least the wrap-around case where the right-hand check won't hold up.

asyncua/common/connection.py
```diff
@@ -277,6 +285,11 @@ def _check_sym_header(self, security_hdr):
             return
 
         expected_tokens = [self.security_token.TokenId, self.next_security_token.TokenId]
+
+        if self._renewal_started:
+            extra_token = self.security_token.TokenId + 1
```
Randelung, line 290: Is the next token always `n+1`? The [spec](https://reference.opcfoundation.org/v104/Core/docs/Part4/5.5.2/) doesn't specify any relation between old and new, potentially making a random new token possible, making prediction impossible. Is it defined somewhere else?

oroulet: it make it accept temporarily the next token. But yes this is not nice. I mainly made that code to test if this solves the issue. So far I got only one report and it is negative....

oroulet: OK. you say tthat it fixes it... I need to have a look at all these merge request...

PR #319, opened 2020-10-23
if you lose the connection the endless loop has to be closed

mar-ar:
Hi,

I lost the connecton to the server and then had this error and tryed to fix it. 

2020-10-23 15:12:02,399 - asyncio:1604 - ERROR - Task exception was never retrieved
future: <Task finished coro=<UaClient._publish_loop() done, defined at asyncua\client\ua_client.py:497> exception=ConnectionError('Connection is closed')>

Any problems with it?

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -526,6 +526,8 @@ class UaClient:
             except UaStructParsingError:
                 ack = None
                 continue
+            except ConnectionError:
+                break
             subscription_id = response.Parameters.SubscriptionId
             if not subscription_id:
                 # The value 0 is used to indicate that there were no Subscriptions defined for which a


ZuZuD:
If the client's connection is closed, any related subscription is lost and you have to reconnect. The publish_task is then either done or cancelled (if you close the session before the timeout raises ```ConnectionError```). 

As you can see [here](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ua_client.py#L458-L463), the next subscription on the reconnected client will then create a brand new publish_task. 

However, what you describe if it's followed by an unability to send/receive publishing on the reconnected client is probably related to #318.

mar-ar:
What you say is totally true. This is just like you say when the session is closed that the ConnectioError didnt raise because I think its the right behavior so we shouldnt raise an error.

ZuZuD:
It's perfectly fine to raise [ConnectionError](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ua_client.py#L145-L147) when the socket is disconnected, because anyway you will have to reconnect and the task will be re-created.

mar-ar:
It is not that bad you are right but you then raise it several times.

Issue #313, opened 2020-10-14
Error while renewing session: Invalid security token id 18, expected one of: [16, 17, 0]


slokhorst:
I have a running client with subscription that does just fine. After a while, I get the following error and the client disconnects:

```
Exception raised while parsing message from server
Traceback (most recent call last):
  File "/lib/python3.8/site-packages/asyncua/client/ua_client.py", line 79, in _process_received_data
    msg = self._connection.receive_from_header_and_body(header, buf)
  File "/lib/python3.8/site-packages/asyncua/common/connection.py", line 323, in receive_from_header_and_body
    self._check_sym_header(security_header)
  File "/lib/python3.8/site-packages/asyncua/common/connection.py", line 282, in _check_sym_header
    raise ua.UaError(f"Invalid security token id {security_hdr.TokenId}, expected one of: {expected_tokens}")
asyncua.ua.uaerrors._base.UaError: Invalid security token id 18, expected one of: [16, 17, 0]
Error while renewing session
Traceback (most recent call last):
  File "/lib/python3.8/site-packages/asyncua/client/client.py", line 387, in _renew_channel_loop
    val = await self.nodes.server_state.read_value()
  File "/lib/python3.8/site-packages/asyncua/common/node.py", line 167, in read_value
    result = await self.read_data_value()
  File "/lib/python3.8/site-packages/asyncua/common/node.py", line 178, in read_data_value
    return await self.read_attribute(ua.AttributeIds.Value)
  File "/lib/python3.8/site-packages/asyncua/common/node.py", line 285, in read_attribute
    result = await self.server.read(params)
  File "/lib/python3.8/site-packages/asyncua/client/ua_client.py", line 353, in read
    data = await self.protocol.send_request(request)
  File "/lib/python3.8/site-packages/asyncua/client/ua_client.py", line 141, in send_request
    self._send_request(request, timeout, message_type),
  File "/lib/python3.8/site-packages/asyncua/client/ua_client.py", line 130, in _send_request
    self.transport.write(msg)
AttributeError: 'NoneType' object has no attribute 'write'
```

I'm connecting to a Siemens S7-1200 PLC's built-in OPC-UA server.

swamper123:
I bet it is happening after ~ 1hr while you are connected the whole time with the Client?

slokhorst:
It happened approximately 10 minutes after connecting.

AndreasHeine:
which version of asyncua? 
[...]

PR #321, opened 2020-10-28
Proposal for #313

Randelung:
It's now 6am, so I'm just dropping this and heading to bed.
Feel free to comment, the thing's been running for an hour now without issues. I think the concept holds.
So apparently the #313 doesn't work in titles.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -70,7 +70,7 @@ class UASocketProtocol(asyncio.Protocol):
                     self.logger.debug('Not enough data while parsing header from server, waiting for more')
                     return
                 if len(buffer_header) < header.body_size:
-                    self.logger.warning(
+                    self.logger.debug(
                         'We did not receive enough data from server. Need %s got %s', header.body_size, len(buffer_header)
                     )
                     return
@@ -705,5 +705,4 @@ class UaClient:
         response = struct_from_binary(ua.WriteResponse, data)
         response.ResponseHeader.ServiceResult.check()
         return response.Results
-
 
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -53,7 +53,7 @@ class UASocketProtocol(asyncio.Protocol):
     def data_received(self, data: bytes):
         # buffer everything
         self.receive_buffer += data
-        asyncio.create_task(self._async_data_received())
+        self.loop.create_task(self._async_data_received())
 
     async def _async_data_received(self):
         """
@@ -705,4 +705,3 @@ class UaClient:
         response = struct_from_binary(ua.WriteResponse, data)
         response.ResponseHeader.ServiceResult.check()
         return response.Results
-
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -29,7 +29,7 @@ class UASocketProtocol(asyncio.Protocol):
         self.logger = logging.getLogger(f"{__name__}.UASocketProtocol")
         self.loop = loop or asyncio.get_event_loop()
         self.transport = None
-        self.receive_buffer: Optional[bytes] = None
+        self.receive_buffer = bytes()
         self.is_receiving = False
         self.timeout = timeout
         self.authentication_token = ua.NodeId()
@@ -39,6 +39,7 @@ class UASocketProtocol(asyncio.Protocol):
         self._connection = SecureConnection(security_policy)
         self.state = self.INITIALIZED
         self.closed: bool = False
+        self.renewal_lock: asyncio.Lock = asyncio.Lock()
 
     def connection_made(self, transport: asyncio.Transport):
         self.state = self.OPEN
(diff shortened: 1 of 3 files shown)


asyncua/common/connection.py
```diff
@@ -303,7 +303,7 @@ def _check_incoming_chunk(self, chunk):
                 else:
                     # Condition for monotonically increase is not met
                     raise ua.UaError(f"Received chunk: {chunk} with wrong sequence expecting:"
-                                     f" {self._peer_sequence_number}, received: {seq_num},"
+                                     f" {self._peer_sequence_number + 1}, received: {seq_num},"
```
Randelung, line 306: It's supposed to be one more than the currently tracked number, which is also the check just above. I got a little confused when the error popped up and said "expecting: 594, received: 594" 😀

swamper123:
Tracvis CI seems to use an pypy3 version which seems to act like Python 3.6.9. ... *sigh*
So some stuff may not work here.

The `send_hello` stuff is breaking, caused by a timeout. Could this be caused by the Lock?

Randelung:
Oh wow, ok, not sure. I've never worked with pypy before. The locks by themselves shouldn't be an issue since they've been [stable for a while](https://docs.python.org/3.5/library/asyncio-sync.html#asyncio.Lock). I'll try to get a pypy environment going I guess.
Do I interpret correctly that the other three tests with normal python tested fine?

asyncua/client/ua_client.py
```diff
@@ -50,42 +51,50 @@ def connection_lost(self, exc):
...
+                    header = header_from_binary(buffer_header)
                 except ua.utils.NotEnoughData:
                     self.logger.debug('Not enough data while parsing header from server, waiting for more')
-                    self.receive_buffer = data
                     return
-                if len(buf) < header.body_size:
-                    self.logger.debug(
-                        'We did not receive enough data from server. Need %s got %s', header.body_size, len(buf)
+                if len(buffer_header) < header.body_size:
+                    self.logger.warning(
```
Randelung, line 73: Whoops, this is supposed to be debug. I had it set higher for better visibility.

Randelung:
```suggestion
                    self.logger.debug(
```

Randelung:
Are you sure it's 3.6.9? I got pypy3.6-v7.3.2 set up, but it says no module named dataclasses for crypto/uacrypto.py, so I assume this wouldn't have worked for a while.
Using pypy3.7-v7.3.2 I can't get my application to run, it hangs on initializing the server (whose source I didn't touch). CTRL-C won't even work to end it. The master branch has the same behavior. It might just be pypy3...
The prosys example crashes on master and runs on fix_313 (except it's noticeably slower than regular python).

Randelung:
Another consideration to be made is whether the asyncio.Protocol approach is best or if a `reader, writer = await asyncio.open_connection(...)` would yield better results. The loop over the `reader` would dictate when packets are being read and processed, so parsing can be offloaded into a different task only if the channel state allows it. (Since dataclasses are being used I'm just also going with the minimum of 3.7.)

swamper123:
> Are you sure it's 3.6.9?

From their page they say:

> PyPy3 implements the Python language version 3.6.9. It has been released, but Python is a large language and it is quite possible that a few things are missing.

grafik

So errors like this may happen:

grafik
(the last line is caused by "Python 3.6") Well this is my observation on this point of the Test.

It seems that the la[test Versions for ppypy3.7 still got some problems ](https://buildbot.pypy.org/summary?branch=%3Ctrunk%3E) among others with asyncio. So we should be patient about that perhaps.

Randelung:
How do you propose we continue? Anything you need me to do?

swamper123:
I have run it with Python 3.6 and it worked.
I tried to get pypy3 running (and I failed with that 😞), so I couldn't test your branch localy.

> How do you propose we continue? Anything you need me to do?

To be honest, I don't know. The last commit of the master branch ran fine with pypy3. 😕 Travis is even running 1:1 the same version as your testbuild did. I kicked the test again and we may see, if something strange happens, but I believe not.

Randelung:
> I have run it with Python 3.6 and it worked.
Wait what? How? Didn't dataclasses arrive in 3.7? server.users and crypto.uacrypto both use dataclasses. asyncio.create_task is 3.7+. Am I missing something?

Pypy was a pain to get running tbh. Took ages to install and compile dependencies and dependency dependencies (am on Windows, hardly any of the packages just installed as promised on the pypy page). I couldn't get the master branch to run the example on 3.6, that could be a Windows problem, though.

swamper123:
Nothing changed... I am out of ideas here. Sorry for that 😕

swamper123:
> Wait what? How? Didn't dataclasses arrive in 3.7? server.users and crypto.uacrypto both use dataclasses. asyncio.create_task is 3.7+. Am I missing something?

Clone your repo, checkout branch fix_313, execute `python3.6 pytest -v -s` in cmd...so localy.

> Didn't dataclasses arrive in 3.7?

Yes, but there is a backport on pypy and dataclasses is (still, but perhaps shouldn't anymore) in `.travis` install.

> Pypy was a pain to get running tbh.

Python and Windows are enemies since I can thought...and I bet it is even worse using pypy ... and more worse run pytest in pypy...and even more worse test asyncio stuff with pytest in pypy on windows (asyncio (python around 3.6) and windows haven't been good friends, don't know if this is solved atm).

Randelung:
> Yes, but there is a backport on pypi for python3.6 and dataclasses

Oooooh, that explains it. I'll replace the create_task with the loop version, then, and try again!
And then I'll try the suggestion that oroulet made about moving the open_secure_channel code into the receive loop.

asyncua/client/ua_client.py
```diff
@@ -50,42 +51,50 @@ def connection_lost(self, exc):
         self.transport = None
 
     def data_received(self, data: bytes):
-        if self.receive_buffer:
-            data = self.receive_buffer + data
-            self.receive_buffer = None
-        self._process_received_data(data)
+        # buffer everything
+        self.receive_buffer += data
+        asyncio.create_task(self._async_data_received())
```
Randelung, line 56: Not 3.6 compatible

Randelung:
```suggestion
        self.loop.create_task(self._async_data_received())
```

Randelung:
Um I assume it's displaying the previous result. I'm getting 1 failed, 461 passed, 5 skipped, 6 warnings, 1 error for both pypy versions now.

swamper123:
> Um I assume it's displaying the previous result. I'm getting 1 failed, 461 passed, 5 skipped, 6 warnings, 1 error for both pypy versions now.

The test was cancelled. I kicked it again

Randelung:
Now two tests are running :D The cancelled one was the previous commit I think.

oroulet:
another fix has been merged

Issue #307, opened 2020-10-05
XML-Import: Opc.Ua.Di.NodeSet2.xml + Opc.Ua.Machinery.NodeSet2.xml


AndreasHeine:
Hey,

i am kind of confused, i looked for quite some time to find the issue but i did not found it yet... maybe someone sees what i dont!

1. Import DI nodeset (Opc.Ua.Di.NodeSet2.xml)
2. Import Machinery nodeset (Opc.Ua.Machinery.NodeSet2.xml)
3. await server.load_type_definitions()
4. IMachineryItemVendorNameplateType ??? and some more are not there...

```xml
<UAObjectType NodeId="ns=1;i=1003" BrowseName="1:IMachineryItemVendorNameplateType" IsAbstract="true">
    <DisplayName>IMachineryItemVendorNameplateType</DisplayName>
    <Description Locale="en">Interface containing identification and nameplate information for a MachineryItem provided by the vendor</Description>
    <Documentation>https://reference.opcfoundation.org/v104/Machinery/v100/docs/8.2</Documentation>
    <References>
      <Reference ReferenceType="HasSubtype" IsForward="false">ns=2;i=15035</Reference>
      <Reference ReferenceType="HasProperty">ns=1;i=6027</Reference>
      <Reference ReferenceType="HasProperty">ns=1;i=6022</Reference>
      <Reference ReferenceType="HasProperty">ns=1;i=6026</Reference>
      <Reference ReferenceType="HasProperty">ns=1;i=6024</Reference>
      <Reference ReferenceType="HasProperty">ns=1;i=6025</Reference>
      <Reference ReferenceType="HasInterface" IsForward="false">ns=1;i=1004</Reference>
    </References>
  </UAObjectType>
```

> INFO:asyncua.common.xmlparser:Parsing node: ns=1;i=1003 1:IMachineryItemVendorNameplateType

???

swamper123:
???
They are in the Nodeset, but aren't loaded?

swamper123:
Does this problem only appear with the Machinery nodesets or are other affected as well?

AndreasHeine:
actually you need DI for Machinery and i want the namplate stuff... tbh i have not tested it with other nodesets!

swamper123:
[...]

PR #323, opened 2020-10-30
added instantiate_optional support on create object method and possible fix for #307

AndreasHeine:
Added Option for create_object() -> instantiate_optional. No breaking change just added options for more control if you create a object.

### DIFF
--- a/asyncua/common/manage_nodes.py
+++ b/asyncua/common/manage_nodes.py
@@ -50,7 +50,7 @@ async def create_folder(parent, nodeid, bname):
     )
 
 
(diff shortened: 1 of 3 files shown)


AndreasHeine:
main reason for this is if you want to implement a opc ua companion spec. server you'll need control about what you instantiate. if you create a instance of a companion-spec.-object-type you most of the time need only the mandatorys.

AndreasHeine:
@oroulet sorry was a week off what is the status on this? the first part instantiate oprionals is just convenience but in my opinion usefull, the second part does not fix parsing but prevents the error from NodeId == Null 

should be mergable, if you have no concerns?

PR #324, opened 2020-10-31
new approach to the new security token race condition

Randelung:
Alternative to #321, possible fix for #313.

Instead of making the receive loop async and using a lock, I use an extra field to pass request parameters (to the parse job) and result (for the return statement) back and forth, meaning there's a new condition of not being able to make a new request if the old one is still running.
Of course, a lock could be used to keep new requests waiting and execute them immediately after, but renewal this quickly is probably just not going to happen.

... he said foreshadowingly. *thunder cracks in the distance*

Forgot to mention that I tried to add_done_callback to the future, but that didn't work. It appears the callback is scheduled asynchronously.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -39,6 +39,9 @@ class UASocketProtocol(asyncio.Protocol):
         self._connection = SecureConnection(security_policy)
         self.state = self.INITIALIZED
         self.closed: bool = False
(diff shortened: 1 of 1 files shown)


oroulet:
Does it fix the issue?

oroulet:
@Randelung @AndreasHeine does that fix your issue?

Randelung:
So far, I haven't detected any more issues on my end.

oroulet:
this not so nice but I have no better idea for now

Issue #308, opened 2020-10-07
import_xml fails on 0.9.0 (worked on 0.8.2)


jcbastosportela:
Great job, first of all, nice to see this project so active!

I have been using asyncua for a while now, but recently I noticed that the ServerTimestamps where being returned as "0", so I decided to try out the most recent version on asyncua via pip (0.9.0).

After the update, on the top of a series of **"warnings"** about `Message: 'overwriting parent target, shouldbe fixed'` that I did not yet investigate there is an **error**:
`asyncua.ua.uaerrors._base.UaError: Non array Variant of type VariantType.NodeId cannot have value None`

This error appears related with one of my custom VariableType. This VariableType's relevant settings are:

![tempsnip jpg](https://user-images.githubusercontent.com/15647115/95343386-c100ec80-08b8-11eb-885e-a576bc233e34.png)

I have already attempted to change the **ValueRank** from _Any_ to _Scalar_ (as on  the attached picture), but problem persists.

Apart from the existence of the child properties this variable type is similar to the **ServerVendorCapabilityType** from the namespace 0 (http://opcfoundation.org/UA) (NumericNodeID 2137). I also attached that.
**Edit:** In fact I just noticed that the **IsAbstract** is checked in **ServerVendorCapabilityType** but not in my custom VariableType.

![image](https://user-images.githubusercontent.com/15647115/95343908-543a2200-08b9-11eb-8dce-d2b3062ae44e.png)

As you may have noticed I am using UaModeler from Unified Automation for developing my information model.

**The question:** Is this new behavior expected?

Note: I am using asyncua for my simulator of an OPC UA server that in reality is a PLC. The information model in the PLC works fine, as it does on 0.8.2

AndreasHeine:
https://github.com/FreeOpcUa/opcua-asyncio/issues/307

jcbastosportela:
Sorry all, just realized that my issue is also being mentioned in #307 
I have just read the title and first comments, but after sawmper123 reports the same issue.
Anything I can do for helping sorting this out?

swamper123:
@jcbastosportela Could you test, if this issue still exist with the current master branch? In case the problem is gone, we could close that case. :detective:

swamper123:
Since no complains happenes from @jcbastosportela : Case closed - 🕵️

Issue #299, opened 2020-09-24
Make lxml dependency optional


joncutting:
Hi,

As far as I can tell, the package lxml is only used for XML import/export, but it is still a requirement even if that feature is not needed. Normally I would say no big deal but for frozen code (e.g. pyinstaller) it adds an extra 2MB to the executable which means additional delay during startup. If there is any way to make lxml optional - or only imported when it is really needed - then it would be nice to have.

Jonathan

AndreasHeine:
thats difficult... not the removal but if you use OPC UA in the way it is designed for, means to implement standardised interfaces/information models via OPC UA Companion Specs. (pure XML-Nodesets) dont get me wrong feel free to PR but most people use Companion Specs. for modeling so its abolutly necessary and a core part of OPC UA!

joncutting:
OK, understood. There are many opinions out there about how best to lazy load, so I took the lazy way and grabbed `[lazy_import](https://pypi.org/project/lazy-import/)`. I submitted PR #300.

swamper123:
I am not very happy with lxml as well, because there may be some security issues in this library (I may need to investigate further into that topic) about xml attacks. But this library is still not in V1.0 so this will be a topic in the future.
If you want to load the addressspace in a fast way, you may have a look at:
https://github.com/FreeOpcUa/opcua-asyncio/blob/582f3a2065b08cce5e4daa5d7b8272e92e1856d0/asyncua/server/address_space.py#L590
Where you can load a pickled Server into the address space. This is fast, but in my opinion not better than the lxml way (mentioned as well in #277 )

swamper123:
With #346 this dependency is now removed. Case closed 🕵️

PR #331, opened 2020-11-13
[WIP]: Remove lxml and replace with buildin xml

swamper123:
Things this PR shall tackle:

- remove lxml and dependencies (#299 )
- ~~fix missing Nodes after import/create(#307, #308 )~~
- ~~fix missing references after import/create~~
- replace lxml with python buildin xml

~~Optional:
check if Parentnodes already contains a Childnode with same Browsename (see also #306 , maybe the problem can be cought here while importing files as well)~~

~~If you got other problems with xml, except the mentioned ones, please post them here as well, so I can take care of them as well in this PR.~~


oroulet:
@swamper123 great you are adressing this. But not ry to do to much ion one PR. First remove lxml, make all tests pass.
MAybe add some tests which are importing some well know models (like theDi, and robotics models) then try to fix bugs around

swamper123:
@oroulet yeah you are right about that. I should keep it small first...I hope that whole stuff is not intermeshed much. >_<

asyncua/common/structures104.py
```diff
@@ -19,7 +19,7 @@ def clean_name(name):
     newname = re.sub(r'^[0-9]+', r'_\g<0>', newname)
 
     if name != newname:
-        logger.warning("renamed %s to %s due to Python syntax", name, newname)
+        logger.warning(f"renamed {name} to {newname} due to Python syntax")
```
oroulet, line 22: Do not do that. The f-string stuff should not be used for logging. Try to use pylint for example to avoid such errors

swamper123: I can revert that. Why are f-strings in logging bad practice?

oroulet: logging uses lazy string computattion. when using f-string syntax  you force creating the string at runtime, even when not used

swamper123:
So testing with Robotics or DI isn't as easy than be thought. While DI is running (more or less, with a dirty workaround because of `ParameterIdentifier` Node whichs parent is `NameSpaceUri` whoms Value is allowed to be empty. Perhaps this is causing these import errors from the standard_address_spaces?) Robotics is missing some Node attributes, which seems to be missing.
That stuff is even more complicated if there are remaining Errors in the Nodeset which may be fixed in v.1.05. That can mislead the whole parsing thing as well.

We may need to take a look on how the Nodes are thrown together. Perhaps there have been stuff changed? Maybe we refactor some class Names to be spec conform, so it is easier to understand why class xyz exists and where to look at in the specs.

I saw as well that[ *.bsd files are obsolete, s](https://github.com/OPCFoundation/UA-Nodeset/blob/v1.04/readme.md)o the tests should use an _test_file.xml_ instead.

I may have to take a look especially on `make_model_from_string` and `make_model_from_file` since there have been discrepancy (maybe caused by the bsd file as example file).

TLDR: XML Topic is aaaaaaahhhhhhhhh....but I am on it.

asyncua/client/client.py
```diff
@@ -558,7 +558,7 @@ async def create_subscription(self, period, handler):
             params.RequestedLifetimeCount = 10000
             params.RequestedMaxKeepAliveCount = 3000
             params.MaxNotificationsPerPublish = 10000
-            params.PublishingEnabled = True
+            params.PublishingEnabled = publishing
```
oroulet, line 561: why do we have that kind of changes in a branch supposed to focus on removing lxml dependency??

asyncua/common/subscription.py
```diff
@@ -370,3 +387,40 @@ def deadband_monitor(self, var, deadband_val, deadbandtype=1, queuesize=0, attr=
...
+        :return: Return a Set Publishing Mode Result
+        """
+        self.logger.info("set_publishing_mode")
+        params = ua.SetPublishingModeParameters()
+        params.SubscriptionIds = [self.subscription_id]
+        params.PublishingEnabled = publishing
+        result = await self.server.set_publishing_mode(params)
+        if result[0].is_good():
+            self.parameters.PublishingEnabled = publishing
+        return result
```
oroulet, line 426: here too??? try to rebase on master

swamper123: Since this is a draft (and it will take some time), you may take a look on it, when it's ready. Otherwise you look through half-baked code.

It will take a while to understand the whole stuff (objectify, xml.etree(if I even use that, don't know if DOM/minidom or SAX be better?), xml-files/Spec thing etc.). It seems that this isn't an easy three liner code to replace, so that tests run fine. 😕

oroulet: I just wanted to see if I could help before you spend too much time on it...

swamper123: Sorry if it sounded rude, it wasn't meant like that. I appreciate your help realy much.
I believe that this Draft will stay the way it is for a while, until the underlaying stuff works the way as expected. The rat tail may be to long for one PR. :/

oroulet: no pb

PR #346, opened 2020-11-25
Replace lxml with xml.etree.ElementTree

swamper123:
Replace iter() with iterfind()
Remove lxml from setup.py
Remove done FIXME Graffitti
Remove lxml from travis

Tbh, I have no clue why I was wandering in other files in the old draft, but it seems to work....

A fix for #299

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ jobs:
 install:
   - pip install python-dateutil
   - pip install pytz
-  - pip install lxml
   - pip install aiofiles
   - pip install aiosqlite
   - pip install pytest --upgrade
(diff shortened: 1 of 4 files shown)


swamper123:
Guess at what version wildcarsearches are provided in xml. 😡 💢 
I may figure out another way, so that python 3.7 is able to handle that stuff as well (either via version or a general solution).

oroulet:
This looks also simple and good. Should I merge? looks like all tests pass

swamper123:
It seems to be okay, since the old code was just iterating over it, searching for xml-attributes/elements etc., just in an pythonlike Object. If you (and other) haven't got anything to critisize, we could merge it (and maybe close some lxml issues).

PR #350, opened 2020-12-01
fix bug when sending event with no message

JoeyFaulkner:
if im trying to send an event like 

```
evgen = await self._server.get_event_generator()
...
evgen.event.Message = None
await evgen.trigger()
```

I got an error like 
```ValueError: A LocalizedText object takes a string as argument "text", not a <class 'coroutine'>, <coroutine object Node.read_browse_name at 0x7f1dae170340>```

It seems to come from this code which likely hasnt been hit since the port from sync to async :)


oroulet:
missing tests here obvisouly...

PR #352, opened 2020-12-04
User manager and permission ruleset fix

Marco9412:
Hello,
I noticed two bugs in UserManager and PermissionRuleSet handling:

1. the user-defined PermissionRuleset is now propagated and used also into the server endpoint with NoSecurity; without this fix a valid user has full acess to the server if uses an unencrypted connection;
2. a clear password is now passed to the user-defined UserManager.get_user when using a Basic256Sha256SignAndEncrypt endpoint.

### DIFF
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -300,9 +300,17 @@ class InternalServer:
         user_name = token.UserName
         password = token.Password
 
+        # TODO Support all Token Types
+        # AnonimousIdentityToken
+        # UserIdentityToken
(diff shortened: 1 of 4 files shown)


oroulet:
before we can consider to merge that work you need to fix the tests. obvisouly these changes break everything so something is wrong with the patch

Marco9412:
@oroulet I added a check to password which is not provided in many tests, now all tests pass correctly.

JoeyFaulkner (approved):
looks good to me!

Issue #353, opened 2020-12-07
xmlparser - requiredModel not handled


swamper123:
Hello there,

the element "RequiredModel" in some xml files (e.g. machinery) is not handled.
This causes crashes, if a special order while importing was not adhered. (import DI.xml -> import Machinery.xml ✔️ ; import Machinery.xml -> import DI.xml ❌ )

Some solution inspirations:
- import just fails, but with error that schema bar requires  schema foo first. (would be quick, but dirty and more helpful)
- a list of all wanted schemas has to be passed to xmlparser with their path, instead of a single string with a path. Then the order will be figured out by that tag and the nodes imported in a working order (if a required schema is not in that list, it shall be tried to find it in the schema folder, otherwise fail)

AndreasHeine:
in the machinery nodeset is a required field which tells you what is needed

```xml
  <Models>
    <Model ModelUri="http://opcfoundation.org/UA/Machinery/" Version="1.0.0" PublicationDate="2020-09-25T00:00:00Z">
      <RequiredModel ModelUri="http://opcfoundation.org/UA/" Version="1.04.7" PublicationDate="2020-07-15T00:00:00Z" />
      <RequiredModel ModelUri="http://opcfoundation.org/UA/DI/" Version="1.02.2" PublicationDate="2020-06-02T00:00:00Z" />
    </Model>
  </Models>
```


so probably we neet to check that before start parsing!?

oroulet:
but how can we know, if DI is required that  Di is already imported? Does not seem to be a mechanism for that.
But maybe we can simply require that all required files are in the same directory..

swamper123:
@oroulet You can request: Objects->Server->Namespaces and each object there has mandatory attributes NamespaceUri, NamespaceVersion and NamespacePublicationDate (and some more...). So getting the existing Informations is not a big thing.

I like the folder idea, perhaps a better idea than passing a list of filepaths....but I may start with the "is on the server, else throw Error" solution and open a new issue for a more complex/better way (perhaps with the folder solution).

swamper123:
Requirement imports should now be checked. Case closed 🕵️

PR #358, opened 2020-12-11
Check xmlimport requirements

swamper123:
First step of a solution for #353

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -44,12 +44,46 @@ class XmlImporter:
             aliases_mapped[alias] = self._to_migrated_nodeid(node_id)
         return aliases_mapped
 
+
+    async def _get_existing_model_in_namespace(self):
+        server_model_list = []
+        server_namespaces_node = await self.server.nodes.namespaces.get_children()
+        for model_node in server_namespaces_node:
+            server_model_list.append({"ModelUri": await(await model_node.get_child("NamespaceUri")).read_value(),
+                                      "Version": await(await model_node.get_child("NamespaceVersion")).read_value(),
+                                      "PublicationDate": str(await(
+                                          await model_node.get_child("NamespacePublicationDate")).read_value())})
+        return server_model_list
+
+    async def _check_required_models(self, xmlpath=None, xmlstring=None):
+        req_models = self.parser.list_required_models(xmlpath, xmlstring)
+        if not req_models:
+            return None
+        server_model_list = await self._get_existing_model_in_namespace()
+        for model in server_model_list:
+            for req_model in req_models:
+                if model["ModelUri"] == req_model["ModelUri"] and \
+                        model["PublicationDate"] >= req_model["PublicationDate"]:
+                            if "Version" in model and "Version" in req_model:
+                                if model["Version"] >= req_model["Version"]:
+                                    req_models.remove(req_model)
+                            else:
+                                req_models.remove(req_model)
+        if len(req_models):
+            for missing_model in server_model_list:
+                _logger.warning("Model is missing: ", missing_model)
+            raise ValueError("Server doesn't satisfy required XML-Models. Import them first!")
+        return None
+
     async def import_xml(self, xmlpath=None, xmlstring=None):
         """
         import xml and return added nodes
         """
+        if (xmlpath is None and xmlstring is None) or (xmlpath and xmlstring):
+            raise ValueError("Expected either xmlpath or xmlstring, not both or neither.")
         _logger.info("Importing XML file %s", xmlpath)
         self.parser = XMLParser()
+        await self._check_required_models(xmlpath, xmlstring)
         await self.parser.parse(xmlpath, xmlstring)
         self.namespaces = await self._map_namespaces(self.parser.get_used_namespaces())
         _logger.info("namespace map: %s", self.namespaces)
@@ -81,10 +115,13 @@ class XmlImporter:
                 missing.append(nd)
             for ref in nd.refs:
                 if ref.forward:
-                    if ref.reftype in [self.server.nodes.HasComponent.nodeid, self.server.nodes.HasProperty.nodeid, self.server.nodes.Organizes.nodeid]:
+                    if ref.reftype in [self.server.nodes.HasComponent.nodeid,
+                                       self.server.nodes.HasProperty.nodeid,
+                                       self.server.nodes.Organizes.nodeid]:
                         # if a node has several links, the last one will win
                         if ref.target in childs:
-                            _logger.warning("overwriting parent target, shouldbe fixed", ref.target, nd.nodeid, ref.reftype, childs[ref.target])
+                            _logger.warning("overwriting parent target, shouldbe fixed",
+                                            ref.target, nd.nodeid, ref.reftype, childs[ref.target])
                         childs[ref.target] = (nd.nodeid, ref.reftype)
         for nd in missing:
             if nd.nodeid in childs:
@@ -166,7 +203,7 @@ class XmlImporter:
         node.RequestedNewNodeId = self._migrate_ns(obj.nodeid)
         node.BrowseName = self._migrate_ns(obj.browsename)
         _logger.info("Importing xml node (%s, %s) as (%s %s)", obj.browsename,
-                         obj.nodeid, node.BrowseName, node.RequestedNewNodeId)
+                     obj.nodeid, node.BrowseName, node.RequestedNewNodeId)
         node.NodeClass = getattr(ua.NodeClass, obj.nodetype[2:])
         if obj.parent and obj.parentlink:
             node.ParentNodeId = self._migrate_ns(obj.parent)
@@ -427,7 +464,8 @@ class XmlImporter:
                 if self.server.nodes.base_structure_type in path:
                     attrs.DataTypeDefinition = self._get_sdef(node, obj)
                 else:
-                    _logger.warning("%s has datatypedefinition and path %s but we could not find out if this is a struct", obj, path)
+                    _logger.warning("%s has datatypedefinition and path %s"
+                                    " but we could not find out if this is a struct", obj, path)
         node.NodeAttributes = attrs
         res = await self._get_server().add_nodes([node])
         res[0].StatusCode.check()
--- a/asyncua/common/xmlparser.py
+++ b/asyncua/common/xmlparser.py
@@ -7,7 +7,6 @@ import base64
 import logging
 from pytz import utc
 
-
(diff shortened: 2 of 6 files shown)


asyncua/common/xmlimporter.py
```diff
@@ -44,12 +44,32 @@ def _map_aliases(self, aliases: dict):
...
+        return True
+
     async def import_xml(self, xmlpath=None, xmlstring=None):
         """
         import xml and return added nodes
         """
         _logger.info("Importing XML file %s", xmlpath)
         self.parser = XMLParser()
+        if not (await self._check_requirements(xmlpath)):
+            raise ImportError("Server doesn't satisfy required XML-Models. Import them first!")
```
oroulet, line 72: we should list missing models when we first know which one are missing

asyncua/common/xmlimporter.py
```diff
@@ -44,12 +44,32 @@ def _map_aliases(self, aliases: dict):
...
         return aliases_mapped
 
+    async def _check_requirements(self, xmlpath):
+        req_models = self.parser.list_required_models(xmlpath)
+        server_model_list = []
+        server_namespaces_node = await self.server.nodes.namespaces.get_children()  # get_node(ua.NodeId(ua.ObjectIds.Server_Namespaces))
+        for model_node in server_namespaces_node:
+            server_model_list.append({"ModelUri": await(await model_node.get_child("NamespaceUri")).read_value(),
+                                      "Version": await(await model_node.get_child("NamespaceVersion")).read_value(),
+                                      "PublicationDate": str(await(await model_node.get_child("NamespacePublicationDate")).read_value())})
```
oroulet, line 54: what doe the for loop over does? maybe put it in a function with a clear name like: _get_existing_model_in_address_space()

asyncua/common/xmlimporter.py
```diff
@@ -44,12 +44,32 @@ def _map_aliases(self, aliases: dict):
...
+
+        for model in server_model_list:
+            for req_model in req_models:
+                if model["ModelUri"] == req_model["ModelUri"]:
+                    if model["Version"] >= req_model["Version"] and model["PublicationDate"] >= req_model["PublicationDate"]:
+                        server_model_list.remove(model)
```
oroulet, line 60: mayeb it is better to raise exception here directly with all the information. If this is the version req that fails, then write it to user

asyncua/common/xmlimporter.py
swamper123: A cleaner way, agreed

asyncua/common/xmlimporter.py
swamper123: If I raise the exception here, I would only detect one missing required model.
I would prefere to return/log all missing required models and then raise the exception, so that the programmer just have to do it once, not every time.

asyncua/common/xmlimporter.py
swamper123: Indeed! Will be added

asyncua/common/xmlimporter.py
swamper123: Checked!

swamper123:
It seems that Version is optional in xml files, but PublicationDate is necessarry for Version checks.
Tests and testfiles have been added.
Fixes have been made.

asyncua/common/xmlimporter.py
```diff
@@ -44,12 +44,46 @@ def _map_aliases(self, aliases: dict):
...
+                _logger.warning("Model is missing: ", missing_model)
+            raise ValueError("Server doesn't satisfy required XML-Models. Import them first!")
+        return None
+
     async def import_xml(self, xmlpath=None, xmlstring=None):
         """
         import xml and return added nodes
         """
+        if not (bool(xmlpath) != bool(xmlstring)):
+            raise ValueError("Expected either xmlpath or xmlstring, not both or neither.")
```
oroulet, line 83: I do not thing using bool is correct here, unless there is a trick I do not know. usually a test to None is written `X is None`

asyncua/common/xmlimporter.py
```diff
@@ -81,10 +115,13 @@ def _add_missing_parents(self, dnodes):
                 missing.append(nd)
             for ref in nd.refs:
                 if ref.forward:
-                    if ref.reftype in [self.server.nodes.HasComponent.nodeid, self.server.nodes.HasProperty.nodeid, self.server.nodes.Organizes.nodeid]:
+                    if ref.reftype in [self.server.nodes.HasComponent.nodeid,
+                                       self.server.nodes.HasProperty.nodeid,
+                                       self.server.nodes.Organizes.nodeid]:
```
oroulet, line 120: you can add a comma at the end of `Organizes.nodeid`so automatic code formater like yapf keep your formatting

asyncua/common/xmlparser.py
```diff
@@ -384,3 +386,33 @@ def _parse_refs(self, el, obj):
...
+        # Get Nodes of Server -> Namespaces -> Uri, Version and Date
+        pass
+
+    def _check_required_models_with_server(self):
+        # sort args in a nice Dict
+        # -> await get_running_models(server) --> in nice dict as well
+        # compare if requirement in server models
+        # Yes? -> all okay, pass
+        # No? -> raise Exception, not importable!
+        pass
```
oroulet, line 400: what are these empty methods?

tests/test_xml.py
```diff
@@ -19,6 +19,8 @@
 
 CUSTOM_NODES_XML_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), "custom_nodes.xml"))
 CUSTOM_NODES_NS_XML_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), "custom_nodesns.xml"))
+CUSTOM_REQ_XML_PASS_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), "test_requirement_pass.xml"))
+CUSTOM_REQ_XML_FAIL_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), "test_requirement_fail.xml"))
```
oroulet, line 23: usinf the new pathlib is much nicer. but that is also just  a tips

asyncua/common/xmlimporter.py
```diff
@@ -44,12 +44,46 @@ def _map_aliases(self, aliases: dict):
...
+        if not req_models:
+            return None
+        server_model_list = await self._get_existing_model_in_namespace()
+        for model in server_model_list:
+            for req_model in req_models:
+                if model["ModelUri"] == req_model["ModelUri"]:
+                    if model["PublicationDate"] >= req_model["PublicationDate"]:
+                        if "Version" in model and "Version" in req_model:
+                            if model["Version"] >= req_model["Version"]:
+                                req_models.remove(req_model)
```
oroulet, line 69: isn'it much cleaner to write `if > and B and C` instead of `if A; if B; if C`?

asyncua/common/xmlimporter.py
```diff
@@ -44,12 +44,46 @@ def _map_aliases(self, aliases: dict):
...
+        server_namespaces_node = await self.server.nodes.namespaces.get_children()
+        for model_node in server_namespaces_node:
+            server_model_list.append({"ModelUri": await(await model_node.get_child("NamespaceUri")).read_value(),
+                                      "Version": await(await model_node.get_child("NamespaceVersion")).read_value(),
+                                      "PublicationDate": str(await(
+                                          await model_node.get_child("NamespacePublicationDate")).read_value())})
+        return server_model_list
+
+    async def _check_requirements(self, xmlpath=None, xmlstring=None):
+        req_models = self.parser.list_required_models(xmlpath, xmlstring)
```
oroulet, line 59: also maybe call it with full name : required_model so it is easier to understand for everyone reading that

oroulet:
Sorry for writing so many comments, it is not that your code is bad, it is just that I am interested in that issue ;-)

asyncua/common/xmlimporter.py
swamper123: In Version 1I had written:

`if (xmlpath is None and xmlstring is None) or (xmlpath and xmlstring):`

But it looks as ugly as that one. I just wanted to add a "XOR" so that just one param was passed ... do you may know a better/more beautiful way?

asyncua/common/xmlparser.py
swamper123: . . . "Nothing to see here folks, just move along". . . (bad leftovers from trying). Will be removed.

tests/test_xml.py
swamper123: Agreed, I will change them for test_xml.py. We should add it as a ToDo package for the other files as well.

asyncua/common/xmlimporter.py
swamper123: I chosed that way for better readability. Otherwise a newline would be needed (even if others laugh about, I like the max 120 char PEP). But in general you are right, it would be a nicer and more efficient way.

oroulet: wirite `and \` or `\ \n and`I do not remember which one i the correct syntax

swamper123:
@oroulet If you say sorry once more, because you (and I) want a better code (and I want to be a better software deveoloper) to push that project a tiny bit further, I will come through your chimney this christmas and gift you chocolat chip cookies _**with raisins.**_ :^)

Please (not just you oroulet) point to every ugly piece in my commits. Otherwise no one is helped.

asyncua/common/xmlimporter.py
oroulet: maybe ugly but at least Version1 is easy to understand. And it is strongly advised to favorize readable code

swamper123: Alright, I will change that

tests/test_xml.py
```diff
@@ -17,8 +16,11 @@
 
 pytestmark = pytest.mark.asyncio
 
-CUSTOM_NODES_XML_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), "custom_nodes.xml"))
-CUSTOM_NODES_NS_XML_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), "custom_nodesns.xml"))
+BASE_DIR = pathlib.Path(__file__).parent
+CUSTOM_NODES_XML_PATH = BASE_DIR.joinpath("custom_nodes.xml").absolute()
```
oroulet, line 20:
```
BASE_DIR = pathlib.Path(__file__).parent.absolute()
CUSTOM_NODES_XML_PATH = BASE_DIR / "custom_nodes.xml"
```

swamper123: Never heared of that possibility...but seems to work local.

PR #361, opened 2020-12-11
Propose: Add variables to generated event filter and event object

Koseng:
Event types can not only contain properties, but also variables. For example the ConditionType, which is the foundation of Alarms and Conditions, contains variables. This change adds the variables to the generated event filter and event object.
https://reference.opcfoundation.org/v104/Core/docs/Part9/5.5.2/

Without the change the Variables in ConditionType of VariableType like EnabledState, Quality or Comment cannot be accessed in the Event with generated filter select clause.

### DIFF
--- a/asyncua/common/events.py
+++ b/asyncua/common/events.py
@@ -125,8 +125,10 @@ async def select_clauses_from_evtype(evtypes):
     clauses = []
     selected_paths = []
     for evtype in evtypes:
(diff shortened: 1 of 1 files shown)


AndreasHeine:
@Koseng ty for that one helped me alot!

Koseng:
@AndreasHeine :-) Great.

PR #363, opened 2020-12-15
Fixes #296

### LINKED ISSUE #296: Memory leak in Master

Randelung:
Refactor recursive to iterative.
Since the method is not supposed to return, it can't be implemented as recursive.
The result was a memory leak of a coroutine and function frame every 10 seconds.


oroulet:
Sorry forgot that one. Why do you think recursion is necessary?

swamper123:
@oroulet I guess in the old solution, you added a new task to an eventloop every time (awaiting stuff). Recursion maximum possible with memory leaks incl.

In this way, you wouldn't add a new coroutine to the task, instead you are looping in two while loops maximum (in case of recursion). This seems to be a good way for me.

swamper123:
@Randelung You may rebase your branch to head of master.

Randelung:
... did that do it? There was a button "update branch", and I clicked it.

oroulet:
I cannot manage to understand that closing code. It must have been changed by several people and does not make much sense anymore. Why do we have a 10s sleep there?
And currently there is a recursive which make it retry awaiting tasks. This is not a recursion at all it is just calling the same methods two times

oroulet:
shit in fact it is calling the same methods forver. what a a shit

swamper123:
@oroulet Wording! :^)
I guess the asyncio.sleep(10) is there, so that multiple tasks have time to be executed. Why 10? I guess that was a gut feeling.

@Randelung I think so...shouldn't have made a conflict.
Sometimes github can handle that as well, sometimes you have to rebase on your own.

Randelung:
I've noticed that this hasn't been merged yet, is there something else you need from me?
I know there's possibly an underlying issue, but the refactor will at least take care of the memory leak for now.
I've updated the branch just now to allow merging again.

oroulet:
Rebase on master and i will have a look again. I need to understand that stuff

Randelung:
The "update branch" on github merged master commits, so the rebase didn't work. The force push just now closed the pull request. I'll make a new one.

Randelung:
Wait, did you merge it at the same time? It says "successfully mergend and closed", I thought it was an error.
Are we sure it merged the correct thing now?

PR #112, opened 2019-12-16
Implemented Alarms and Conditions

mar-ar:
Hi guys,

still one big chunk... sorry but everything belongs at least a bit together.
All tests work.
#105 is also included

If you delete a GeneratesEvent or HasNotifier Reference you should know what you do but I think we should assume that.

Sorry that it took that time but I had no time.

PR #365, opened 2020-12-18
[WIP] Alarm conditions1

JoeyFaulkner:


AndreasHeine:
i am quite deep into part 9 and 10 i will have a look on that 🧐 in the evening.

Koseng:
Hi, I'll also give some more detailed feedback today or tomorrow. I think the handling and understanding of the ConditionId is not correct. The ConditionId is not related to the emitting node.

asyncua/common/event_objects.py
```diff
@@ -292,6 +292,12 @@ class TransitionEvent(BaseEvent):
     def __init__(self, sourcenode=None, message=None, severity=1):
         super(TransitionEvent, self).__init__(sourcenode, message, severity)
         self.EventType = ua.NodeId(ua.ObjectIds.TransitionEventType)
+        self.add_property('Transition/Id', None, ua.VariantType.Variant)
+        self.add_variable('Transition', None, ua.VariantType.LocalizedText)
+        self.add_property('FromState/Id', None, ua.VariantType.Variant)
+        self.add_variable('FromState', None, ua.VariantType.LocalizedText)
+        self.add_property('ToState/Id', None, ua.VariantType.Variant)
+        self.add_variable('ToState', None, ua.VariantType.LocalizedText)
```
oroulet, line 300: might be stupide, but can't we generate that file from some xml?

AndreasHeine: @oroulet thats what i mentioned on gitter about event_objects.py as i worked on TransitionEvents missing fields

Koseng:
Hi, as mentioned I think the ConditionId is not yet handled correctly. I know this is the most confusing part about the whole Alarms/Conditions. Sorry for so much text, hope it helps. Maybe it's already clear for everyone, but in my experience most of the time there is some confusion.

**So how does the basic mechanism of Alarms/Conditions work?**

At some point in time an Alarm "appears", later after some time it "disappears". In OPC UA that is modeled with two events, one for the alarm appearing (with Retain=True) and one for the alarm disappearing (with Retain=False). The client needs to monitor the event stream and figure out by itself which alarms still exist (those who did not yet get a Retain=False).

But how can the client correlate the two corresponding events? That is via the ConditionId. So each existing Alarm must have a unique ConditionId. When the Alarm "appears", that ConditionId with Retain=True is used. When at some point later in time the Alarm "disappears" exactly the same ConditionId with Retain=False must be used.

A lot of confusion comes from the fact, that the ConditionId is not an usual property or variable of the event. It is the NodeId of the Event. So yes, in that case, an event has a NodeId. 

**So let's summarize.** The two fields of the condition event as the foundation for Alarms/Conditions are:
- ConditionId (NodeId of the object. The unique identifier for that concrete Alarm, identifying the alarm.)
- Retain (Boolean which controls whether the Alarm appears or disappears.

https://reference.opcfoundation.org/v104/Core/docs/Part9/5.5.2/

You can test whether the server works correctly with UaExpert. On Document->EventView next to the Events Tab, there is the Alarms tab. There the alarm/conditions need to show up correctly. If you click on one Alarm you can see the details and the ConditionId as NodeId. 

I'll write another comment regarding the implementation and my complete alarms client example #359.

UaExpert showing alarms:

uaExpertAlarms

Koseng:
**Regarding the WIP and the client example #359**

Since they had the glorious idea to model the ConditionId as the NodeId of the event (and not as a simple property, would have been to easy), clients have to add a special SimpleAttributeOperand to the select clause of the event filter:

```
conditionIdOperand = ua.SimpleAttributeOperand()
conditionIdOperand.TypeDefinitionId = ua.NodeId(ua.ObjectIds.ConditionType)    
conditionIdOperand.AttributeId = ua.AttributeIds.NodeId 
```
So every client interested in Alarms/Conditions has to do that. See bottom of https://reference.opcfoundation.org/v104/Core/docs/Part9/5.5.2/.

**The really cool thing is, that does already work correctly when python asyncua is used with a proper Alarm server.** The auto generated event does contain an auto generated field "NodeId", due to the SimpleAttributeOperand.

On the left side of the screenshot is the Event-Object in the debugger. It has the generated NodeId field, which is the ConditionId.
eventObject

**What does that mean for the Python Alarms server?**

- The server needs to correctly react to the SimpleAttributeOperand with a.AttributeIds.NodeId in the event filter select clause. If that AttributeOperand does exist, the server needs to return the events NodeId  as part of the event. If it does not exist, it should not be part of the event. 
-  You should think about how to name that field in the event. Name=NodeId as with the autogenerator. Or Name=ConditionId. I think both is possible.
- I did not not test the python server in the pull request regarding that.

So before all those proposed changes for the events only properties where implemented (I added variables in #361, but there still the underlying properties are missing) . But events can also contain variables (with again underlying properties) and can have a NodeId. The variables are needed for other cases as well like the above mentioned state machines.

_Before:_
```
Event
- Property 1
- Property 2
```

_Needed:_
```
Event
- NodeId
- Property1
- Property2
- Variable1
      - Prop1
      - Prop2
- Variable2
       - Prop1
       - Prop2
 - Property3
```
Now the event object has/needs a hierarchy. Since the implementation of the event object so far is a flat structure, the hierarchy is proposed to be adapted within the names using "/" as separator (in the code of the pull request). Like in the Details of the UaExpert screenshot.
```
NAME                         
- Variable1 = same value                     
- Variable1/Prop1  = same value        
- Variable2/Prop2  = same value      
```
**What does it mean for the python client?**
Since everyone has to change the select clause of the event filter. Maybe it is a good idea to not only have _sub.subscribe_events(...)_, but also _sub.subscribe_conditions(...)_ and automatically add the necessary AttibuteOperand in subscribe_conditions.

**Ideas going forward:**

- Maybe it is a good idea to split that huge effort into several check-ins. 
- You could start with everything in  asyncua/common/ and asyncua/schemas. (Already useable for clients)
- Then step by step do the server stuff.

asyncua/common/event_objects.py
Koseng: @oroulet , @AndreasHeine 
I think the code for generating the file is in the pull request. In schemas/generate_event_objects.py and schemas/generate_model_event.py. It is just so much stuff in the pull request.

oroulet: do you mean that this file is autogenerated? I do not see any header saying so...

AndreasHeine: > I do not see any header saying so...

because its modified?

AndreasHeine: it has been modified not regenerated... thats why i added the datestamp last time see line 1

Koseng: No, I did not mean the file is autogenerated. I just wanted to say there is code to generate it.
=> So probably that file shouldn't be merged like this in the pull request. It should be newly generated and then added as a whole.

examples/server-alarms-and-conditions.py
```diff
@@ -0,0 +1,107 @@
...
+        self.con_gen.event.add_property('NodeId', con_obj.nodeid, ua.VariantType.NodeId)
+
+        alarm_obj = await noti_node.add_object(idx, "AlarmObject")
+        alarm = self.server.get_node(ua.NodeId(10637))
+        self.alarm_gen = await self.server.get_event_generator(alarm, self.server.nodes.server,
+                                                               notifier_path=[ua.ObjectIds.Server])
+
+    async def generate_condition(self, retain):
+        self.con_gen.event.ConditionName = 'Example Condition'
+        self.con_gen.event.Message = ua.LocalizedText("Some Message")
```
Koseng, line 51: Add missing ConditionId:
```suggestion
    async def generate_condition(self, retain):
        self.con_gen.event.NodeId = ua.NodeId("Id111_EC") # Unique ConditionId
        self.con_gen.event.ConditionName = 'Example Condition'
        self.con_gen.event.Message = ua.LocalizedText("Some Message")
```

examples/server-alarms-and-conditions.py
```diff
@@ -0,0 +1,107 @@
...
+        self.con_gen.event.BranchId = ua.NodeId(0)
+        if retain == 1:
+            self.con_gen.event.Retain = True
+        else:
+            self.con_gen.event.Retain = False
+        await self.con_gen.trigger()
+
+    async def generate_alarm(self, active):
+        self.alarm_gen.event.ConditionName = 'Example Alarm1'
+        self.alarm_gen.event.Message = ua.LocalizedText("hello from python")
```
Koseng, line 62: Add missing ConditionId:
```suggestion
    async def generate_alarm(self, active):
        self.alarm_gen.event.NodeId = ua.NodeId("Id222_EA") # Unique ConditionId
        self.alarm_gen.event.ConditionName = 'Example Alarm1'
        self.alarm_gen.event.Message = ua.LocalizedText("hello from python")
```

Koseng (changes requested):
Hi guys,
this is my first ever github review. So if I should it handle somehow differently, please feedback. I made now some tests with the examples and a lot of stuff is already working well.

Here the necessary changes for the ConditionId on server side. In the client example still the extended select_clause of the event filter is missing.

oroulet:
@Koseng maybe this is your first review, but anyway looks like you had many good points ;-) good work

Koseng:
**ConditionRefresh Method**

Now I tested the ConditionRefresh() Method. It seems it is not yet fully working as intended. The whole idea of ConditionRefresh() is, when a client comes late to the game, he wants to be informed about the existing Alarms/Conditions and calls therefore ConditionRefresh().

Case 1:
I run the server example and create conditions before any client is connected. Now when I connect a client which calls ConditionRefresh() it does not get the existing Conditions. 

Case 2:
I run the server example and first connect the client. Now when I create conditions and call afterwards ConditionRefresh(), then it works as intended. But the usual scenario is Case 1.

=> It looks like the existing conditions are saved in a subscription. But they also need to be saved if there is no client subscription at all. So there we need a change in implementation.

examples/server-alarms-and-conditions.py
AndreasHeine: `self.alarm_gen.event.NodeId = ua.NodeId("Id222_EA") # Unique ConditionId`
this ConditionId topic confuses me if its not like:
a condition is usually a instance of a condition type in the addressspace! so why not use the instance's nodeid?

Koseng: The ConditionId is the identifier of an alarm/condition. A device, or machine can have multiple existing alarms at the same time. How to distinguish them from each other? The client gets an ConditionEvent with Retain=False. How does the client know which of the 10 existing alarms/conditions is not existing anymore? You need an Id for that. It is like the primary key in a database. For that purpose there is the ConditionId, and for for that reason it also got its name.  

It is completely up to the server after which rules he creates the ConditionId, as long as each existing condition has a unique ConditionId. You could just use GUIDs. Or for example on a machine tool you could create a combined string using the alarm number and some parameters, as long as you know, that combination can exist only once at a time.

Why not use the instance's nodeid? Because something like that does not exist. A condition/alarm is an event. An event only exists temporarily and is gone. An event does not exist in the adresspace like a node. There is only a node in the addresspace where you can register for the events. An event is not a node and therefore there is no "instance's nodeid".

Because there is no "instance's nodeid", they added something with a similar meaning and called it ConditionId (of type NodeId). How to fill it, is up to the server.

AndreasHeine: > Why not use the instance's nodeid? Because something like that does not exist. A condition/alarm is an event. An event only exists temporarily and is gone.

this is not true, conditions are more then just events!

for example a implementation of a levelalarm which is a subtype of ConditionType ->AcknowledgeableConditionType -> AlarmConditionType -> LimitAlarmType -> ExclusiveLimitAlarmType -> "ExclusiveLevelAlarmType":
Unbenannt
mylevelalarm has the node id "ns=5;s=MyLevel.Alarm" which is unique
Unbenannt1
Unbenannt2

> It is completely up to the server after which rules he creates the ConditionId, as long as each existing condition has a unique ConditionId. You could just use GUIDs. Or for example on a machine tool you could create a combined string using the alarm number and some parameters, as long as you know, that combination can exist only once at a time.

there might be no rule but it is a good idea to look on the opc foundation defined conditions and there implementation  / usage!

you can also look into the "Unified Automation Cpp Demo Server" the same game like the prosys demo server!

Koseng: Hi,
I know real world servers where it is done as I described. And you're right there seem to be implementations as you describe as well.

It seems there are two common approaches, the servers I know use approach 2.

1. Create a node for each possible Condition/Alarm instance in the address space. Maybe that makes more sense, if you have only a couple of different Condition/Alarms or you have only a few very specialised Alarms like the ExclusiveLevelAlarm.

2. Do not create a node for each possible Condition/Alarm instance in the address space. Manage the conditions in a server internal data structure. Maybe that makes more sense, when you have thousands of different alarms in your system. Otherwise you would need to create thousands of nodes.

It is interesting, the specification also always describes two ways to do it and has examples for both approaches.
https://reference.opcfoundation.org/v104/Core/docs/Part9/6.3/
https://reference.opcfoundation.org/v104/Core/docs/Part9/B.2/

In January I can talk to a member of the opc ua technical council and get more clarification. That is the really hard part of OPC UA. They allow so many variants, there are so many ways to do something and the specification is huge.

asyncua/common/event_objects.py
JoeyFaulkner: aha! this is my mistake, thank you for pointing it out! can you point me towards how this file is generated? as i did just edit it by hand :)

JoeyFaulkner:
@Koseng firstly thank you very much for all your feedback! For your first review it was fantastic, I learned a lot and understand much more about alarming and conditions now. 

To give some context, in this PR I am rebasing #112 (from @mar-ar ) and trying to pass tests/update deprecated code. I would really like to see alarming capabilities merged into master here, but I'm very new to the OPC UA Alarming + Conditions specification and still quite new to this codebase. I am not sure how best to proceed, I could attempt to implement the comments here (which i will with pleasure, as it seems i will learn a lot in the process)  or if someone feels like they have the drive (and critically better understanding) to implement the changes listed then that could also work.

examples/server-alarms-and-conditions.py
AndreasHeine: to get both 1. and 2. we need a class for each condition and a method to install them to the addressspace (in this case we need to use the node id) at the end of install we call init method that would be 1. 
the other way (2.) we only call init and give them an condition id
the key is here to have all functionality in a condition-type-class and make the addressspace implementation optional

swamper123: @Koseng Afaik: 1) is clear I guess, with every Node has it's own instance. 
In version 2) the ConditionSource (example: LevelMeasurement in a Tank) reference to the AlarmType. Via the _SourceNode Property_ (known from the BaseEventType) the _ConditionSource_ is known as well and an Alarm is triggered. So many Nodes may share the same Alarm, is that correct?

If yes, 2) may realy be a solution to spare and hide extra Nodes, even if 1) is more intuitive (in my opinion). We still have to deal with both, since Variant 3) (HasCondition used with an instance declaration) is wanted as well.

So it isn't necessarry to add an additional ConditionId, just the HasCondition Reference and the SourceNodeProperty attribute of the Event is needed?!

And if you want to use the condition in special domains, the ConditonClassId and ConditionSubClassId are useable.

Koseng: [https://reference.opcfoundation.org/v104/Core/docs/Part9/4.2](https://reference.opcfoundation.org/v104/Core/docs/Part9/4.2)
“Condition instances are specific implementations of a ConditionType. It is up to the Server whether such instances are also exposed in the Server’s AddressSpace. ”

I discussed the topic with Mr. Damm. Why are there two ways in the specification, one exposing the Condition-Instances in the AddressSpace, and the other way without exposing them? He said the reason is, because there are systems like the SIMATIC S7 PLC, where the server just doesn't know which relevant alarms/conditions for the system do exist. There are thousands of different alarms, created on the fly, even by different subsystems. Sometimes alarms are also dynamically build together. 

So, if you are in a scenario where you have a couple of alarms/conditions, which you know exactly, you should create the condition instances on startup and expose them in the AddressSpace. That is more common in the process industry, for example for some alarms of a tank. If possible, this is the preferred way, since the client can follow the OPC UA philosophy of browsing the AddressSpace and the client can find out which conditions do exist.

If you are in a scenario with a generic PLC like the SIMATIC S7 or with some controls of machine tools, you need to handle the Condition-Instances in a very dynamic way, since you don't know the conditions beforehand. For that scenario, you don't expose the condition instances in the AddressSpace. If conditions occur, only then the instances are created on the server dynamically. If the alarm/condition disappears, you can get rid of the instance as well. Also the ConditionId=NodeId is created dynamically, just for the lifetime of the condition instance. Usually then, many/all alarms are of the same type.

Then I asked, how to call a method, for example on an instance of AcknowledgeableConditionType, if the instance is not exposed in the AddressSpace? Because in my mind you always need a node from the addressSpace, where you can call the method on. He said that is not correct. A method call has 2 parameters. An ObjectId and a MethodId. In that case the client can execute a method call using the ConditionId as ObjectId.

So in the case of the python SDK, the client would directly call `async def call_method(parent, methodid, *args)` in `methods.py` without using a node object. “Parent” would be the ConditionId. From that perspective, the parameter name “parent” can be a bit misleading. It is just the identifier for the object you call the method on. See implementation in `async def _call_method(server, parentnodeid, methodid, arguments):` where the parameters for the request are called ObjectId and MethodId.

@swamper123 
“So many Nodes may share the same Alarm, is that correct?”
I don't exactly understand what you mean by that. 

In the case of not exposing the conditions in the AddressSpace, you just have a node where you can register for all the alarm/condition events. For example on a SIMATIC S7 you could have a node “ControlAlarms” where you subscribe for the AlarmConditionType-Events. The only purpose of the node is that there is a node where you can subscribe to the events of that type. Then if different alarms occurs, the alarm instances are created dynamically on the server side without exposing in the addressspace and the events are sent. Which each event you get the ConditionId=NodeId to keep track of existing alarms. Also the client can use the ConditionId to call a method on the internal condition instance.

Koseng: So to summarize:
- You need to create Condition-Instances on the server. The unique id of a Condition-Instance is the ConditionId of type NodeId.
- There are scenarios where the Condition-Instances are created and removed dynamically and not exposed in the AddressSpace
- There are scenarios where Condition-Instances are created static on startup and exposed via the AddressSpace.

AndreasHeine: tbh have never seen a system which dont expose its alarms... is there a demoserver out there?
S7-1500 only support the dataaccess (read/write/subscribe/methods and since fw2.6 client functionality) maybe he spoilt siemens dev status ^^ 

i agree with the philosophy part but i also understand the usecase not exposing that alarm...

swamper123: > In the case of not exposing the conditions in the AddressSpace, you just have a node where you can register for all the alarm/condition events.

That was the answer to my question, thanks.

Koseng: @AndreasHeine I'll sent you an email.

Koseng:
Hi guys,

that PR here is just too huge. I did a first step and made a PR for the stuff in "common" and "schemas". See #415. Also clients will benefit from those changes.

After rebasing the changes of the following files can be removed.
   - events.py
   - generate_model_event.py
   - generate_event_objects.py

Another PR will handle event_objects.py

dremsol:
Instead of creating a new issue I think that my problem best fits this discussion. Currently we are building an A&C client which communicates to a S71500 server. We would like to acknowledge alarms using a method call on the ConditionID (as S71500 doesn't expose the node in Adress Space). Using UAexpert this works as expected (using event view) and a Wireshark snippet of the request looks as follows;
```
CallRequest
    RequestHeader: RequestHeader
    MethodsToCall: Array of CallMethodRequest
        ArraySize: 1
        [0]: CallMethodRequest
            ObjectId: NodeId
                .... 0101 = EncodingMask: Opaque (0x5)
                Namespace Index: 3
                Identifier ByteString: d9f4db127731b642af538b5c6943f509
            MethodId: NodeId
                .... 0001 = EncodingMask: Four byte encoded Numeric (0x1)
                Namespace Index: 0
                Identifier Numeric: 9111
            InputArguments: Array of Variant
                ArraySize: 2
                [0]: Variant
                    Variant Type: ByteString (0x0f)
                    ByteString: 3dd4dc4d11338340b110a8791dd10f2f
                [1]: Variant
                    Variant Type: LocalizedText (0x15)
                    Value: LocalizedText
                        EncodingMask: 0x03, has locale information, has text
                        Locale: en-US
                        Text: Succeed
```

However, when trying to implement the method call with this library we get the following error;
`"The requested operation has no match to return."(BadNoMatch)`

A simpler way to reproduce this is by using the CLI command `uacall` with the conditionid as the parameter to nodeid agument (this is the ObjectId in the above Wireshark snippet). Which results in following error suggesting that the conditionid based search is not able to find the specific condition.
`"The node id refers to a node that does not exist in the server address space."(BadNodeIdUnknown)`

Based on the above discussion I'm not sure if conditionid based method calls are currently supported. Maybe somebody can elaborate on that point?

dremsol:
Thanks for your answer. I agree that it should be simple and I'm fairly sure that I use the correct ConditionId as I just copy it from UAExpert for the current active alarm. 
image

So in this case the command will be looking like
`uacall -u <url> -v INFO -n b=d962c15325443a4ea705db6a4978591d -i 3 --user <user>--password <password>-m 0:Acknowledge <inputs>`

schroeder-:
The [Acknowledge](https://reference.opcfoundation.org/search/37?t=Acknowledge) [Method](https://reference.opcfoundation.org/search/37?t=Method) is used to acknowledge an [Event Notification](https://reference.opcfoundation.org/search/37?t=Event%20Notification) for a [Condition](https://reference.opcfoundation.org/search/37?t=Condition) instance state where [AckedState](https://reference.opcfoundation.org/search/37?t=AckedState) is False. Normally, the [NodeId](https://reference.opcfoundation.org/search/37?t=NodeId) of the object instance is passed as the [ObjectId](https://reference.opcfoundation.org/search/37?t=ObjectId) to the [Call](https://reference.opcfoundation.org/search/37?t=Call) [Service](https://reference.opcfoundation.org/search/37?t=Service). However, some [Servers](https://reference.opcfoundation.org/search/37?t=Servers) do not expose [Condition](https://reference.opcfoundation.org/search/37?t=Condition) instances in the [AddressSpace. ](https://reference.opcfoundation.org/search/37?t=AddressSpace.)Therefore, [Servers](https://reference.opcfoundation.org/search/37?t=Servers) shall allow [Clients](https://reference.opcfoundation.org/search/37?t=Clients) to call the [Acknowledge](https://reference.opcfoundation.org/search/37?t=Acknowledge) [Method](https://reference.opcfoundation.org/search/37?t=Method) by specifying [ConditionId](https://reference.opcfoundation.org/search/37?t=ConditionId) as the [ObjectId.](https://reference.opcfoundation.org/search/37?t=ObjectId.) The [Method](https://reference.opcfoundation.org/search/37?t=Method) cannot be called with an [ObjectId](https://reference.opcfoundation.org/search/37?t=ObjectId) of the [AcknowledgeableConditionType](https://reference.opcfoundation.org/search/37?t=AcknowledgeableConditionType) [Node](https://reference.opcfoundation.org/search/37?t=Node).

schroeder-:
Ok I see this is what you did, but I am not sure if uacall  can parse byte nodeids correct. Maybe this is a issue.

schroeder-:
Also BadNoMatch match means mostly you got the wrong EventId, i would guess.

dremsol:
Thanks for pointing me in the right direction. I make a call to `client.nodes.objects.call_method()`. However, the call should be made on the ConditionID/NodeID returned by the event as in following [example](https://github.com/FreeOpcUa/opcua-asyncio/blob/9aad5a4d6d4c079f58c0050b97b683951edba5b3/examples/client_to_uanet_alarm_conditions.py#L22).

Could you tell me how to make the proper method_call on an conditionid which is not exposed in address space?

schroeder-:
You have to use a NodeId as first parameter in call_method. A browse_name requires the node to be in the addresspace.
```Python
node = client.get_node(condition_id)  # id recived from the condition 
await node.call_method(ua.NodeId(ua.AlarmConditionType_Acknowledge), event_id, ua.LocalizedText())  # event_id recived from condition

```

dremsol:
Works like a charm, thanks for your help @schroeder-

Koseng:
For event subscriptions an eventfilter is needed. If no eventfilter is given, it is auto generated to provide all event attributes. In case of conditions there is one very special attribute (ConditionId) which needs a very special entry in the eventfilter. That eventfilter entry is auto added on 'subscribe_alarms_and_conditions'.

PR #366, opened 2020-12-18
Fix uals -l asyncio error

kohtala:
uals fails with -l option and subscription loop did not properly handle task completion.

### DIFF
--- a/asyncua/server/server.py
+++ b/asyncua/server/server.py
@@ -46,7 +46,7 @@ class Server:
     Create your own namespace and then populate your server address space
     using use the get_root() or get_objects() to get Node objects.
     and get_event_object() to fire events.
-    Then start server. See example_server.py
+    Then start server. See server-example.py
     All methods are threadsafe
 
     If you need more flexibility you call directly the Ua Service methods
@@ -75,7 +75,6 @@ class Server:
 
     def __init__(self, iserver: InternalServer = None, loop: asyncio.AbstractEventLoop = None, user_manager=None):
         self.loop: asyncio.AbstractEventLoop = loop or asyncio.get_event_loop()
-        _logger = logging.getLogger(__name__)
         self.endpoint = urlparse("opc.tcp://0.0.0.0:4840/freeopcua/server/")
         self._application_uri = "urn:freeopcua:python:server"
         self.product_uri = "urn:freeopcua.github.io:python:server"
--- a/asyncua/tools.py
+++ b/asyncua/tools.py
@@ -329,7 +329,7 @@ async def _uals():
             elif args.long_format == 1:
                 await _lsprint_1(node, args.depth - 1)
             else:
-                _lsprint_long(node, args.depth - 1)
+                await _lsprint_long(node, args.depth - 1)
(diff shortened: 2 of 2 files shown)


kohtala:
I had these in my workdir. I lost the actual error message the subscription loop had and find little time to make clean pull request. Just want to make them available.

asyncua/server/internal_subscription.py
```diff
@@ -50,6 +50,13 @@ async def start(self):
         if self.data.RevisedPublishingInterval > 0.0:
             self._task = self.loop.create_task(self._subscription_loop())
 
+            @self._task.add_done_callback
+            def subscription_loop_done(task):
+                if not task.cancelled():
+                    e = task.exception()
+                    if e is not None:
+                        self.logger.error("Exception in subscription loop", exc_info=e)
+
```
oroulet, line 59: It is the first time I see that kind of code: using callbacks just to display exceptions.
The original implementation was more reable. ButI know that sometimes exceptions in asyncio code are swallowed, I thought it was a python bug but maybe we do something wrong...

asyncua/server/server.py
```diff
@@ -75,7 +75,6 @@ class Server:
 
     def __init__(self, iserver: InternalServer = None, loop: asyncio.AbstractEventLoop = None, user_manager=None):
         self.loop: asyncio.AbstractEventLoop = loop or asyncio.get_event_loop()
-        _logger = logging.getLogger(__name__)
```
oroulet, line 78: that line should be moved out of class, not removed

asyncua/tools.py
```diff
@@ -383,7 +383,7 @@ def _lsprint_long(pnode, depth, indent=""):
         else:
             print("{0}{1:30} {2:25} {3:25}".format(indent, name.to_string(), bname.to_string(), node.nodeid.to_string()))
         if depth:
-            _lsprint_long(node, depth - 1, indent + "  ")
+            await _lsprint_long(node, depth - 1, indent + "  ")
```
oroulet, line 386: the rest is good. This is a valid fix

oroulet:
also tests are broken b these changes...

asyncua/server/internal_subscription.py
kohtala: I had this in my worktree for some time already, but unfortunately can not be sure why I wrote it. I recall something. During application startup the start had been called but then something failed and the reference was lost and task was collected without event loop getting a chance to run the exception handler.

I usually make sure the exception from created tasks are handled at the task creator rather than in the task itself. But I am not sure now if it is supposed to allow collecting the exception in this case also. Do this may not be a solution but an incomplete draft.

kohtala: I did not find nor remember why I was making this change. The function is marked low level, so no reason to start using it.

asyncua/server/server.py
kohtala: It was already out of class. This was duplicate.

asyncua/tools.py
kohtala: I'll put it in own commit and push new version.

Issue #314, opened 2020-10-14
[Topic] Support for HA Client


ZuZuD:
I wonder if the community has already considered integrating a high-availability client like described in the `Part 4 section 6.6.2.4.5` of the protocol specification to this library.

For quite some times now, I worked with a HA Client based on the python-opcua client. I am currently rewriting this on top of asyncua and I aim to support the ```Hot``` mode (```6.6.2.4.5.4```) and from there, the ```Warm``` (```6.6.2.4.5.3```) mode shouldn't be too much of a lift.

In a nutshell, the spec describes a higher level client which establishes connections to two identical servers, and promote the server with the highest healthy service level (> 200) as the primary. The difference between primary and secondary being the publishing mode (enabled/disabled) for the subscriptions. Also, the sampling must be activated on the secondary for `Hot`, but disabled for the`Warm` mode.

I plan to have an abstraction layer with an API similar to the Client. A user could then leverage it to create an ideal configuration that will be subsequently applied to the lower clients according to their role. We could use several tasks to:

1. Check the service level
2. Promote a primary and reconnect clients individually 
3. Regularly merge the ideal and real configurations when needed.

I would like to know if there's people interested with such a feature and if this could be a good fit within the project. If that's the case, I will be able to suggest some PR soon to discuss the details.

swamper123:
Since we are striving to a full OPC UA implementation, we would welcome every PR that brings us a step further.

This kind of Client/Behaviour is new for me so I may need some time to read and (hopefully) understand the specs (https://reference.opcfoundation.org/v104/Core/docs/Part4/6.6.2/#6.6.2.4).

ZuZuD:
> Since we are striving to a full OPC UA implementation, we would welcome every PR that brings us a step further.
> 
> This kind of Client/Behaviour is new for me so I may need some time to read and (hopefully) understand the specs (https://reference.opcfoundation.org/v104/Core/docs/Part4/6.6.2/#6.6.2.4).
> 
> EDIT: I am not even sure if we may have a handling for ServiceLevels in the Server?!

I used a different server so far, however from the latest build it seems we do not support it:
```
In [1]: from asyncua import Client, Node, ua
    ...: c = Client(url="opc.tcp://0.0.0.0:4840")
    ...: await c.connect()
    ...: assert await c.get_node(ua.ObjectIds.Server_ServiceLevel).get_value() is None
```

swamper123:
[...]

PR #367, opened 2020-12-19
[HaClient] Asyncua ha client

ZuZuD:
See discussion #314. 

This adds a Client interface named HaClient which supports two or more servers for high availability purposes.
You'll find some examples and tests to guide you through this new feature.

Under the hood, it starts a keepalive task to monitor the status and service level for each server.

It also includes an HaManager task which promotes a primary and reconnect unhealthy clients. Unhealthy clients are detected based on the socket status and keepalive data collected.

The requests received by users as well as the HaManager decisions produce an ideal state called the "ideal_map". Regularly, the reconciliator task kicks in, lock this configuration, and applies the "ideal_map" configuration.
It then stores the actual subscriptions status into the "real_map" using the lower level components of the library.

Current limits:

- We only support HA WARM mode and datachange notifications (no event/status_change).
- We support multiple subscriptions, but a node should only be subscribed once.

Component details:

- HaClient:
    - Configuration based on dataclasses: HaConfig (required) and HaConfigSecurity (optional)
    - Responsible to start up sides tasks: keepalive, hamanager, reconciliator
    - Mutates the ideal_map via VirtualSubscription
    - Generic hooks (i.e: to monitor subscription performance)

- KeepAlive (task):
    - Regularly hits the server to check its service_level / status.

- HaManager (task):
    - Promote primary / secondaries
    - Reconnect disconnected/unhealthy client based on the keepalive
    feedback.

- Reconciliator (task):
    - Applies the ideal configuration
    - Perform checks on the call responses
    - Mutates the real_map

- VirtualSubscription:
    - Key component of the ideal and real maps.
    - Exposes an interface similar to the subscription but only
    store the settings.


oroulet:
Thanks. Obviously you have done a lot of work and the code looks very clean.
 But this is so much code and it looks so complicated that I am unsure what to do with it. I am afraid it will quickly becomes a huge code amount that is in repository but is very little used, gets broken and nobody can fix it...
 We can start by waiting a little and see if people use that kind of features

ZuZuD:
> Thanks. Obviously you have done a lot of work and the code looks very clean.
> But this is so much code and it looks so complicated that I am unsure what to do with it. I am afraid it will quickly becomes a huge code amount that is in repository but is very little used, gets broken and nobody can fix it...
> We can start by waiting a little and see if people use that kind of features

Thanks for your feedback @oroulet and I totally understand your point. To give you more context, I currently use this HaClient in production against more than 20 OPC-UA "clusters". Consequently, I'll likely stay an active contributor should you merge this. I may also know other contributors that are interested or already testing this feature. I will check with them if they can share their experience here to build up more support.

oroulet:
I closed it by mistake ;-)

asyncua/client/ha/ha_client.py
```diff
@@ -0,0 +1,566 @@
...
+        pass
+
+    async def hook_on_unsubscribe(self, **kwargs):
+        pass
+
+
+class KeepAlive:
+    """
+    Ping the server status regularly to check its health
+    """
```
oroulet, line 428: Isnt't that KeepAlice clas a (better) duplicate of the KeepAlive class in client.py? Could we merge them?

asyncua/client/ha/reconciliator.py
```diff
@@ -0,0 +1,403 @@
...
+                        monitoring=vs.monitoring,
+                        sub_name=sub_name,
+                    )
+                )
+                tasks_add.append(task)
+                # add the sub name to real_map
+        await asyncio.gather(*tasks_add, return_exceptions=True)
+        await asyncio.gather(*tasks_del, return_exceptions=True)
+
+    async def update_nodes(self, real_map, ideal_map, targets: Set[str]):
```
oroulet, line 211: several methods in that class are too long, they will be hard to maintain

asyncua/client/ha/utils.py
```diff
@@ -0,0 +1,35 @@
+import asyncio
+import hashlib
+import logging
+import pickle
```
oroulet, line 4: maybe that file should better be called ha_common or similar. utils is often a missused name

asyncua/client/ha/ha_client.py
ZuZuD: There was a KeepAlive thread in [python-opcua](https://github.com/FreeOpcUa/python-opcua/blob/master/opcua/client/client.py#L32-L78) which was replaced by an async [task](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/client.py#L394-L413). Since its role is to renew the secure_channel, its frequency depends on the `session_timeout` and `secure_channel_timeout`. However, users of the HaClient could want to check the server status more frequently. Keeping those tasks apart feels like the right SoC.

asyncua/client/ha/reconciliator.py
ZuZuD: Refactored in the following commit.

asyncua/client/ha/utils.py
ZuZuD: Renamed from `utils` to `common`.

Issue #371, opened 2020-12-22
Opcua File Example


jadamroth:
This isn't an issue, but to help anyone else who comes along looking for an example of reading and writing to Opcua File Objects like in the [opcua docs](https://reference.opcfoundation.org/v104/Core/docs/Part5/C.1/).

I didn't see any other examples or issues of anyone implementing it, so I thought it might be useful. I'm new to this library and don't know the proper structure or else I'd submit a pull request to add the feature. 

I didn't test the write method below because I didn't need it personally, but I think it should work (or at least be solvable if it isn't)

Thanks for the library

```python
from asyncua import ua

class OpcuaFile:
    """ class object for reading files in opcua """

    def __init__(self, client, node=None, index=None, name_of_node=None):
        """
        :param client: opcua client (e.g. async with Client(url=url) as client)
        :param node: node of file object (e.g. node = client.get_node("ns=2;s=nameOfNode")
        :param index: namespace, integer (ns=2)
        :param name_of_node: string, name of the node
        """
        self._client = client

        # set node
        if node is not None:
            self._node = node
        elif index is not None and name_of_node is not None:
            self._node = client.get_node("ns=" + str(index) + ";s=" + name_of_node)
        else:
            raise Exception("missing class init arguments")

    async def open(self, open_mode):
        """ open file method """
        open_node = await self._node.get_child("Open")
        arg = ua.Variant(open_mode, ua.VariantType.Byte)
        return await self._node.call_method(open_node, arg)

    async def get_size(self):
        """ gets size of config """
        size_node = await self._node.get_child("Size")
        size = self._client.get_node(size_node)
        return await size.read_value()

    async def read(self):
        """ read file method, returns contents as file """
        handle = await self.open(ua.OpenFileMode.Read.value)
        size = await self.get_size()

        read_node = await self._node.get_child("Read")
        arg1 = ua.Variant(handle, ua.VariantType.UInt32)
        arg2 = ua.Variant(size, ua.VariantType.Int32)
        contents = await self._node.call_method(read_node, arg1, arg2)
        await self.close(handle)
        return contents

    async def write(self, file_contents):
        """ writes contents to file """
        handle = await self.open(ua.OpenFileMode.Write.value)

        write_node = await self._node.get_child("Write")
        arg1 = ua.Variant(handle, ua.VariantType.UInt32)
        arg2 = ua.Variant(file_contents, ua.VariantType.String)
[...]

PR #372, opened 2020-12-23
Client Read File Transfer Method

jadamroth:
This is the pull request regarding the issue I created an example in issue #371 for [Opcua File Transfers](https://reference.opcfoundation.org/v104/Core/docs/Part5/C.1/) for opcua clients

I couldn't immediately get the write file method working as shown in the example, so I only included the read file method for now.

I can try to work on writes in the coming weeks when I have more time, in which case I'll submit another pull request

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -690,23 +690,10 @@ class Client:
         arg1 = ua.Variant(handle, ua.VariantType.UInt32)
         return await file_node.call_method(read_node, arg1)
 
-    async def read_file(self, node=None, index=None, name_of_node=None):
+    async def read_file(self, file_node):
         """
-        :param node: node of file object (e.g. node = client.get_node("ns=2;s=nameOfNode")
-        :param index: namespace, integer (ns=2)
-        :param name_of_node: string, name of the node
+        :param file_node: node of file object (e.g. node = client.get_node("ns=2;s=nameOfNode")
         """
-        # set file node
-        file_node = node
-        if index is not None and name_of_node is not None:
-            file_node = self.get_node("ns=" + str(index) + ";s=" + name_of_node)
-        if file_node is None:
-            raise Exception("Not enough parameters for file node")
-
-        return await self._read_file(file_node)
-
-    async def _read_file(self, file_node):
-        """ read file method, returns contents as file """
         handle = await self._open_file(file_node, ua.OpenFileMode.Read.value)
         size = await self._get_file_size(file_node)
 
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -23,7 +23,6 @@ _logger = logging.getLogger(__name__)
 class Client:
     """
     High level client to connect to an OPC-UA server.
-
     This class makes it easy to connect and browse address space.
     It attempts to expose as much functionality as possible
     but if you want more flexibility it is possible and advised to
@@ -35,11 +34,9 @@ class Client:
         :param url: url of the server.
             if you are unsure of url, write at least hostname
             and port and call get_endpoints
-
         :param timeout:
             Each request sent to the server expects an answer within this
             time. The timeout is specified in seconds.
-
         Some other client parameters can be changed by setting
         attributes on the constructed object:
         See the source code for the exhaustive list.
@@ -110,17 +107,13 @@ class Client:
     async def set_security_string(self, string: str):
         """
         Set SecureConnection mode.
-
         :param string: Mode format ``Policy,Mode,certificate,private_key[,server_private_key]``
-
         where:
-
         - ``Policy`` is ``Basic128Rsa15``, ``Basic256`` or ``Basic256Sha256``
         - ``Mode`` is ``Sign`` or ``SignAndEncrypt``
         - ``certificate`` and ``server_private_key`` are paths to ``.pem`` or ``.der`` files
         - ``private_key`` may be a path to a ``.pem`` or ``.der`` file or a conjunction of ``path``::``password`` where
           ``password`` is the private key password.
-
         Call this before connect()
         """
         if not string:
@@ -542,10 +535,8 @@ class Client:
         """
         Create a subscription.
         Returns a Subscription object which allows to subscribe to events or data changes on server.
-
         :param period: Either a publishing interval in milliseconds or a `CreateSubscriptionParameters` instance.
             The second option should be used, if the asyncua-server has problems with the default options.
-
         :param handler: Class instance with data_change and/or event methods (see `SubHandler`
             base class for details). Remember not to block the main event loop inside the handler methods.
         """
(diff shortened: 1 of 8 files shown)


asyncua/client/client.py
```diff
@@ -672,5 +672,50 @@ async def write_values(self, nodes, values):
...
+    async def read_file(self, node=None, index=None, name_of_node=None):
+        """
+        :param node: node of file object (e.g. node = client.get_node("ns=2;s=nameOfNode")
+        :param index: namespace, integer (ns=2)
+        :param name_of_node: string, name of the node
+        """
+        # set file node
+        file_node = node
+        if index is not None and name_of_node is not None:
+            file_node = self.get_node("ns=" + str(index) + ";s=" + name_of_node)
```
AndreasHeine, line 702: What if the node has another identifier type
i | NUMERIC (UInteger)
s | STRING (String)
g | GUID (Guid)
b | OPAQUE (ByteString)
https://documentation.unified-automation.com/uasdkhp/1.0.0/html/_l2_ua_node_ids.html

AndreasHeine: cosmetics:
index is not None == not index

AndreasHeine:
is there a way you can provide a python server example aswell

asyncua/client/client.py
jadamroth: I removed the last 2 parameters (index, and name_of_node) and will just have user pass in their node as the only parameter.

I think this is easier to understand and less likely for user to make mistakes which will throw exception

jadamroth:
I will try to get to a working example of the server when I get some free time. I'll need to dive in to the source for server.

AndreasHeine:
i am working in statemachines atm in opc us spec there is a FileTransferStateMachine discribed maybe we can join forces !?
https://reference.opcfoundation.org/v104/Core/ObjectTypes/FileTransferStateMachineType/
https://reference.opcfoundation.org/v104/Core/docs/Part5/C.4.6/

jadamroth:
I apologize, but I'm not familiar with this spec and do not currently have enough time to commit to working on this right now.

If we run across any specs needed for our application that aren't including in this library, I'll commit to working on those.

I'll try to get the working server and write file methods in the next few days

AndreasHeine:
there is no need for apologize! i am just writen the statemachine code. the file topic is quite interesting for machines like milling-machines where you could deploy g-code files via opcua :)

oroulet:
I have an issue with that stuff. The Node and Client classes are already far too big and I am not really keen into adding new methods there for a functionality that is a bit suspicious. Why adding a file read to network protocol??? file read can be implemented as you want on sevrer side, using node value or custom methods. 

we could have it but then we need to find some smart way to move the code away in another file and have at most one metod in Client class...Ideally known by using some external class ala

class MyFileReader:
   def __init__(self, my_file_node):
        xxx

jadamroth:
You're correct in that it shouldn't have been in the base Client class. I am new to the library and don't fully understand the structure yet. I created a new client class ua_file.py, updated the example, and restored client class to its original.

Certainly people could write a custom method on the server, but I'm implementing based upon what the [opcua specs state](https://reference.opcfoundation.org/v104/Core/docs/Part5/C.2.1/). 

Please let me know of any other objections

jadamroth:
One thing that I think should be reconsidered is if the user should have more control over the open, close methods. 

For reads, you can use the same handle from open method several times before closing. However, I tried to make the code easier under the hood.

oroulet:
looks much better. that way you can add as much functionality as you want.
if you want to make even more fancy you can implement the `_aenter__, __aexit__`stuff so people can write
`async with Uafile(node) as f:
    f.read()
    `
    instead of you read_once method()

asyncua/client/ua_file.py
```diff
@@ -0,0 +1,48 @@
+from asyncua import ua
+
+
+class UaFile:
+
+    def __init__(self, file_node):
+        self._file_node = file_node
+
+    async def __aenter__(self):
```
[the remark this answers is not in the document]
jadamroth: I was thinking about opening it in this method, but I ultimately decided not to because I wanted to make one class to handle all file method operations (read, write, erase, append). For example, it wouldn't work well if a user wanted to open, read a file, then write a new file because when the client opens the file on our opcua server, we need to specify which operation we're using. 

I could create classes that are solely designated for each operation FileRead(), FileWrite(), FileErase() that inherits this UaFile class, and then it would make sense to open with this function and close on exit.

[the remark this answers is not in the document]
jadamroth: Check new commit for separating classes

oroulet:
looks fine, but I do not really see why you splitted these in two classes. One was fine.. I cannot remember any file API that differentiate between a File and a FileRead object. That would aso reduce the number of liens in example. But if you have reasons for that I am fine

jadamroth:
I personally don't like separating them into two classes. I was just showing it as an option to use open with __aenter__ .

I like keeping one class. However, with one class, I don't think you should call open() on the __aenter__ because you need to specify your file operation to the server on open(fileOperation) (which will be different for read, write, erase, and append)

oroulet:
How do they do with the built-in open() we should try to copy that api so we do not surprise anyone
So you can either use a function like open() or pass (optionally?) the open type to class constructor

jadamroth:
Here's an example of how the node-opcua public api handles their file operations, and how I'd recommend ours to similarly reflect - https://www.npmjs.com/package/node-opcua-file-transfer

oroulet:
@jadamroth The node-opcua guys choose (or had to implement due to limited language) a much lower level ua api. So I do not think they are a good reference

oroulet:
here is the standard (modern) python way to open a file

```
    with open('sample.txt', "r") as file_object:
        # read file content
        data = file_object.read()
``` 

I think we should do the same. something like
```
with asyncua.open(node, readtype) as uafile:
      data uafile.read()
      ```

**Comment by oroulet** [comment]
but your proposition (althoug with one class ) is fine too
```
   async with Client(url=url) as client:
        file_node = client.get_node("ns=2;s=NameOfNode")
        async with UaFile(file_node) as ua_file:
            # read file
            contents = await ua_file.read()
            print(contents)
```

**Comment by oroulet** [comment]
anyway that code looks good. I am wondering if you could remove code from Node and Client classes by writing such classes for other features. The entire server discovery is a good candidate. And history reading in Node

**Comment by jadamroth** [comment]
I would certainly think that you can separate classes for most features that have no required attributes internal to Node and Client classes

I still need to update the UaFile branch to revert back to one class. I like the idea of structuring the UaFile as pythonic. I'll commit changes tonight

**Comment by jadamroth** [comment]
Updated to make it pythonic and one class. 

I think it looks good for reads, but it might have to change when we include writes. Currently as is, if a user wants to handle different consecutive file operations, then they'll have to one of the following

1) create a separate UaFile class for writes. Something like this
```python
async with UaFile(file_node, 'r') as ua_file:
    contents = await ua_file.read()  # read file
    print(contents)
async with UaFile(file_node, 'w') as ua_file:
    await ua_file.write('new file contents')  # write file
```
2) handle the file logic internally themselves. Something like
```python
async with UaFile(file_node, 'r') as ua_file:
    contents = await ua_file.read()  # read file
    await ua_file.close()
    await ua_file.open('w')
    await ua_file.write('new file contents')
```

oroulet:
I think that limitation is completely fine. Do you often both read and open from the same file?
I am merging that one so you can always improve in another PR if you want

PR #387, opened 2021-01-05
Add python 3.8 to github actions

swamper123:


oroulet:
I am not sure about that one. Actions have an energy cost. I was thinking that if it works for 3.7 and latest then it works for all supported versions

swamper123:
I think you are right for the moment, but we have to watch carefully then for future releases, what versions we want to test. 
Think about the steps between 3.5 and 3.7, especially in the asyncio parts. 
I just thought you made a typo and forgot 3.8 to add. ^^

PR #388, opened 2021-01-05
export DataTypeDefinition to xml

oroulet:


asyncua/common/xmlexporter.py
```diff
@@ -142,6 +142,7 @@ async def node_to_etree(self, node):
         Returns:
         """
         node_class = await node.read_node_class()
+        print("EXPORT", node, node_class)
```
swamper123, line 145: Replace the print with a log info and add a whitespace at the end to seperate them a bit better. 
Perhaps formatted string here? Shouldn't be critical stuff.

asyncua/common/xmlexporter.py
```diff
@@ -295,8 +296,40 @@ async def add_etree_datatype(self, obj):
         Add a UA data type element to the XML etree
         """
         obj_el = await self._add_node_common("UADataType", obj)
+        dv = await obj.read_attribute(ua.AttributeIds.DataTypeDefinition)
+        sdef = dv.Value.Value
+        if sdef:
+            # FIXME: can probably get that name somewhere else
+            bname = await obj.read_attribute(ua.AttributeIds.BrowseName)
```
swamper123, line 303: I don't know where else the browsename should be stored, if not in the node attributes?! The passed obj is everytime a Node, isn't it?

asyncua/common/xmlexporter.py
swamper123, line 145: Wouldn't an info log be nicer? Otherwise there will be a lot of print spam.

oroulet:
merged in another branch

PR #391, opened 2021-01-09
fixes and xml export

oroulet:


asyncua/common/structures104.py
```diff
@@ -4,15 +4,18 @@
 from enum import IntEnum
 import logging
 import re
+from typing import Union, List, TYPE_CHECKING, Tuple
 
 from asyncua import ua
 from asyncua import Node
 from asyncua.common.manage_nodes import create_encoding, create_data_type
+if TYPE_CHECKING:
```
swamper123, line 12: Just a Question: What shall it does?

oroulet: good question. This seems to be the standard ay to solve circular imports created for typing only (The Server and Client import are only here for type hints, they are never run)

oroulet: quite ugly stuff though... but this is a python language limitation

PR #400, opened 2021-01-13
even more fixes

oroulet:


setup.py
```diff
@@ -1,6 +1,12 @@
 from setuptools import setup, find_packages
 import sys
 
+# read the contents of your README file
+from os import path
```
swamper123, line 5: I thought we are switching to pathlib, instead of using os.path?!

oroulet: hehe. It was quick copy/paste from a pypi recipe. good enough for now. Both are in standard lib anyway

PR #404, opened 2021-01-18
unlucky default value for ResultMask!

AndreasHeine:
@oroulet good that you asked... somehow i missed that only "nodeClassMask" 0 is returning all! sorry about that -_-°

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -675,6 +675,7 @@ class Client:
         for node in nodes:
             desc = ua.BrowseDescription()
             desc.NodeId = node.nodeid
+            desc.ResultMask = 63
             nodestobrowse.append(desc)
         parameters = ua.BrowseParameters()
         parameters.View = ua.ViewDescription()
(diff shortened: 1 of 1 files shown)


asyncua/client/client.py
```diff
@@ -675,6 +675,7 @@ async def browse_nodes(self, nodes):
         for node in nodes:
             desc = ua.BrowseDescription()
             desc.NodeId = node.nodeid
+            desc.ResultMask = 63
```
oroulet, line 678: would be great if you used an enum instead of hardcoced value. That mask must be defined as some enum somwhere in uaprotocol.py

oroulet:
mye bedre sånn ja ;-)

oroulet:
du burde forøvrog ha en test for denne nye metoden... og lage den samme på server siden antar jeg samt legge testen i test_common.py

swamper123:
@oroulet Wording! :^) Ikke alle kan snakke norsk. ^^

AndreasHeine:
hvordan kommer jeg til biblioteket? -_-°

oroulet:
shit. wrong language. I am not that good at not mixing things up. 
I see the german traduction is not that good. but I had a few typo so maybe google got confused

Issue #360, opened 2020-12-11
Getting "InvalidSignature" error when renewing secure channel when working with secure connexion and high server load


kaeros:
We might have found a problem on signature validation when working with secure connexions.

Issue occurs when the asyncua is connected with a security_policy equal to : Basic256Sha256.

When the server is facing high load and/or multiple simultaneous security channel renewal, an exception is raised in the verify method at line 375: (since Verifier is an abstract class, it is the VerifierHMac256 class which is triggered with this error)

  File "asyncua\client\ua_client.py", line 82, in _process_received_data
  File "asyncua\common\connection.py", line 333, in receive_from_header_and_body
  File "asyncua\common\connection.py", line 59, in from_header_and_body
  File "asyncua\crypto\security_policies.py", line 162, in verify
  File "asyncua\crypto\security_policies.py", line 375, in verify
  cryptography.exceptions.InvalidSignature

For info this is the code of the culprit: 

![image](https://user-images.githubusercontent.com/6350442/101933548-a7439480-3bdc-11eb-992f-6a1d472e74d4.png)

What we can see after investigating in this method is that there is a desynchronisation of signature when the check is performed during channel renewal. So for us there is 2 possible route causes for this to happen:

- The wrong signature which is invalidated by the method is a remnant of a previous signature which have not been updated
- The wrong signature which is invalidated by the method is a future signature which have not been updated yet in the "expected" variable

The thing is, unlike other opcua clients, there is no "retry" alogorithm if an Invalid Signature is detected. So in our case the code will automatically close the concerned subscription if that exception is raised. 

So my question is : do you have any idea why we are facing a signature desynchronisation during secure channel renewal ? And, if not, do you think a "retry behaviour" could be implemented on this specific case to be more robust in case of desynchronisation on signature verification ? Maybe one error could be raised as a warning but 2/3 multiple consecutive bad signature check will raise the InvalidSignature error which then close the subscription ? Maybe implement some kind of "current" and "pending" system on signature (quite like it is done for security tokens)

AsyncIO methodology is not yet my forte and the same thing could be said for openSSL's related subjects so I am open for suggestions to implement a retry algorithm or, even better, for a fix on that desynchronisation issue. 
If needed I can provide multiple screenshot of actual signature desynchronisation with their values (expected and tested). Not sure how it can be used since the data is encrypted and not related to a specific timestamp though. 

Thank you in advance for your time

Best regards

swamper123:
[...]

PR #409, opened 2021-01-21
[InvalidSignature] Fix race condition during secure channel renewal

ZuZuD:
Fix the racy secure_channel renewal issue reported in #360 

1. On secure_channel symmetric keys renewal, keep the previous Verifier/Decryptor in memory.
2. When `InvalidSignature` exception occurs, try to decrypt the message and verify the signature with the previous keys. 
3. For security concerns, delete the saved keys from memory as soon as they have expired.


asyncua/common/connection.py
```diff
@@ -228,23 +251,17 @@ def revolve_tokens(self):
...
-            message,
-            self._max_chunk_size,
-            message_type=message_type,
-            channel_id=self.security_token.ChannelId,
-            request_id=request_id,
-            token_id=self.security_token.TokenId,
-        )
+        chunks = MessageChunk.message_to_chunks(self.security_policy, message, self._max_chunk_size,
+                                                message_type=message_type, channel_id=self.security_token.ChannelId,
+                                                request_id=request_id, token_id=self.security_token.TokenId)
```
oroulet, line 264: please try to not change to much format ,it makes diff hard to read. In that case a trailing comma means all arguments on one line. This is supported by both yapf and black automated formatter. So let's keep it

oroulet: that makes formating robuts between editor and different max line length

asyncua/common/connection.py
```diff
@@ -6,14 +6,21 @@
 from asyncua import ua
 from ..ua.ua_binary import struct_from_binary, struct_to_binary, header_from_binary, header_to_binary
 
+try:
+    from ..crypto.uacrypto import InvalidSignature
+except ImportError:
+    class InvalidSignature(Exception):
+        pass
```
oroulet, line 13: I guess that ugly hack is for the case the crypto libraries are not installed? maybe we make these libraries required...

ZuZuD: Correct

ZuZuD:
I finally got the linter right.

oroulet:
merged

PR #415, opened 2021-01-23
Support variables with properties in events and extend file generators for Alarms and Conditions

Koseng:
This request is the first step towards alarm and conditions as in #365. It contains the basic changes in "common" and "schemas". It is based on the code of @mar-ar with some refactoring. In my previous commit, adding variables to events, still the subproperties of variables where missing.

Another pull request will contain the newly generated event_objects.py.

Those changes are the common basics for client and server. With the changes in that PR, pyhon opc ua clients should fully support alarms and conditions.

I successfully tested it with 2 different Alarms and Conditions Server, using the python client_to_uanet_alarm_conditions example.

So in respect to #365 all changes in the following files are done:
- events.py
- generate_model_event.py
- generate_event_objects.py

UPDATE: The PR now also contains:
- event_objects.py

After rebasing the changes in #365, those files can be removed.

When I find time, I'll write some Test Cases for Variables with Properties.


Koseng:
Ok, this PR now also contains the newly generated event_objects.py.

oroulet:
Can I squash all of these? There are bunch of trange commits. Or you want to clean up the history yourself?

Koseng:
@oroulet you can really do what you like and think is best. I'm not yet used to the git stuff.

asyncua/common/events.py
```diff
@@ -121,21 +139,29 @@ async def get_filter_from_event_type(eventtypes):
...
-                op.BrowsePath = [browse_name]
-                clauses.append(op)
-                selected_paths.append(browse_name)
-    return clauses
+        for property in await get_event_properties_from_type_node(evtype):
+            await append_new_field_to_select_clauses(property, select_clauses, None)
+        for variable in await get_event_variables_from_type_node(evtype):
+            await append_new_field_to_select_clauses(variable, select_clauses, None)
+            for subproperty in await variable.get_properties():
+                await append_new_field_to_select_clauses(subproperty, select_clauses, variable)
```
deeTEEcee, line 163: @Koseng Just an fyi, these variables can be nested further so my understanding is we currently don't support nested variables because this will only give us 1 level

PR #425, opened 2021-01-25
Updated Schemas and standard_addresspaces to newest UA-Nodeset version

swamper123:
Updated Schemas to latest UA-Nodeset version
Upgraded servers standard_addresspaces to latest UA-Nodeset version


AndreasHeine:
we need a workflow which runs every time opc foundation nodesets changes^^

swamper123:
Discussion in #424 :^)

asyncua/common/event_objects.py
```diff
@@ -1,5 +1,5 @@
 """
-Autogenerated code from xml spec. Date: 2021-01-23 15:07:13.892295
```
oroulet, line 2: we should really do something to not log only generation date but which spec branch is used

swamper123: Shall I add that in my PR or in a seperate one?

swamper123:
It seems that we have some runtime errors!

In the first run, an error was thrown from `test_remove_bad_nodes_from_both_maps`.
In the second run it happened in `test_secure_channel_key_expiration`...and python3.7 just ended.
Now the third run, everything is fine, without any changes made!

asyncua/common/event_objects.py
oroulet: as you want... maybe adding to that one is less work..

swamper123: Done

Koseng:
263 files changed?

AndreasHeine:
thats normal... it autogenerated code from opc foundation xml files :)

Koseng:
The PR now contains all the files of [https://github.com/OPCFoundation/UA-Nodeset](https://github.com/OPCFoundation/UA-Nodeset), not only the generated ones. Is that intended?

AndreasHeine:
oh your right!

swamper123:
@Koseng Ahm...that was not intended, indeed. 😅 
I will fix that (and add the new files, which have been merged today in UA-Nodeset).

asyncua/common/event_objects.py
oroulet: done? I only see a data here ...

asyncua/server/standard_address_space/standard_address_space_part10.py
```diff
@@ -2,8 +2,13 @@
 # -*- coding: utf-8 -*-
 """
 DO NOT EDIT THIS FILE!
-It is automatically generated from opcfoundation.org schemas. 
-Date:2020-06-15 19:56:58.216353
+It is automatically generated from opcfoundation.org schemas.
+
+Model Uri:http://opcfoundation.org/UA/
+Version:1.04.7
```
oroulet, line 8: ok here. nice

asyncua/common/event_objects.py
oroulet: we should add it for that one too, if not too much work

swamper123: Forgot that one. Will be added and made a bit nicer.

PR #426, opened 2021-01-25
fix _make_monitored_item_request in sync wrapper

daniel-brosche:
minor fix in sync wrapper due underlying API change


ZuZuD:
Good catch thanks!

oroulet:
@ZuZuD please use rebase so you do not get all this merge commits. Otherwise all good

PR #429, opened 2021-01-29
Auto generate client event filter for Alarms and Conditions

Koseng:
**Current solution for normal events:**
The current solution for subscribing to events is great. Setting the event-filter is optional. If not set, an event-filter containing all properties and variables, is auto-generated. The client coder does not need to have knowledge about the complex filter system. If he has, he can go fancy as well.

**Issue with the condition events:**
In a similar way that should be the case for an alarms/conditions client as well. But for subscribing to condition-events an extra operand in the select clause of the event filter is necessary (like in example _client_to_uanet_alarm_conditions_). Each client does need to add the same code, does need to cope with the filters, and just needs to realize that this has to be done. 

So it should be autogenerated as well. There are 2 possible approaches.

- **The implicit approach:**  
Stick to `subscribe_events` and when autogenerating the event-filter, check for the existence of a ConditionType. If a ConditionType is involved, auto-add the extra operand in the select clause.
- **The explicit approach**: 
Add `subscribe_conditions` to the subscriptions interface. `subscribe_conditions` does the same as `subscribe_events`, but adds the extra operand in the select clause.

@oroulet, which approach shall we follow? I tend a little bit more to the explicit approach, since I don't like too much implicit magic happening. But in that case I really think both approaches would be fine. So what do you prefer?

Code in the PR shows the explicit approach and how it simplifies everything for the client in _client_to_uanet_alarm_conditions_.

Related to #365

### DIFF
--- a/asyncua/common/subscription.py
+++ b/asyncua/common/subscription.py
@@ -211,13 +211,13 @@ class Subscription:
             evfilter = await self._create_eventfilter(evtypes)
         return await self._subscribe(sourcenode, ua.AttributeIds.EventNotifier, evfilter, queuesize=queuesize)
 
-    async def subscribe_conditions(self,
-                                   sourcenode: Node = ua.ObjectIds.Server,
-                                   evtypes=ua.ObjectIds.ConditionType,
-                                   evfilter=None,
-                                   queuesize=0) -> int:
+    async def subscribe_alarms_and_conditions(self,
+                                              sourcenode: Node = ua.ObjectIds.Server,
+                                              evtypes=ua.ObjectIds.ConditionType,
+                                              evfilter=None,
+                                              queuesize=0) -> int:
         """
-        Subscribe to condition events from a node. Default node is Server node.
+        Subscribe to alarm and condition events from a node. Default node is Server node.
         In many servers the server node is the only one you can subscribe to.
         If evtypes is not provided, evtype defaults to ConditionType.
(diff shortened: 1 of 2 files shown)


oroulet:
are explicit and implicit the right words here? just wondering. 
Anyway yes. I think adding subscribe_condition (or subscribe_alarm or both) makes the feature much discoverable for users and generally easier to use. People who want something special can always make their own filter.
So I am fine with that PR

oroulet:
I never really looked closely at alarms and conditions, but you may also want to have two methods subscribe_alarms and subscribe_conditions if they are different... this is just suggestion

Koseng:
From a technical point of view subscribe_conditions would be most correct, since alarms derive form conditions. But from a users point of view and to make clear when to use the method, subscribe_alarms_and_conditions is better. I agree, changed.

From my point of view ready to be merged.

Regarding the wording ;-):

After googling:

- Explicit means something is made clear and stated plainly.
- Implicit means something is implied but not stated directly.

I think that describes the differences of the two approaches in a nice way.

Koseng:
The AlarmConditionType stuff inherits from ConditionType. The addon in the filter is needed for the ConditionType (and since AlarmConditionType inherits for that as well). It's allways the same, so I think "subscribe_alarms_and_conditions" would be good. If you do something with alarms/conditions, just use that.

Koseng:
I don't mind. We can choose between:
- subscribe_conditions or
- subscribe_alarms_and_conditions

PR #435, opened 2021-02-05
Nodeset updater

swamper123:
Get newest UA-Nodeset update from github UA-Nodeset repo
Remove old Nodesets
Generate new Nodesets
WIth errorhandling
See Discussion in #424
Critical reviews are welcomed.


schemas/update_ua_nodeset.py
```diff
@@ -0,0 +1,56 @@
...
+    pass
+
+
+def get_new_nodeset(timeout=120):
+    if path.isdir("./UA-Nodeset"):
+        rename("./UA-Nodeset", "./UA-Nodeset-old")
+    elif path.isdir("./UA-Nodeset-master"):
+        rename("./UA-Nodeset-master", "./UA-Nodeset-master-old")
+    try:
+        check_call(["git", "clone", "https://github.com/OPCFoundation/UA-Nodeset.git", "UA-Nodeset-master"], timeout=timeout)
```
oroulet, line 20: git can have an option to only clone the last revision and not hte entire history. you should use it here

schemas/update_ua_nodeset.py
```diff
@@ -0,0 +1,56 @@
+from shutil import rmtree
+from os import unlink, path, rename
```
oroulet, line 2: here has pathlib a better api for all these things

schemas/update_ua_nodeset.py
```diff
@@ -0,0 +1,56 @@
...
+logger = logging.getLogger(__name__)
+
+
+class UpdateError(Exception):
+    pass
+
+
+def get_new_nodeset(timeout=120):
+    if path.isdir("./UA-Nodeset"):
+        rename("./UA-Nodeset", "./UA-Nodeset-old")
```
oroulet, line 16: maybe -safe instead of old

schemas/update_ua_nodeset.py
```diff
@@ -0,0 +1,56 @@
+from shutil import rmtree
+from os import unlink, path, rename
+from subprocess import check_call, CalledProcessError, TimeoutExpired
+from runpy import run_module
```
oroulet, line 4: did not know that one. Is that from standard lib? otherwise we should avoid 3rd party dependency just for that script

swamper123: [runpy ](https://docs.python.org/3/library/runpy.html?highlight=runpy#module-runpy)is in the standard lib. 
Haven't known of it before as well, but it does what it should do and doesn't make changes on the existing generate_*.py stuff.

swamper123:
That PR was converted to Draft, in cause of a problem with #436.

Since we got a problem to import part12, I took a closer look on that topic and it seems that[ importing each Part is not uptodate anymore.](https://opcfoundation.org/forum/opc-ua-standard/differences-and-relations-between-nodeset2-services-xml-nodeset2-xml-and-sum-of-nodeset2-part-xml/#p2867)

I will make a PR in the next days, which will clean up the part X.xml rests and replace them with the services.xml file. When accepted, that PR will be updated and converted to PR back again.

schemas/update_ua_nodeset.py
```diff
@@ -0,0 +1,70 @@
...
+        else:
+            rm_tree(child)
+    path.rmdir()
+
+
+def get_new_nodeset(timeout=120):
+    p_v1 = Path("./UA-Nodeset")
+    p_v1_backup = Path("./UA-Nodeset-backup")
+    p_v2 = Path("./UA-Nodeset-master")
+    p_v2_backup = Path("./UA-Nodeset-master-backup")
```
oroulet, line 26: This is a detail but I think you are not using the power of pathlib. Path should not be called so often for similar paths:

```
cwd = Path(".")
p_v1 = cwd / "./UA-nodeset-back"
p_v1_backup = cwd / "./UA-nodeset-backup"
etc
```

oroulet: ok even better so you do not repeat name and risk typo or using wrong directory

```
cw = Path.cwd()
target = cmd / "UA-nodeset-master"
backup = target.parent / target.name + "_backup"
etc...
```

schemas/update_ua_nodeset.py
oroulet: great

schemas/generate_model.py
```diff
@@ -21,13 +26,21 @@
                  "RegisterServerRequest", "RegisterServer2Response"]
 # structs that end with Request or Response but are not
 NotRequest = ["MonitoredItemCreateRequest", "MonitoredItemModifyRequest", "CallMethodRequest"]
-OverrideTypes = {}  # AttributeId": "AttributeID",  "ResultMask": "BrowseResultMask", "NodeClassMask": "NodeClass", "AccessLevel": "VariableAccessLevel", "UserAccessLevel": "VariableAccessLevel", "NotificationData": "NotificationData"}
-OverrideNames = {}  # {"RequestHeader": "Header", "ResponseHeader": "Header", "StatusCode": "Status", "NodesToRead": "AttributesToRead"} # "MonitoringMode": "Mode",, "NotificationMessage": "Notification", "NodeIdType": "Type"}
+OverrideTypes = {}  # AttributeId": "AttributeID",  "ResultMask": "BrowseResultMask", "NodeClassMask": "NodeClass",
+# "AccessLevel": "VariableAccessLevel", "UserAccessLevel": "VariableAccessLevel",
+# "NotificationData": "NotificationData"}
+OverrideNames = {}  # {"RequestHeader": "Header", "ResponseHeader": "Header", "StatusCode": "Status",
```
oroulet, line 32: shit that is old code I probably made in the first ever version of python-opcua client..

swamper123: Good old times :^)
I just trimmed it to 120 chars max (since my IDE is drowning me in such warnings). 
No clue what it is and what it does, since I didn't wanted to touch it for that PR.

swamper123: @oroulet Any changes wanted? 
If no, I would like to take a look on these generated files to modernize/optimize them a bit (in another PR).

Issue #336, opened 2020-11-18
Missing standard_address_spaces in fill_address_space


swamper123:
There exist more standard_address_space_parts than used in standard_address_space.py.

Some newer parts seemes to be not imported in [standard_address_space.py](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/server/standard_address_space/standard_address_space.py) while running server-minimal.py.

Part12 is containing remaining Nodes if I run server-mininal.py, but perhaps it's just me?

swamper123:
Since the part_X.xml is obsolete and the updater fixed, this case is closed :detective:

PR #436, opened 2021-02-05
Add missing part 12 to be created

swamper123:
It seems that part 12 was missing in the standard_addresspace.
Fix for #336


oroulet:
I think there was onlu 11 parts when I wrote that stuff... Also it takes some time to load all these things. but maybe this is onie the things that @AndreasHeine will fix with his profile support

swamper123:
The problem seems to be a missing Node "ServerConfiguration", ns=0;i=12637, from Opc.Ua.NodeSet2.Services.xml.
Tbh I don't know if they just put all standard Nodes in that file (61,5k lines of xml) or if there are many Nodes similar and some "nice to have" Nodes as well...

PR #443, opened 2021-02-11
Modernize generate files

swamper123:
Pep8 conformity
Remove obsolete generate script (the one with .bsd files)
Reduce write instructions by merging them together
Remove unused vars
Replaced prints with logs


schemas/generate_model.py
```diff
@@ -24,31 +20,18 @@
...
-
-# "NodesToRead": "AttributesToRead"} # "MonitoringMode": "Mode",
-# "NotificationMessage": "Notification", "NodeIdType": "Type"}
-
-
-# some object are defined in extensionobjects in spec but seems not to be in reality
-# in addition to this list all request and response and descriptions will not inherit
-# NoInherit = ["RequestHeader", "ResponseHeader", "ChannelSecurityToken", "UserTokenPolicy", "SignatureData",
-# "BrowseResult", "ReadValueId", "WriteValue", "BrowsePath", "BrowsePathTarget", "RelativePath",
-# "RelativePathElement", "BrowsePathResult"]#, "ApplicationDescription", "EndpointDescription"
```
oroulet, line 43: I did not notice these changes, but they are highly suspicious. They have a huge impact on the low level code

schemas/generate_model.py
```diff
@@ -24,31 +20,18 @@
...
+OverrideTypes = {}
 
 
 class Bit(object):
```
oroulet, line 29: if you want to clena up things. These should be dataclasses

schemas/generate_model.py
```diff
@@ -339,32 +300,17 @@ def parse(self):
...
                 self.model.enum_list.append(enum.name)
-            # else:
-            # print("Not implemented node type: " + tag + "\n")
+            else:
+                _logger.debug("Not implemented node type: " + tag + "\n")
         return self.model
 
     def add_extension_object(self):
         obj = Struct()
-        obj.name = 'ExtensionObject'
```
oroulet, line 348: the deleted extensionobject is probably the one giving me issues here...

schemas/generate_model.py
swamper123: We use out commented code on lowlevel?

schemas/generate_model.py
swamper123: I agree that this shouldn't be. Just the fields should be compromised.

PR #445, opened 2021-02-12
Fix missing references while import xml files

swamper123:
Add method to add reverse references, if missing
fix for #307

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -127,30 +127,21 @@ class XmlImporter:
         return nodes
 
     async def _add_missing_reverse_references(self, new_nodes):
-        # References which shall be uni-directional only:
-        # GuardvariableType NodeId(15113)
-        # Has Guard Reference Type NodeId(15112)
-        # TransitionVariableType NodeId(2762)
-        # StatemachineType NodeId(2771)
-        # StateVariableType  NodeId(2755)
-        # Two-stateVariableType NodeId(8995)
-        # StateType NodeId(2307)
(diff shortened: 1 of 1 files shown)


asyncua/common/xmlimporter.py
```diff
@@ -119,13 +119,45 @@ async def import_xml(self, xmlpath=None, xmlstring=None):
...
+        # TransitionVariableType NodeId(2762)
+        # StatemachineType NodeId(2771)
+        # StateVariableType  NodeId(2755)
+        # Two-stateVariableType NodeId(8995)
+        # StateType NodeId(2307)
+        # TransitionType NodeId(2310)
+        # FiniteTransitionVariableType NodeId(2767)
+        __unidirectional_nodes = {ua.NodeId(15113, 0), ua.NodeId(15112, 0), ua.NodeId(2762, 0), ua.NodeId(2771, 0),
+                                  ua.NodeId(2755, 0), ua.NodeId(8995, 0), ua.NodeId(2307, 0),
+                                  ua.NodeId(2310, 0), ua.NodeId(2767, 0), ua.NodeId(17603, 0)}
```
oroulet, line 142: should be possible to use ua.ObjectIds.XXXX instead of node numbers. This is hard to read and will break over time

oroulet: Otherwise the PR looks good

swamper123: I will change that now. Just a moment

oroulet: also no need to precise namepace=0 this is defualt

PR #476, opened 2021-03-10
fixes structs and xml

oroulet:


swamper123:
Since the importer shouldn't cause any trouble (so far), we could use a working xml file, import it, export it and compare them with each other. It should be the same (except the generated date I guess).

swamper123:
Perhaps we could use a schema from the UA-Nodeset for that. They have a mix of different kind of Nodes and especially we want to catch UaDataTypes atm.

oroulet:
I ran `xmllint --schema ../schemas/UA-Nodeset-master/Schema/UANodeSet.xsd  structs_and_enum.xml ` and fixed the only error it found (The order error you mentioned earlier)
yes you could try to importan export. but remember that the xmls from foundation are also generated from some software that is not bug free

PR #478, opened 2021-03-10
refactor recursive to iterative approach of awaiting cancelled tasks

Randelung:
It's 10 LoC...
I reforked the repository and recreated the changes.

How are you meant to synchronize a forked project with the original? Add an origin and MERGE the changes? That would put me one commit ahead of the original project...
Also, why did you need a rebase? Wouldn't the merge of master into the branch have been just as effective?

### DIFF
--- a/asyncua/common/xmlexporter.py
+++ b/asyncua/common/xmlexporter.py
@@ -6,9 +6,9 @@ import logging
 import functools
 from collections import OrderedDict
 import xml.etree.ElementTree as Et
-from copy import copy
 import base64
 from dataclasses import fields, is_dataclass
+from enum import Enum
 
 from asyncua import ua
 from asyncua.ua.uatypes import type_string_from_type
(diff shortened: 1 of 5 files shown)


oroulet:
@Randelung I am not sure I understand your question. Basically you have two ways to make branches. Either merge master and create merge commit everytime. I do not anyon doing that but it seems to be the default or rebase your branches on master `git pull --rebase` which in my opinion is much cleaner

oroulet:
@Randelung OK I had a look. I really think we should split that stuf in two methods something like:

```
def close_tasks_loop(self):
     while True:
              self._close_tasks()
              awaut asyncio.sleep(whaterver)

def _close_tasks(self):
      all the current code without the "while True"
      
      ```
      
      Does that make sense to you?

Randelung:
Goddammit.

PR #484, opened 2021-03-16
refactor recursive to iterative approach of awaiting cancelled tasks

Randelung:


asyncua/server/binary_server_asyncio.py
```diff
@@ -149,22 +149,23 @@ async def stop(self):
...
-                logger.exception("Unexpected crash in BinaryServer._await_closing_tasks")
-        if recursive:
+    async def close_task_loop(self):
+        while True:
+            await self._close_tasks()
             await asyncio.sleep(10)
-            await self._await_closing_tasks()
+
+    async def _close_tasks(self):
+        task = self.closing_tasks.pop()
```
oroulet, line 164: you are only closing one task here. it is missing the `while self.closing_tasks`

asyncua/server/binary_server_asyncio.py
```diff
@@ -149,22 +149,23 @@ async def stop(self):
...
-            task = self.closing_tasks.pop()
-            try:
-                await task
-            except asyncio.CancelledError:
-                # this means a stop request has been sent, it should not be catched
-                raise
-            except Exception:
-                logger.exception("Unexpected crash in BinaryServer._await_closing_tasks")
-        if recursive:
+    async def close_task_loop(self):
```
oroulet, line 158: since not used outside of that class it can be "private": start with underscore

oroulet:
OK. looks good now I think. does it solves your issue?

Randelung:
Yeah, I didn't observe any more memory increase back when the issue first arose.
The one task and function frame increase every 10s was a dead giveaway.

PR #502, opened 2021-03-25
Fix for importing nodesets with different order

swamper123:
Made some changes to suppress migration, via optional parameter.

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -31,8 +31,7 @@ class XmlImporter:
         xml_uris = self.parser.get_used_namespaces()
         server_uris = await self.server.get_namespace_array()
         namespaces_map = {}
-        xml_uris_enum = enumerate(xml_uris)
-        for ns_index, ns_uri in xml_uris_enum:
+        for ns_index, ns_uri in enumerate(xml_uris):
             ns_index += 1  # since namespaces start at 1 in xml files
             if ns_uri in server_uris:
                 namespaces_map[ns_index] = server_uris.index(ns_uri)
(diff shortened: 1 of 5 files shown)


asyncua/common/xmlimporter.py
```diff
@@ -31,7 +31,8 @@ async def _map_namespaces(self):
         xml_uris = self.parser.get_used_namespaces()
         server_uris = await self.server.get_namespace_array()
         namespaces_map = {}
-        for ns_index, ns_uri in enumerate(xml_uris):
+        xml_uris_enum = enumerate(xml_uris)
+        for ns_index, ns_uri in xml_uris_enum:
```
oroulet, line 35: that was a very strange change. I never saw an enumeration on two lines ;-)

asyncua/common/xmlimporter.py
```diff
@@ -105,7 +106,7 @@ async def import_xml(self, xmlpath=None, xmlstring=None):
         nodes = []
         for nodedata in nodes_parsed:  # self.parser:
             try:
-                node = await self._add_node_data(nodedata)
+                node = await self._add_node_data(nodedata, suppress_migration=True)
```
oroulet, line 109: can'tyou call that argument: no_ns_migration or even no_namespace_migration that is easier to understand

asyncua/common/xmlimporter.py
```diff
@@ -262,6 +268,25 @@ def _get_add_node_item(self, obj):
             node.TypeDefinition = self._migrate_ns(obj.typedef)
         return node
 
+    def _get_add_node_item_suppressed(self, obj):
```
oroulet, line 271: the work "suppressed" is also a bit stranged here. Not sure what you mean

oroulet:
also it would be nice if you had a minimal example that has that issue, so you could add a test. I am really not sure what you are trying to fix

asyncua/common/xmlimporter.py
swamper123: It should do the same as _get_add_note_item(), just without the migration, they should just take the obj.nodeid etc. But I will rework that.

swamper123:
> so you could add a test

There is already an existing test for `test_xml_import_additional_ns` which is breaking...but I don't know why atm.

PR #508, opened 2021-03-25
load custom types one by one

oroulet:


asyncua/common/structures.py
```diff
@@ -89,6 +89,7 @@ class {self.name}:
             if uatype == 'List[ua.Char]':
                 uatype = 'String'
             code += f"    {sfield.name}:{uatype} = {sfield.value}\n"
+        print(code)
```
swamper123, line 92: Is that print neccesarry`?

oroulet: it is debug code. That commit is just for debugging a bug report from today

PR #510, opened 2021-03-26
remove use of loop paramter in Client as of current asynio best pract…

oroulet:
…ices

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -29,7 +29,7 @@ class Client:
     use UaClient object, available as self.uaclient
     which offers the raw OPC-UA services interface.
     """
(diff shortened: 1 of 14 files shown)


oroulet:
@ZuZuD that changes breaks the Uaclient . I was unsure how you want to change it. Can you have a look?

oroulet:
@ZuZuD I tried fixing the ha:client... tests pass but they are many asyncio warning... maybe they were there anyway...

ZuZuD:
@oroulet I will try to have a look by this evening.

oroulet:
tests are passing so I am merging, but nice if you have a look

PR #515, opened 2021-03-26
Adding more sync functions

roin93:
Added the function I needed in the SyncWrapper.
The export_xml_by_ns() function is currently not working for me, but I am not sure if this is caused by the SyncWrapper or some other issue.

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -308,10 +308,6 @@ class Server:
         pass
 
     @syncmethod
-    def export_xml_by_ns(self, path=None, namespaces=None):
-        pass
-
-    @syncmethod
     def get_namespace_index(self, url):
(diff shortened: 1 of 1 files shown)


oroulet:
Maybe i removed that export_by_ns? I kind of remember there was an issue with it 🤔

roin93:
Feel free to remove it.

oroulet:
I checked and it that method is away. Can you remove that line? it is much easier from you since you created the merge request

swamper123:
@oroulet Do you mean that one?

https://github.com/FreeOpcUa/opcua-asyncio/blob/0a1cce0783d93493515ff92e55952e63909fc834/asyncua/server/server.py#L571

Issue #506, opened 2021-03-25
How to get ua_class of DataType of specific NameSpace


jcbastosportela:
Hi all. First of all, great work keeping this up to date with that standard!

I am using asyncua (0.9.0) for a tester client for a PLC that implements an OPC UA server. I am developing my own inf model.

The PLC creates its own OPC UA DataTypes on its own Namespace. I have my own inf model also loaded into that PLC, so it also has my defined DataTypes.

I have a method that takes one argument of a custom structured DataType, lets say `LinearOutputDeviceConfigDataType`.
In order to handle the method on the PLC I have to define a plc data type binary matching the OPC UA definition. The name that I give to the plc data type becomes a OPC UA DataType on the PLC's NameSpace. In my case I was giving the same name as on the Information Model I created. When from asyncua client I call the method giving as argument an object of the type get with `get_ua_class('LinearOutputDeviceConfigDataType')` apparently I am getting an instance of the DataType from the PLC's NameSpace, and it causes the PLC to reject the call.

When I rename the PLC's data type to `LinearOutputDeviceConfigDT` this works.

Consider the image bellow:
![image](https://user-images.githubusercontent.com/15647115/112498447-d397b100-8d86-11eb-9ddc-116f1eee0d34.png)

So, my question really is: is there a way of getting the ua_class of the specific Namespace? Or is there other way of getting the class of the input argument that makes sure that it is of the correct namespace?

Thank you so much for your help.

Regards,
Portela

oroulet:
ahah I knew that one would pop up one day. you cannot...

oroulet:
but I just made a PR: https://github.com/FreeOpcUa/opcua-asyncio/pull/508

jcbastosportela:
> but I just made a PR: #508

Very good man! Very fast.
I will try it as soon as possible.

jcbastosportela:
I was using 0.9.0, and when I checkout this commit many new things are coming I believe. With this version I get an exception on `load_data_type_definitions()`. Opened a new issue for this https://github.com/FreeOpcUa/opcua-asyncio/issues/509#issue-841309238

jcbastosportela:
So, yesterday after finding the issue I have tried some dirty hacks to get to try this functionality. Basically I have made similar changes (commit 4fb0d26b265b55adaebcc340f9ba366ed09640cc), catching same exceptions, but passing instead of raising, only to get through that failing DataType, which in my case seems to be the only one causing issues.
[...]

PR #517, opened 2021-03-28
Call get_python_classes while there are codes failing

jcbastosportela:
Call get_python_classes while there are codes failing (most likely due to dependency order)

### DIFF
--- a/asyncua/common/structures.py
+++ b/asyncua/common/structures.py
@@ -220,32 +220,46 @@ async def load_type_definitions(server, nodes=None):
         generators.append(generator)
         generator.make_model_from_string(xml)
         # generate and execute new code on the fly
-        generator.get_python_classes(structs_dict)
-        # same but using a file that is imported. This can be usefull for debugging library
-        # name = node.read_browse_name().Name
-        # Make sure structure names do not contain charaters that cannot be used in Python class file names
-        # name = clean_name(name)
-        # name = "structures_" + node.read_browse_name().Name
-        # generator.save_and_import(name + ".py", append_to=structs_dict)
+        
+        have_missing_data_types = True
+        prev_num_failed_codes = -1
+        while have_missing_data_types:
+            structs_dict, failed_codes = generator.get_python_classes(structs_dict)
 
-        # register classes
-        # every children of our node should represent a class
-        for ndesc in await node.get_children_descriptions():
-            ndesc_node = server.get_node(ndesc.NodeId)
-            ref_desc_list = await ndesc_node.get_references(refs=ua.ObjectIds.HasDescription, direction=ua.BrowseDirection.Inverse)
-            if ref_desc_list:  # some server put extra things here
-                name = clean_name(ndesc.BrowseName.Name)
-                if name not in structs_dict:
-                    _logger.warning("%s is found as child of binary definition node but is not found in xml", name)
-                    continue
-                nodeid = ref_desc_list[0].NodeId
-                ua.register_extension_object(name, nodeid, structs_dict[name])
-                # save the typeid if user want to create static file for type definitnion
-                generator.set_typeid(name, nodeid.to_string())
+            if len(failed_codes) > 0:
+                if prev_num_failed_codes > 0 and len(failed_codes) >= prev_num_failed_codes:
+                    raise Exception("The previous number of missing data types is the same after retry.")
+            else:
+                # all parsed
+                have_missing_data_types = False
+            
+            prev_num_failed_codes = len(failed_codes)
 
-        for key, val in structs_dict.items():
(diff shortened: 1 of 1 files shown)


asyncua/common/structures.py
```diff
@@ -220,32 +220,52 @@ async def load_type_definitions(server, nodes=None):
...
+        
+        have_missing_data_types = True
+        prev_num_failed_codes = -1
+        while have_missing_data_types:
+            # currently the 'get_python_classes' makes a single run on the data types from the information
+            # model being loaded; if a certain DataType depends on a DataType that is defined 'later' on
+            # the information model(s) it will fail because it makes use of an yet undefined Python class.
+            # So for solving this without changing deeply the current implementation of 'get_python_classes'
+            # we get get how many DataTypes failed to be converted to Python classes and repeat until the failed
+            # conversions is zero or not lower than the previous iteration.
```
oroulet, line 232: OK. that might make things more robust.
but can you move code under the `while'`in a method so the code is easier to read . So we clearly split between your retry  mechanism and the structure code

jcbastosportela: I can do so, but for that the section `# register classes`  will also need to move to that function. I don't know if that is desired.

jcbastosportela: I am having a look, but seems difficult to decouple the __retry__ from the __registration of structures__ because when there are "failed codes", before the retry it is necessary to register the successfully executed structures.

I am going to push an attempt, but again __retry__ from the __registration of structures__ are inside the coupled. :(

asyncua/common/structures.py
```diff
@@ -196,31 +196,36 @@ def set_typeid(self, name, typeid):
...
-    and include them in you code
+async def register_node_children_as_python_classes(server, node, generator):
+    """ Registers in 'ua' all children of a Node
+    server: Server
+        OPC UA server
+    node: Node
+        The node to be processed
+    generator: StructGenerator
+        Struct generator from the xml description of the node
+    Return dict of the registered structures
```
oroulet, line 207: The doc string is rather wrong now. We are are reading the value (an xml string) of a node. byt default "0:OPC Binary"

oroulet:
please stop adding all these merge commit. It is much cleaner to rebase your changes on top of master `git pull -r origin master`for example

jcbastosportela:
Are there any ambitions of getting this fix merged?

It aims to fix the issue I mention in https://github.com/FreeOpcUa/opcua-asyncio/issues/506#issuecomment-808912938 under 'EDIT'

jcbastosportela:
Is there any ambition to get this PR merged? As far as I can see the current master still have the same issue while loading data type definitions.

oroulet:
Also generally, this send to be a quote ugly hack. Exceptions should only be raised and catched in case of errors. Here you are using exceptions to avoid sorting the XML objects

jcbastosportela:
> Also generally, this send to be a quote ugly hack. Exceptions should only be raised and catched in case of errors. Here you are using exceptions to avoid sorting the XML objects

Well, I agree with you. But I feel that sorting the DataType nodes so all dependencies are resolved may be a much bigger change. It is necessary to consider that a Custom Structured DataType may have members that are other DataTypes defined on that namespace; so this would always require to make some parsing to the XML objects of the DataTypes looking to its members types and determining dependencies and check if all of them are already met, and based on that calculate its position on a list.

Obviously it is a solvable problem, I can try to approach the problem in that way. @oroulet do you suggest to reject this approach (even as an intermediary fix) and wait for a solution where the XML objects are sorted?
I personally would like to have the feature working ASAP and tackle "a nicer implementation" afterwards, but I totally understand if you don't want this code on the project.

oroulet:
We can leave that PR open, but I would rather not merge that. It make understanding code even more complicated. 
Also I am surprised I never got that issue. I suppose the xml usually come ordered. you must have encountered some special case to have that issue, maybe even a buggy xml

oroulet:
also this is for adding support for an older versin of protocl, newer versions do not use these xml nodes, they use the DataTypeDefiniont attributes

jcbastosportela:
> We can leave that PR open, but I would rather not merge that. It make understanding code even more complicated.
> Also I am surprised I never got that issue. I suppose the xml usually come ordered. you must have encountered some special case to have that issue, maybe even a buggy xml

I use UAModeler to create the models I use. I think it really is a matter of alphabetic order. When I look on the generated .xml file from UAModeler the DataTypes are somewhat sorted alphabetically.

> also this is fixingan older version of protocl, newer versions do not use these xml nodes, they use the DataTypeDefiniont attributes

Yeah, fair enough. I think the `load_data_type_definitions` is currently working correctly.

Issue #451, opened 2021-02-16
Type checking missing - invalid Values in Nodes possible


swamper123:
**Describe the bug** <br />
If a Node can be written on server side, a Client is able to write an invalid Value (like 123 for a boolean Node). A Type checking is not done.

**To Reproduce**<br />
See Code in #450 

**Expected behavior**<br />
Client should receive a Bad_TypeMismatch and server should raise Bad_TypeMismatch.

**Screenshots**<br />
See #450 

**Version**<br />
opcua-asyncio Version (e.g. master branch, 0.9): All

PR #519, opened 2021-03-29
Add typecheck for variant type

swamper123:
Fix for #451 and #450 discussion.

Tested with our own Client, since UaExpert won't allow such a request for normal Nodes.

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -706,7 +706,7 @@ class AddressSpace:
         if attval is None:
             return ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid)
 
-        if value.Value.VariantType is not attval.value.Value.VariantType and (nodeid.NamespaceIndex == 0 and nodeid.Identifier not in (
+        if (value.Value.VariantType != attval.value.Value.VariantType) and not (nodeid.NamespaceIndex == 0 and nodeid.Identifier in (
                     ua.ObjectIds.Server_ServerCapabilities_OperationLimits_MaxNodesPerRead,
                     ua.ObjectIds.Server_ServerCapabilities_OperationLimits_MaxNodesPerHistoryReadData,
                     ua.ObjectIds.Server_ServerCapabilities_OperationLimits_MaxNodesPerHistoryReadEvents,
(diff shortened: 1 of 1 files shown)


swamper123:
It seems that some tests probem appear with that PR.
Either it's my fault and it breaks, or the tests are not well enough (and I will fix them).

swamper123:
Just a question:
```python
async def test_service_fault(server, admin_client):
    request = ua.ReadRequest()
    request.TypeId = ua.FourByteNodeId(999)  # bad type!
    with pytest.raises(ua.UaStatusCodeError):
        await admin_client.uaclient.protocol.send_request(request)
```
This test fails, because UaStatusCodeError BadTypeMismatch raised....but it expects any UaStatusCode Error, does it?

swamper123:
It seems that:

https://github.com/FreeOpcUa/opcua-asyncio/blob/80a4cc2635e8888ddda051f6dc282e2873570d0e/tests/util_enum_struct.py#L29

is making trouble.
Inside of Node 6009, the value attribute is empty. Type is VariantType.Null, Value is None etc., but the attribute DataType seems to be right.

grafik

I will not change write_attribute_value, since this could be used for any attribut. Not sure where stuff is causing trouble here atm.

oroulet:
Sorry I was very late looking at that PR. 
I think the principle is OK, if there if VariantType is not the expected one we should raise type mismatch (apart from Vraianttype.Null that can replace most of the other types).
It is going to break some things, but we probably need to fix these things ,also all people writting Int64 where UInt32 is expected will break too..

But why do you exclude all these standard nodes?

asyncua/server/address_space.py
```diff
@@ -706,6 +706,41 @@ async def write_attribute_value(self, nodeid, attr, value):
         if attval is None:
             return ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid)
 
+        if value.Value.VariantType != attval.value.Value.VariantType:
+            if not (nodeid.NamespaceIndex == 0 and nodeid.Identifier in (
```
oroulet, line 710: not (xxx and ddd in (ccc)) is far too complicated test. I am sure you can change the order to make it understadnable.

some like 

if xxxx:
   pass
do_somehing()

oroulet:
I made an alterntive where all tests pass just by correctly handling NULL type

oroulet:
https://github.com/FreeOpcUa/opcua-asyncio/pull/587

swamper123:
> Sorry I was very late looking at that PR.
No worries, haven't touched it in a while anyway. 😅 

> I think the principle is OK, if there if VariantType is not the expected one we should raise type mismatch (apart from Vraianttype.Null that can replace most of the other types).

A VariantType.Null that will be overwritten by something or something that will be overwritten by VariantType.Null?

> But why do you exclude all these standard nodes?

I am not sure anymore tbh...I think something broke in the tests with these notes when they are initialized. Changing the Type was not expected in this draft. I will study the specs again, what the excepted behaviour is.

Allowing exceptions for this case would may still lead to problems. Otherwise each half-broken client could change int32 to Null and the Problem would still exist.

oroulet:
>> Allowing exceptions for this case would may still lead to problems. Otherwise each half-broken client could change int32 to Null and the Problem would still exist.

Valid point. Does spec says anything about that? maybe only admin should be allowed to write Null or maybe nobody...

I close that one. comment on the other PR

Issue #522, opened 2021-03-30
test_cli_tools has runtimeerror issue


swamper123:
**Describe the bug** <br />
test_cli_tools causes a Runtime error -> sometimes tests fail.

Log from a failed test:

```
 ________________________________ test_cli_tools ________________________________

running_server = 'opc.tcp://127.0.0.1:40399'

    async def test_cli_tools(running_server):
        # admin privileges are only needed for uawrite
        url = running_server.replace("//", "//admin@")
        default_opts = ["mock_func", "-u", f"{url}"]
        rw_node = ["-n", RW_NODE]
        call_node = ["-n", ROOT_NODE]
        write_val = ["val_to_write"]
        rw_opts = default_opts + rw_node
        write_opts = rw_opts + write_val
        call_opts = default_opts + call_node + write_val
    
        tool_opts = {}
        tool_opts[uaread] = rw_opts
        tool_opts[uals] = rw_opts
        tool_opts[uawrite] = write_opts
        tool_opts[uahistoryread] = rw_opts
        tool_opts[uaclient] = default_opts
        tool_opts[uadiscover] = default_opts
        tool_opts[uacall] = call_opts
    
        for tool in (uaread, uals, uawrite, uahistoryread, uaclient, uadiscover, uacall):
            # It's necessary to mock argv, else the tool is invoked with *pytest's* argv
            with patch.object(sys, 'argv', tool_opts[tool]):
                with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
                    result = executor.submit(tool)
                    for future in concurrent.futures.as_completed([result]):
                        exception = future.exception()
                        # py3.6 returns SystemExit(0,)
                        str_exp = repr(exception).replace(",", "")
>                       assert str_exp == "SystemExit(0)"
E                       assert 'ConnectionRe....1\' 40399)")' == 'SystemExit(0)'
E                         - SystemExit(0)
E                         + ConnectionRefusedError(111 "Connect call failed ('127.0.0.1' 40399)")

tests/test_tools.py:45: AssertionError
------------------------------ Captured log setup ------------------------------
INFO     root:internal_server.py:65 No user manager specified. Using default permissive manager instead.
INFO     asyncua.server.internal_session:internal_session.py:45 Created internal session Internal
[...]

PR #524, opened 2021-03-30
server read_attribute_value

AndreasHeine:
Unbenannt1

### DIFF
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -295,8 +295,6 @@ class InternalServer:
         """
         directly read datavalue of the Attribute
         """
-        # https://github.com/FreeOpcUa/opcua-asyncio/pull/524
-        await asyncio.sleep(0)  
         return self.aspace.read_attribute_value(nodeid, attr)  
 
     def set_user_manager(self, user_manager):
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -295,6 +295,7 @@ class InternalServer:
         """
         directly read datavalue of the Attribute
         """
+        # https://github.com/FreeOpcUa/opcua-asyncio/pull/524
         await asyncio.sleep(0)  
         return self.aspace.read_attribute_value(nodeid, attr)  
 
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -291,7 +291,7 @@ class InternalServer:
         """
         await self.aspace.write_attribute_value(nodeid, attr, datavalue)
 
-    async def read_attribute_value(self, nodeid, attr=ua.AttributeIds.Value):
+    def read_attribute_value(self, nodeid, attr=ua.AttributeIds.Value):
         """
         directly read datavalue of the Attribute
         """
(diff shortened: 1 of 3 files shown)


asyncua/server/internal_server.py
```diff
@@ -291,6 +291,13 @@ async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Va
         """
         await self.aspace.write_attribute_value(nodeid, attr, datavalue)
 
+    async def read_attribute_value(self, nodeid, attr=ua.AttributeIds.Value):
+        """
+        directly read datavalue of the Attribute
+        """
+        await asyncio.sleep(0)
```
AndreasHeine, line 298: any better ideas?

swamper123 (approved):
I am a bit confused, why write_attribute_value is a coroutine, but read_attribute_value not ... Anyway I like the idea and can't see a problem at the moment with that, since this is server sided.

oroulet:
Is that an optimized server side read? like the write we currently have?
How much faster is it? or at least what kind of time consuming operation does it save?:
Also it should be async if it must be async (has some await somewhere inside) otherwise sync is fine

asyncua/server/internal_server.py
oroulet: what is that line? It sould work withtout quite fine. Ans also that method does not need to be a coro

AndreasHeine:
it ment to be used as a serverside read without a nodeclassinstance so you save memmory and i like to use it for kind of gateway/aggregation pure serverside interaction in which storing just the nodeid is in my oppinion smarter then having thousands of nodeclass instances (with this change you can choose which way to go) 

i was kind of confused that addspace.read_attribute_value is not a coro... i would make it async and test if there are some sideeffects!

oroulet (changes requested):

oroulet:
5 commits and manye discussion gor 5 simple lines ;-) but now it looks OK t least ;-)
obs btw: you should add test and maybe a line in one example because otherwise nobody is never going to find out it exist and without tests it will probably get broken at some points

AndreasHeine:
the overall result counts ;-)
adding tests!

tests/test_server.py
```diff
@@ -653,6 +653,14 @@ async def check_custom_type(ntype, base_type, server: Server, node_class=None):
     assert await ntype.get_child("2:PropertyString") in properties
     assert (await(await ntype.get_child("2:PropertyString")).read_data_value()).Value.VariantType == ua.VariantType.String
 
+async def test_server_read_write_attribute(server: Server):
+    node = await server.get_objects_node().add_variable(0, "0:TestVar", 0, varianttype=ua.VariantType.Int64)
```
swamper123, line 657: Since our tests all run on one Testserver, all nodes should be removed at the end of the test. Otherwise later tests may fail, because somebody wanted to add a "TestVar" Node as well.
Just add  `await opc.opc.delete_nodes([node])` at the end of the test.

AndreasHeine: ok

AndreasHeine:
some how if i run test local they fail with file not found error (maybe something missing in my branch...):
test

but test_server.py is passing so it should be fine!

tests/test_server.py
oroulet: orgive them a name you know nobody else will use. 
Ideally these should be handled in pytest  fixtures... but it is not

swamper123:
This is caused by #522 , happend to me as well.

PR #525, opened 2021-04-01
Fix typo in comment

AndreasHeine:


oroulet:
That comment is old. Are Bytes defined at all in that file?

oroulet:
Ok it is defined...

PR #526, opened 2021-04-01
refactor

AndreasHeine:
-removed import
-used uatypes datetime

### DIFF
--- a/asyncua/common/ua_utils.py
+++ b/asyncua/common/ua_utils.py
@@ -3,7 +3,6 @@ Usefull method and classes not belonging anywhere and depending on asyncua libra
 """
 
 from dateutil import parser
-from datetime import datetime
 from enum import Enum, IntEnum
 import uuid
 import logging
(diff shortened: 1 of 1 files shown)


oroulet:
I am not really sure what is the best thing to do here. I wrote these classes mainly for typing. And made them inherit python types because it was easy. But that was not used anywhere. You are writing the first use here.

AndreasHeine:
currently messing around with the datavalue.ServerTimestamp = ua.DateTime.utcnow() is not set and causes "waiting for initial data" ...

AndreasHeine:
```python
def value_to_datavalue(val, varianttype=None):
    """
    convert anyting to a DataValue using varianttype
    """
    if isinstance(val, ua.DataValue):
        datavalue = val
    elif isinstance(val, ua.Variant):
        datavalue = ua.DataValue(val)
        datavalue.SourceTimestamp = ua.DateTime.utcnow()
        # datavalue.ServerTimestamp = ua.DateTime.utcnow() # <--- causes "waiting for initial data" 
    else:
        datavalue = ua.DataValue(ua.Variant(val, varianttype))
        datavalue.SourceTimestamp = ua.DateTime.utcnow()
        # datavalue.ServerTimestamp = ua.DateTime.utcnow() # <--- causes "waiting for initial data" 
    return datavalue
```

swamper123:
Something causes trouble. The ServerTime value itself looks like (from 11712 - "MaxNodesPerTranslateBrowsePathsToNodeIds"), an OperationalLimit Node):
grafik

But via UaExpert these are not browsable:
grafik

Could there be an error of how ServerTimestamp is structured?

PR #533, opened 2021-04-06
check for tasks before cancelling them

### LINKED ISSUE #337: cannot stop client if no connection established
Steps to reproduce:
1. Ensure that the client is *not* able to connect to the given opc-ua server URI
2. `client.connect()`
3. `client.disconnect()`

The client will raise an `AttributeError`.
If one catches this, the asyncua thread hangs.

Randelung:
Fixes #337.
Have a look at the changes and see if you want the `None` tasks to not trigger a log warning or not.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -508,11 +508,14 @@ class Client:
         """
         Close session
         """
-        self._renew_channel_task.cancel()
-        try:
-            await self._renew_channel_task
-        except Exception:
-            _logger.exception("Error while closing secure channel loop")
+        if self._renew_channel_task is not None:
+            self._renew_channel_task.cancel()
+            try:
+                await self._renew_channel_task
+            except asyncio.CancelledError:
+                pass
+            except Exception:
+                _logger.exception("Error while closing secure channel loop")
         return await self.uaclient.close_session(True)
 
     def get_root_node(self):
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -266,7 +266,7 @@ class UaClient:
         await asyncio.wait_for(asyncio.get_running_loop().create_connection(self._make_protocol, host, port), self._timeout)
 
     def disconnect_socket(self):
-        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
+        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED or self.protocol is None:
             self.logger.warning("disconnect_socket was called but connection is closed")
             return None
         return self.protocol.disconnect_socket()
@@ -282,7 +282,7 @@ class UaClient:
         close secure channel. It seems to trigger a shutdown of socket
         in most servers, so be prepare to reconnect
         """
-        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
+        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED or self.protocol is None:
             self.logger.warning("close_secure_channel was called but connection is closed")
             return
         return await self.protocol.close_secure_channel()
@@ -311,9 +311,15 @@ class UaClient:
 
     async def close_session(self, delete_subscriptions):
         self.logger.info("close_session")
-        self.protocol.closed = True
-        if self._publish_task and not self._publish_task.done():
+        if self._publish_task and not self._publish_task.done() or self._publish_task is not None:
             self._publish_task.cancel()
+            try:
+                await self._publish_task
+            except asyncio.CancelledError:
+                pass
+        if self.protocol is None:
+            return
+        self.protocol.closed = True
         if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
             self.logger.warning("close_session was called but connection is closed")
             return


asyncua/client/ua_client.py
```diff
@@ -266,7 +266,7 @@ async def connect_socket(self, host: str, port: int):
         await asyncio.wait_for(asyncio.get_running_loop().create_connection(self._make_protocol, host, port), self._timeout)
 
     def disconnect_socket(self):
-        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
+        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED or self.protocol is None:
```
ZuZuD, line 269: I faced the same issue with the [HaClient](https://github.com/FreeOpcUa/opcua-asyncio/blob/750ef8956f0845ecb6c6a1817fa6128df10cbcfb/asyncua/client/ha/ha_client.py#L544-L546). I figured fixing this issue at such a low level could lead to nasty race conditions:

1- I send a connection request which hasn't completed yet (server loaded or high RTT)
2- In the meantime I send a disconnection which fast return since there's no protocol yet.
3 - Connection succeeds but my program called disconnect and hence assumes there's no connection => connection leak

The low level client calls must guarantee the responses are consistent and reliable, that's why I still believe it's safer to manage this at a higher level.

Randelung: I see your point.
If the low level calls must guarantee consistency, then the low level client would have to track open connection requests and kill them if necessary. Moving the handling up into high level code would achieve the opposite of what you described, everything below that high level would be possibly inconsistent (as is the case now).

I see two alternatives.
- Receive the (transport, protocol) tuple from the create_connection method instead of separating them, thereby granting the possibility to cancel the connection establishment altogether. Put it in a separate task (the wait_for is already 90% there, anyway) and keep a reference to it. Disconnect() just kills that task, too, in addition to what it is doing now. Both transport and protocol remain unassigned.
- Keep a flag synchronous with the connection request and kill the connection once established. This approach sucks because if you cancel and then re-establish the connection you can't tell the difference between the first and second request, which means you'd have to block connection requests until the first one has been established and destroyed, before the next one is allowed to happen.

I'll see what I can work out towards option 1.

asyncua/client/ua_client.py
```diff
@@ -282,7 +282,7 @@ async def close_secure_channel(self):
         close secure channel. It seems to trigger a shutdown of socket
         in most servers, so be prepare to reconnect
         """
-        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
+        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED or self.protocol is None:
```
oroulet, line 285: Can't you write: 
`if self.protocol is None or self.protocol.state == UASocketProtocol.CLOSED:` ?

asyncua/client/ua_client.py
```diff
@@ -311,9 +311,15 @@ async def activate_session(self, parameters):
 
     async def close_session(self, delete_subscriptions):
         self.logger.info("close_session")
-        self.protocol.closed = True
-        if self._publish_task and not self._publish_task.done():
+        if self._publish_task and not self._publish_task.done() or self._publish_task is not None:
```
oroulet, line 314: same here. That line is too complex. I suspect it can be written in a more error proff way.

asyncua/client/ua_client.py
```diff
@@ -311,9 +311,15 @@ async def activate_session(self, parameters):
...
             self._publish_task.cancel()
+            try:
+                await self._publish_task
+            except asyncio.CancelledError:
+                pass
+        if self.protocol is None:
+            return
+        self.protocol.closed = True
         if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
```
oroulet, line 323: you tested 2 lines over that self.protocol is not None

asyncua/client/ua_client.py
ZuZuD: Calling disconnect when the connection is not established and raising is consistent at such level. Low level calls are not meant to be smart but functional, as upper abstraction layers can manage these errors which also has the benefit to create a clear separation of concerns.

Thus said you should be able to move all of this in a task and cancel it, but such error management is too complicated to live here imho.

asyncua/client/ua_client.py
Randelung: Yep. I was kind of anticipating to make the whole check for None separate to avoid the warning, but if you want to keep it as is I'll merge the expressions.

asyncua/client/ua_client.py
Randelung: Whoops, I can just revert this. The added test is entirely redundant.

asyncua/client/ua_client.py
Randelung: I was checking for the task to be None, not the protocol. If there's no protocol, the rest of the method is unnecessary, thus the extra return here.

asyncua/client/ua_client.py
Randelung: I've put the extra task into client.connect and cancelled it in client.disconnect, as you suggested.

asyncua/client/ua_client.py
Randelung: I've moved it around a little, maybe that took care of what you meant.

asyncua/client/ua_client.py
```diff
@@ -266,10 +268,13 @@ async def connect_socket(self, host: str, port: int):
...
-        if self.protocol and self.protocol.state == UASocketProtocol.CLOSED:
+        if self.protocol is None or self.protocol.state == UASocketProtocol.CLOSED:
             self.logger.warning("disconnect_socket was called but connection is closed")
-            return None
-        return self.protocol.disconnect_socket()
+            return
+        self.protocol.disconnect_socket()
+        # removing the protocol will result in a fail in the ha_test. IMO returning to the initial state of None
+        # and allowing garbage collection would be a good idea.
+        # self.protocol = None
```
ZuZuD, line 277: The tests indeed rely on `UASocketProtocol` to check the client has failed-over. 
By changing [wait_clients_socket](https://github.com/FreeOpcUa/opcua-asyncio/blob/750ef8956f0845ecb6c6a1817fa6128df10cbcfb/tests/conftest.py#L372-L378) to pass when `state == UASocketProtocol.CLOSED` you will get rid of this error.

ZuZuD (changes requested):

tests/conftest.py
```diff
@@ -372,10 +372,16 @@ async def ha_client(ha_config, ha_servers):
 async def wait_clients_socket(ha_client, state):
     for client in ha_client.get_clients():
         for _ in range(RETRY):
-            if client.uaclient.protocol and client.uaclient.protocol.state == state:
-                break
+            try:
+                result = state is None and client.uaclient.protocol is None or \
```
ZuZuD, line 376: Why not changing just this bit?
```
if (
    client.uaclient.protocol == state or 
    client.uaclient.protocol and client.uaclient.protocol.state == state
):
```
nit: `is` compares object and `==` compares values, better being explicit here.

tests/conftest.py
```diff
@@ -372,10 +372,16 @@ async def ha_client(ha_config, ha_servers):
...
+                        client.uaclient.protocol.state is state
+                if result:
+                    break
+            except AttributeError: # protocol doesn't exist yet (connection attempt ongoing), but target is not None
+                pass
             await sleep(SLEEP)
-        assert client.uaclient.protocol.state == state
+        assert result
+
```
ZuZuD, line 384: Why not changing just this bit?
```
if (
    client.uaclient.protocol == state or 
    client.uaclient.protocol and client.uaclient.protocol.state == state
):
    [...]
assert client.uaclient.protocol == state or client.uaclient.protocol.state == state
```
nit: `is` compares object and `==` compares values, better being explicit here.

asyncua/client/client.py
```diff
@@ -253,6 +263,14 @@ async def disconnect(self):
...
         """
         _logger.info("disconnect")
+        if self._connect_task and not self._connect_task.done():
+            self._connect_task.cancel()
+            try:
+                await self._connect_task
+            except asyncio.CancelledError:
+                pass
+            finally:
+                self._connect_task = None
```
ZuZuD, line 273: If we end up here, we have a `connect` coroutine running. What's the use-case where we need to set `self._connect_task = None` again?

tests/conftest.py
Randelung: The method gives the test several tries before asserting a False, so that if after 3 seconds the test comes out positive it'll return immediately instead of waiting the full timeout. In order to not evaluate the expression twice (also DRY) I used a storage variable.
The reason for `is` instead of `==` is that the arguments in the function are enums, so either UaProtocol.INITIALIZED/OPEN/CLOSED or None. IMO any other value would be wrong and shouldn't be accepted.

asyncua/client/client.py
Randelung: If we end up here, a connect task is running and being awaited inside the try block just above. Another task doing something else might call disconnect and cancel the connect task, resulting in a CancelledError. In any case, the _connect_task is either cancelled or done after the await inside the try block, which is why resetting it to None signals that it's not running.
It's not explicitly necessary, but returning to initial state means you only have to worry about "there's a task and it's running or there's no task" and not check if the task is done or cancelled or whatever when disconnecting.

asyncua/client/client.py
```diff
@@ -229,8 +230,17 @@ async def connect_and_find_servers_on_network(self):
...
-
+    
     async def connect(self):
+        self._connect_task = asyncio.create_task(self._connect())
+        try:
+            await self._connect_task
+        except asyncio.CancelledError:
+            pass
+        finally:
+            self._connect_task = None
```
oroulet, line 241: why putting the connection in the background then waiting for it?? It does not make any sense

Randelung: So that disconnect can cancel the task. The connect behavior is supposed to be blocking (to preserve the behavior it has so far), but disconnect should be able to cancel the process. Therefore, a task is needed to be cancellable and awaiting it keeps the process blocking.

oroulet: in theory in a normal applicaiton you call `await connect()`then `await disconnect()` An application calling disconnect() while doing calling connect() is doing something quite unusual and should handle its own shit itself. They will anyway do a much better job than we do at low level

Randelung: ¯\\_(ツ)\_/¯ the use case was [given by ZuZuD](https://github.com/FreeOpcUa/opcua-asyncio/pull/533#discussion_r607767989). Without it, checking for an existing protocol before trying to disonnect it would be sufficient.

oroulet: I never looked at his code. But for sure he should have his own tasks and can cancel them. It does need even more tasks inside his tasks

Randelung: I don't follow. It was the first code review for this PR and the reason why a simple check for client.protocol was insufficient.
If you don't think it's necessary I'll revert it.

ZuZuD: > in theory in a normal applicaiton you call `await connect()`then `await disconnect()` An application calling disconnect() while doing calling connect() is doing something quite unusual and should handle its own shit itself. They will anyway do a much better job than we do at low level

I can't agree more with this and that's why we chose to deal with this at the HA level.

oroulet: @Randelung sorry I did not follow and do not have the capacity to folow everything. So I just check PR when I got 5 minutes and comment what I see. I do not know the history but clearly a wrapper making  task of a method and just awaiting it ooks very strnage and you must have a very clear use case for such a thing. It does not seem to fit the current API

Randelung: Yeah, no worries, I know how it is.
How do we continue? I think at this point it's more of a design choice.

PR #541, opened 2021-04-07
write_values & read_values

Ilyazyk:


asyncua/sync.py
```diff
@@ -224,6 +224,14 @@ def get_node(self, nodeid):
     def connect_and_get_server_endpoints(self):
         pass
 
+    @syncmethod
+    def write_values(self):
+        pass
+    
+    @syncmethod
+    def read_values(self):
```
oroulet, line 232: you need to copy the entire def line from Client, incl arguments

PR #544, opened 2021-04-08
add service result code to log

Randelung:
Having run into 
```
2021-04-08 14:55:44.939 WARNING  asyncua.client.ua_client.UASocketProtocol: ServiceFault from server received  in response to CreateMonitoredItemsRequest
```
and used Wireshark to check for details, I thought it might be useful to have the response code, too.
```
2021-04-08 14:55:44.939 WARNING  asyncua.client.ua_client.UASocketProtocol: ServiceFault (BadTooManyOperations) from server received  in response to CreateMonitoredItemsRequest
```

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -156,8 +156,8 @@ class UASocketProtocol(asyncio.Protocol):
         data = data.copy()
         typeid = nodeid_from_binary(data)
         if typeid == ua.FourByteNodeId(ua.ObjectIds.ServiceFault_Encoding_DefaultBinary):
-            self.logger.warning("ServiceFault from server received %s", context)
             hdr = struct_from_binary(ua.ResponseHeader, data)
+            self.logger.warning("ServiceFault (%s) from server received %s", hdr.ServiceResult.name, context)
             hdr.ServiceResult.check()
             return False
         return True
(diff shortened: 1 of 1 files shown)


asyncua/client/ua_client.py
```diff
@@ -156,8 +156,8 @@ def check_answer(self, data, context):
         data = data.copy()
         typeid = nodeid_from_binary(data)
         if typeid == ua.FourByteNodeId(ua.ObjectIds.ServiceFault_Encoding_DefaultBinary):
-            self.logger.warning("ServiceFault from server received %s", context)
             hdr = struct_from_binary(ua.ResponseHeader, data)
+            self.logger.warning("ServiceFault (%s) from server received %s", hdr.ServiceResult.name, context)
```
oroulet, line 160: what does that ServiceResult struct contains? maybe there are other interesting data in it?

Randelung: image
The rest is of limited use I think.

oroulet: in theory the ServiceDiagnositic can give some info, but I have never seen it being used by any server. We could print it anyway though...

Randelung: Just as a whole?
```python
self.logger.warning("ServiceFault (%s, diagnostics: %s) from server received %s", hdr.ServiceResult.name, hdr.ServiceDiagnostics, context)
```
Just looking at the class, it appears very versatile. I assume most fields are usually empty?

oroulet: yes. I have never seens any not empty field in it....
Looks OK

Randelung: :D
I'll add it in, then.

PR #550, opened 2021-04-14
[test_tools] Implement client retry to fix race

ZuZuD:
Fix for #522

Implement a client retry decorator for test_tools to workaround the race condition between the server start and first client connect. That's certainly the best approach since the server is running in a separate thread for this specific test and it's consequently difficult to get its status.

I was able to repro by inserting a sleep in [add_server_methods](https://github.com/FreeOpcUa/opcua-asyncio/blob/2dcba36ee25999411521feaf5355702beb790d7d/tests/test_common.py#L26) 

I've performed various tests around the retry function to make sure it only triggers with the provided exception. Let's see if 5s is enough for our validation pipeline.

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -10,6 +10,7 @@ import asyncio
 from datetime import datetime
 from datetime import timedelta
 import math
+import time
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


swamper123:
I will rerun the checks a couple of times, to be sure. But thanks for the fix. It was very confusing. ^^

oroulet:
should be possible to add a condition so we know when server is ready. Let me have a short look

swamper123:
Well...that would be the better option ^^

swamper123:
@oroulet Found anything? Or still taking a short look? ;)

oroulet:
I started but then got cought by something else....

ZuZuD:
Closing in favour of #552

PR #553, opened 2021-04-28
added missing requirements to allow easy setup for developers

ratara:
Open idea:
- I would suggest fixed versions for the requirements to maintain a stable version and to prevent unforseen breaking changes. But we have to maintain that versions and i am not sure, if there is somebody willing to do so.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -4,3 +4,9 @@ coverage
 pytest-cov
 pytest-mock
 asynctest
+python-dateutil
+pytz
+aiofiles
+cryptography
+sortedcontainers
+aiosqlite


ratara:
Well, I guess these unit test fails are meant, if we are speaking of unstable? - I obviously did not change any behaviour.

oroulet:
your branch is out of date . Rebase on master. I think the last commit os the one fixing that issue

dev_requirements.txt
```diff
@@ -4,3 +4,9 @@ coverage
 pytest-cov
 pytest-mock
 asynctest
+python-dateutil
+pytz
+aiofiles
+cryptography
+sortedcontainers
+aiosqlite
```
oroulet, line 12: here you are repeated dependencies which are already in requirement.py (thorugh the use of a dot, so the dependencies are fetched from setup.py)

oroulet: so not, we should not do that

Issue #322, opened 2020-10-28
Tests share same Server per testfile


swamper123:
Hello together,

I mentioned in #306 that I got some problems with the tests, which interfered eachother.

While testing the stuff through, I thought just changing the fixture to "function" would solve the problem.
But I suspect, that there is a memory leak happening which causes Travis freeze to death (10min. Timeout without response).

I would recommened that either each test cleans up behind itself and finishes clearly (not causing "Address already in use" and similar errors) or for every test a fresh server has to be buildUp and tearedDown afterwards somehow (which I would prefere and accept longer test runs).

swamper123:
*ping*
Any comments on this?
Shall we rewrite the existing tests?
Should we make better usecases where tests are allowed to influence each other in strict areas?
Does it make sense?
Pro/Cons?

ratara:
One of the Golden Blues of unit testing is that tests have to be independent of each other. Breaking this rule breaks the test in so many ways, it should always be a red line.

I highly recommend that the test are setup and teared down correctly including the server.

oroulet:
@ratara thank you for you impressive insight. Now i am waiting for you PR

oroulet:
@swamper123 yes many tests share the same server (not all) due to the performance reasons.
And yes some cleanup and more machinery to clean up after each tests would be nice. There are a bunch of functions already but i do not have overview anymore

ratara:
@oroulet I am no native speaker but i got it. :-)

Well i am not sure how much work is to do but i started a try. I got the code and now I am trying to understand whats going on.
1. It seems as there are missing dependencies like dateutil and more which are not mentioned in the dev_requirements.txt. Btw. there are no fixed versions of the dependencies, we should fix that to get no surprises.
2. I invested only 10 minutes to look but i did not found any howto setup the development system. This would help a lot.
3. I am new to pytest but after running a test discovery over 600+ files have changes. Is that some sort of bug or wrong usage by me? - I used VScode and the integrated commands for testing.
4. I feel like flooding the thread. Is there any place to talk about details without disturbing this thread?

oroulet:
1- this software is written by me and others when we have time, so yes many things are missing. just send PR for every things you find
2- no doc but you can look at setup.py file. It list all dependencies that are known
3. writting here is fine, or you can start a discussion: https://github.com/FreeOpcUa/opcua-asyncio/discussions

ratara:
[...]

PR #559, opened 2021-05-03
Draft: Improve history reading

HarrySky:
See #558

Should be backwards compatible. Changes:
- Make `ReturnBounds` parameter configurable
- Make possible to read history recursively with `recursive` argument

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -514,7 +514,7 @@ class Node:
             rpath.Elements.append(el)
         return rpath
 
-    async def read_raw_history(self, starttime=None, endtime=None, numvalues=0):
+    async def read_raw_history(self, starttime=None, endtime=None, numvalues=0, return_bounds=True, recursive=False):
         """
         Read raw history of a node
         result code from server is checked and an exception is raised in case of error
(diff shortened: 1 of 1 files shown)


HarrySky:
Instead of recursive argument - separate method could be created

oroulet:
Is there any reason not to have recursive True all the time? I mean this is a high level method, users expect it to return all requested data. I think all high level browse methods work that way. If several calls are needed they do it anyway

HarrySky:
I tried not to break existing codebases that don't expect recursive behaviour :smile:

oroulet:
I am rather sure, people using that method expected a recursive behavior. Just did not think that there might be cases where the server does not answer everything.
I would remove the recursive option and make it behave recursively in all cases

HarrySky:
Done, made it recursive by default without argument

HarrySky:
It seems that there are some problems with server code if recursive is the default. Client works from that commit as expected

HarrySky:
Just in case, I use https://github.com/open62541/open62541 as OPC UA server

oroulet:
What symptoms?

HarrySky:
I mean CI failed here on some server code if I understood logs correctly

HarrySky:
So, should I bring back `recursive=False`? Or can we fix server tests somehow?

oroulet:
your code looks correct, so we really should fix the server...

HarrySky:
After quick search I found a possible problem: #322 (not sure, though, but maybe history is piling up from other tests and when we read it recursively - it gives more records). I do not know much about asyncua server, to be honest

oroulet:
I had a look and I thinnk th eserver is not beahving correctly. It returns a ContinuationPoint when it should not. It always a return a ContinuationPoint if NumValuesPerNode is specified, it should probably not. It will need to be fixed.
https://github.com/FreeOpcUa/opcua-asyncio/blob/d2ba37e3dd0d9a8e054fac7f15be4b8dd34d1066/asyncua/server/history.py#L147

As far as I understand the conitnuation poitns has to be used only dues to internal server limitations, we do not have any so we should never return one OR maybe set a reasonable value there such as as 100000 for example

HarrySky:
Yeah, in my case server sends maximum N values to one history read request. So I need to use `ContinuationPoint` to read all the data available

oroulet:
do you se N yourself or is it something the server has decided? what is the value of N?

HarrySky:
Server sets it, N can be anything, for testing we currently use N=100

```cpp
/* We want the server to serve a maximum of 100 values per request. This
 * value depend on the plattform you are running the server. A big server
 * can serve more values, smaller ones less. */
setting.maxHistoryDataResponseSize = 100;
```

oroulet:
OK. We should make that configurable too then

oroulet:
fixed it: https://github.com/FreeOpcUa/opcua-asyncio/pull/560

oroulet:
merged

PR #560, opened 2021-05-03
Hist

oroulet:


asyncua/server/history_sql.py
```diff
@@ -241,7 +238,7 @@ def _get_bounds(start, end, nb_values):
             start_time = end.isoformat(" ")
             end_time = start.isoformat(" ")
         if nb_values:
-            limit = nb_values + 1  # add 1 to the number of values for retrieving a continuation point
+            limit = nb_values
```
HarrySky, line 241: Why +1 was removed here? If I understood it correctly - continuation point is timestamp of next (+1) value, that was not returned in request?

HarrySky (approved):
Hey, left a question, but otherwise looks good to me :shipit:

asyncua/server/history_sql.py
oroulet: nb_values is the limit requested by user, then no cont point is required. There is a limit for server limitation further down. Not really efficient, but that sql code  is more example code than anything else

PR #496, opened 2021-03-23
generate protocol classes from NodeSet.Service.xml and not deprecated…

oroulet:
… bsd

PR #583, opened 2021-06-02
[Fix] Revert to UTC for request header

ZuZuD:
## Issue

We broke UTC timestamp with #496 which means clients in a different TZ than the server aren't able to connect anymore. See the stack trace:

```
BadInvalidTimestamp                       Traceback (most recent call last)
<ipython-input-1-7c6bfb3082fd> in <module>
     33         # We create a Client Subscription.
---> 34         subscription = await c.create_subscription(1000, handler)
     35         #nodes = [

~/github/opcua-asyncio/asyncua/client/client.py in create_subscription(self, period, handler, publishing)
    554         subscription = Subscription(self.uaclient, params, handler)
--> 555         await subscription.init()
    556         return subscription

~/github/opcua-asyncio/asyncua/common/subscription.py in init(self)
     86     async def init(self):
---> 87         response = await self.server.create_subscription(self.parameters, callback=self.publish_callback)
     88         self.subscription_id = response.SubscriptionId  # move to data class

~/github/opcua-asyncio/asyncua/client/ua_client.py in create_subscription(self, params, callback)
    450         request.Parameters = params
--> 451         data = await self.protocol.send_request(request)
    452         response = struct_from_binary(ua.CreateSubscriptionResponse, data)

~/github/opcua-asyncio/asyncua/client/ua_client.py in send_request(self, request, timeout, message_type)
    153
--> 154         self.check_answer(data, f" in response to {request.__class__.__name__}")
    155         return data

~/github/opcua-asyncio/asyncua/client/ua_client.py in check_answer(self, data, context)
    162             hdr = struct_from_binary(ua.ResponseHeader, data)
--> 163             hdr.ServiceResult.check()
    164             return False

~/github/opcua-asyncio/asyncua/ua/uatypes.py in check(self)
    334         if not self.is_good():
--> 335             raise UaStatusCodeError(self.value)
    336

BadInvalidTimestamp: "The timestamp is outside the range allowed by the server."(BadInvalidTimestamp)

During handling of the above exception, another exception occurred:

BadInvalidTimestamp                       Traceback (most recent call last)
```

This diff reverts the headers TS to UTC as it is the time standard for OPC-UA: https://reference.opcfoundation.org/v104/Core/docs/Part3/8.38/

## Testing

Client in a different TZ can connect once patched:
```
In [1]: import sys
   ...: sys.path.insert(0, "../")
   ...: import asyncio
   ...: from asyncua import Node
   ...: from asyncua import ua
   ...: from asyncua import Client
   ...:
   ...: class SubscriptionHandler:
   ...:     """
   ...:     The SubscriptionHandler is used to handle the data that is received for the subscription.
   ...:     """
   ...:     def datachange_notification(self, node: Node, val, data):
   ...:         """
   ...:         Callback for asyncua Subscription.
   ...:         This method will be called when the Client received a data change message from the Server.
   ...:         """
   ...:         _logger.info('datachange_notification %r %s', node, val)
   ...: c = Client(url='opc.tcp://127.0.0.1:4901')
   ...: async with c:
   ...:         handler = SubscriptionHandler()
   ...:         subscription = await c.create_subscription(0.001, handler)
   ...:         print(subscription)
<asyncua.common.subscription.Subscription object at 0x7f65a3dc3400>
```

### DIFF
--- a/asyncua/ua/uaprotocol_auto.py
+++ b/asyncua/ua/uaprotocol_auto.py
@@ -1,6 +1,6 @@
 """
 Autogenerate code from xml spec
-Date:2021-03-25 13:57:54.246658
+Date:2021-06-03 15:58:52.995576
 """
 
 from datetime import datetime
@@ -2666,7 +2666,7 @@ class RequestHeader:
     data_type = NodeId(ObjectIds.RequestHeader)
 
     AuthenticationToken: SessionAuthenticationToken = field(default_factory=NodeId)
-    Timestamp: UtcTime = field(default_factory=datetime.now)
+    Timestamp: UtcTime = field(default_factory=datetime.utcnow)
     RequestHandle: IntegerId = 0
     ReturnDiagnostics: UInt32 = 0
     AuditEntryId: String = None
@@ -2695,7 +2695,7 @@ class ResponseHeader:
 
     data_type = NodeId(ObjectIds.ResponseHeader)
 
-    Timestamp: UtcTime = field(default_factory=datetime.now)
+    Timestamp: UtcTime = field(default_factory=datetime.utcnow)
     RequestHandle: IntegerId = 0
     ServiceResult: StatusCode = field(default_factory=StatusCode)
     ServiceDiagnostics: DiagnosticInfo = field(default_factory=DiagnosticInfo)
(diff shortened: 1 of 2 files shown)


oroulet:
Sorry I screwed your branch by clicking on wrong github button .
But anyway something is wrong here. you only have changes in autogenerated code, that probably should not even be commited (my fault). You probably change the generating code file. Where are these changes?

ZuZuD:
> Sorry I screwed your branch by clicking on wrong github button .
> But anyway something is wrong here. you only have changes in autogenerated code, that probably should not even be commited (my fault). You probably change the generating code file. Where are these changes?

Well caught @oroulet.

I have integrated the time standard change into the generator script `schemas/generate_protocol_python.py` and triggered a new protocol generation, everything looks good now.

PR #587, opened 2021-06-03
add TypeCheck server side

oroulet:
This is a very fast implmentation but it does not check for the reall type but but for the previous type. This is not exactly the same... but maybe it is OK

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -706,12 +706,8 @@ class AddressSpace:
         if attval is None:
             return ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid)
 
-        if value.Value.VariantType != attval.value.Value.VariantType:
-            if value.Value.VariantType == ua.VariantType.Null or  attval.value.Value.VariantType == ua.VariantType.Null:
-                pass
-            else:
-                _logger.critical("Write refused: Variant: %s with type %s does not have expected type: %s", value.Value, value.Value.VariantType, attval.value.Value.VariantType)
-                return ua.StatusCode(ua.StatusCodes.BadTypeMismatch)
+        if not self._is_expected_variant_type(value, attval, node):
+            return ua.StatusCode(ua.StatusCodes.BadTypeMismatch)
 
         old = attval.value
         attval.value = value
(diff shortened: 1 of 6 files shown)


oroulet:
>> Allowing exceptions for this case would may still lead to problems. Otherwise each half-broken client could change int32 to Null and the Problem would still exist.

Valid point. Does spec says anything about that? maybe only admin should be allowed to write Null or maybe nobody...

asyncua/server/address_space.py
```diff
@@ -721,6 +724,27 @@ async def write_attribute_value(self, nodeid, attr, value):
 
         return ua.StatusCode()
 
+    def _is_expected_variant_type(self, value, attval, node):
+        if value.Value.VariantType == ua.VariantType.Null:
+            # we accept overwrite with Null, not sure if this is OK in spec...
+            return True
```
oroulet, line 730: maybe we should not allow. I will try later, if needed, maybe admin should be able to do that

swamper123: [I found that piece in Part4 5.10.4.](https://reference.opcfoundation.org/v104/Core/docs/Part4/5.10.4/):

> 
If a Server allows writing of Attributes with the DataType LocalizedText, the Client can add or overwrite the text for a locale by writing the text with the associated LocaleId. **Writing a null String for the text for a locale shall delete the String for that locale**. Writing a null String for the locale and a non-null String for the text is setting the text for an invariant locale. Writing a null String for the text and a null String for the locale shall delete the entries for all locales. If a Client attempts to write a locale that is either syntactically invalid or not supported, the Server returns Bad_LocaleNotSupported. The Write behaviour for Value Attributes with a LocalizedText DataType is Server specific but it is recommended to follow the same rules.

I guess this was tried to test here:

https://github.com/FreeOpcUa/opcua-asyncio/blob/3b974f1407fa4c90d14a13ae8b18e20e22aef8e1/tests/test_xml.py#L229

But this null string thing makes things ... difficult. Haven't found anything usefull to that topic atm. 😕

oroulet: null string is something else. In that case the Varianttype is String but the string is None.

swamper123: In that perticular case we have to catch DataValue as well, otherwise an AttributeError is thrown.

swamper123: https://reference.opcfoundation.org/v104/Core/docs/Part3/5.2.7/:

_"The optional WriteMask Attribute exposes the possibilities of a client to write the Attributes of the Node. The WriteMask Attribute does not take any user access rights into account, that is, although an Attribute is writable this may be restricted to a certain user/user group.

If the OPC UA Server does not have the ability to get the WriteMask information for a specific Attribute from the underlying system, it should state that it is writable. If a write operation is called on the Attribute, the Server should transfer this request and return the corresponding StatusCode if such a request is rejected."_

So changing the type is allowed, if WriteMask doesn't exist or it is explicitly allowed to write on an attribute. So overwriting seems to be legit in some cases.

PR #589, opened 2021-06-03
fix ExpandedNodeId DataClass

AndreasHeine:
https://github.com/FreeOpcUa/opcua-asyncio/discussions/588

Issue veryfied with Prosys Simulation Server aswell as with UA CPP Demo Server!

### DIFF
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -289,6 +289,7 @@ def list_to_binary(uatype, val):
 
 
 def nodeid_to_binary(nodeid):
+    print(nodeid)
     if nodeid.NodeIdType == ua.NodeIdType.TwoByte:
         data = struct.pack('<BB', nodeid.NodeIdType.value, nodeid.Identifier)
     elif nodeid.NodeIdType == ua.NodeIdType.FourByte:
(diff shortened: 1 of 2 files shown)


AndreasHeine:
fix order of dataclass filed
need to find all places which broke... node.py, ua binary.py, maybe more places whre ua.NodeId is used and add ua.ExpandedNodeId with some logic

oroulet:
What the issue exactly? Looks like you duplicated a lot of code from NodeId

AndreasHeine:
decodeing error... made a draft not done yet 😉 but tired!

guess we need to split up the two nodeid classes because it does not work at all ^^

AndreasHeine:
we also need tests for that stuff!

oroulet:
what is the issue exactly? The order of fields was wrong?. What is the correct order of fields? I was looking at doc now and the order looked correct...

oroulet:
OK. after a second look, maybe the order is wrong

oroulet:
but instead of copying all that code you make _NodeId base class for example and make both NodeId and ExpandedNodeId inherit them. Just tested now and it works fine with dataclass

oroulet:
Also nice we did not release that code. I was afraid of something like that ;-)

oroulet:
also do you have any ideas about how to compare an ExpandednodeId and a NodeId? maybe they are never the same because we never know was server a NodeId belong to...

AndreasHeine:
let me figure out what all is effected then we can make some classes stuff but if i compare NodeId and ExpandedNodeId in spec. the order in both are different not sure if it is possible in dataclasses to change the order in an inheriting class!? and i guess some of the code needs to be replaced/deleted i just copied it for now to see if the decode error is gone...

https://reference.opcfoundation.org/v104/Core/docs/Part4/7.11/
https://reference.opcfoundation.org/v104/Core/DataTypes/BaseDataType/

they are no subtypes of each other so i think it makes sense to make individual classes!?

oroulet:
1) Even if they are not subtype, they share code for sure. If they share many methods we can make them inherit a common base class.
2) If they share only a few methods we can make a few functions they both call. But I am rather certain 1) is a better solution
Also the code has worked for 10 years with one class for both so somehow it worked with a lot of similar code

oroulet:
I do not think we can change order of members in a dataclass. And even if we can, it is probably a quite ugly hack, so better doing something else

AndreasHeine:
if we make a baseclass with the shared code we need to make some kind of dependency injection

but if i look in released version it also is not working maybe it never has worked? tbh i have never used the ExpandedNodeId stuff....
no test nothing... we should write tests for all dataclasses and make a testrun to encode and decode them

oroulet:
I never really used them but I remember testing them against prosys the first years so something has worked.

oroulet:
not sure what you mean by dependency injection in this case. it should be quite straightforward.
At least if you jsut fix the order and move common code in a base class you should have fixed any serializatino error. Was there other errors?

oroulet:
Does it work now? At least code looks much better.
Also fix linting do github tried to run the tests

AndreasHeine:
no stuck in ua_binary.py the whole struct pack looks strange

oroulet:
Ok. I will have a look i wrote that stuff recently

AndreasHeine:
ExpandedNodeId(ServerIndex=0, NamespaceUri='http://www.prosysopc.com/OPCUA/StaticNodes', NamespaceIndex=0, NodeIdType=<NodeIdType.String: 3>, Identifier='Boolean')

bytearray(b'\x83\x00\x00\x07\x00\x00\x00Boolean*\x00\x00\x00http://www.prosysopc.com/OPCUA/StaticNodes')
Received an error: ErrorMessage(Error=StatusCode(value=2147942400), Reason='Bad_DecodingError (code=0x80070000, description="Unexpected NodeId Encoding Byte -125")')

AndreasHeine:
https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.2/#5.2.2.10

> The ExpandedNodeId is encoded by first encoding a NodeId as described in 5.2.2.9 and then encoding NamespaceUri as a String.

oroulet:
From your link above it's looks like original implementation was correct. At least the order

AndreasHeine:
see my last commit i revoeked some of the stuff there is still something wrong with the encoding of the ExpandedNodeId the regualr is working fine

according to spec. the serverindex has to be zero not null/none
uri can be null/none or empty string

AndreasHeine:
just a question: why we dont pack that classes always in order of the filds?

oroulet:
@AndreasHeine We pack all classes in the order of their field  BUT NodeId for historical reasons I suppose. I did not remember that.
So in fact the entire encoding/decoding is in these function and we should not need to plit NodeId and ExpandedNodeId
But now I checked the code AND the spec and could not find any error. so back to the original question. What is the issue?

AndreasHeine:
long things short:

run this against prosys sim server!
```python
import asyncio
from asyncua import Client, ua

async def main():
    client = Client("opc.tcp://127.0.0.1:4840")
    await client.connect()

    # ExpandedNodeId 
    node = client.get_node("nsu=http://www.prosysopc.com/OPCUA/StaticNodes;s=Boolean")
    res = await client.uaclient.read_attributes([node.nodeid], ua.AttributeIds.Value)
    print(res)

    await client.disconnect()

asyncio.run(main())
```

back to almost original "maybe":
-expanded node id to string is not correct
-serverindex has to be zero not null/none
this also needs to be changed (but it depend on your view if we should mirror the field order like the struct definition in spec.)
https://github.com/FreeOpcUa/opcua-asyncio/blob/7993f14f78ceb2a2eed5ef65d3ed0233edfc2910/asyncua/ua/uatypes.py#L474

oroulet:
image
Neither wireshark parser, nor prosys seems to accept the nodeid here. wireshar stops already at the encoding byte. We wrote 0.83 which I think is correct it 0x3 + 0x80 as in spec. Arre you sure ReadRequest accepts an ExpandedNodeId? Maybe that is the issue

oroulet:
I do not think ExpandedNodeId are accepted everywhere, only at some specific places

oroulet:
I had a look a few places and I think where an ExpandedNodeId is accepted, this is clearly specified. It is not specificied in ReadRequest so it does not accept  an ExpandedNodeId... as far as I understand

AndreasHeine:
hm the difference is "boiled down" you can use the uri instead of index

oroulet:
In that case 0x83 as encoding is refused an 0x03 is accepted. So what do you think the encoding should be?

oroulet:
Btw if you have access to some ua ask. Make a simple client creating an expandedNodeId and make a readRequest. I bet it will fail. Unless the client transform the expandedNodeId to a NodeId in the background but you will see that in Wireshark

AndreasHeine:
still looking some stuff up there is no clear sentence which makes it clear...

https://opcfoundation.org/forum/opc-ua-standard/expanded-nodeid-vs-nodeid/

AndreasHeine:
http://opclabs.doc-that.com/files/onlinedocs/QuickOpc/Latest/User%27s%20Guide%20and%20Reference-QuickOPC/OPC%20UA%20Node%20IDs.html

oroulet:
I looked at opcua structures and ExpandedNodeId are use 36 times. for example AddNodeItems accept an ExpandedNode as parent. ReadRequest ony takes a NodeId. My current conclusion is that you cannot use an ExpandedNodeId there.

AndreasHeine:
if it is not accepted we propably should raise an error if someone passes an expanded nodeid to read, write umd browse?

oroulet:
closing until someone shows there is a bug here

Issue #594, opened 2021-06-07
Parentnodes with same Browsename not allowed, but should be


swamper123:
**Describe the bug** <br />
[From the specs](https://reference.opcfoundation.org/v104/Core/docs/Part4/5.8.4/):

_" Since BrowseNames shall be unique in the context of type definitions, a Client may create a browse path that is valid for a type definition and use this path on instances of the type. [...] If a Node has multiple targets with the same BrowseName, the Server shall return a list of NodeIds. "_

This is not provided in our server, because I missinterpreted it and made everynode Browsename unique.
If a list of Nodes is returned, like metioned in the spec, if multiple Browsenames of a parent are unique, is not tested.

**To Reproduce**<br />
Import two nodes with the same Browsename to the same parent.

**Expected behavior**<br />
In case of browsing, this should return a list, instead of a Node. The Client has to figure out, which Node is wanted.

**Version**<br />
Python-Version: all
opcua-asyncio Version: all

swamper123:
[Answer from Randy Armstrong about special cases of uniqueness](https://opcfoundation.org/forum/opc-ua-standard/uniqueness-of-browsenames/)

> OPC UA is a full mesh network with overlapping hierarchies which means the BrowseName cannot be assumed to be unique.
> 
> The exceptions are:
> 1) Properties. All targets of HasProperty reference from any Node must have unique BrowseNames.
> 2) TypeDefinitions: All targets of hierarchical references from an ObjectType or VariableType must have unique names.
> 3) InstanceDeclarations: All targets of hierarchical references from an Object, Variable or Method InstanceDeclaration must have unique names.


**ToDo:**
Revert uniqueness checks, except if a special case occurs.

jcbastosportela:
> [Answer from Randy Armstrong about special cases of uniqueness](https://opcfoundation.org/forum/opc-ua-standard/uniqueness-of-browsenames/)
> 
> > OPC UA is a full mesh network with overlapping hierarchies which means the BrowseName cannot be assumed to be unique.
> > The exceptions are:
> > 
> > 1. Properties. All targets of HasProperty reference from any Node must have unique BrowseNames.
> > 2. TypeDefinitions: All targets of hierarchical references from an ObjectType or VariableType must have unique names.
> > 3. InstanceDeclarations: All targets of hierarchical references from an Object, Variable or Method InstanceDeclaration must have unique names.
> 
> **ToDo:**
[...]

PR #615, opened 2021-06-29
Fix #611 scrambled datatypes nodeids

### LINKED ISSUE #611: Scrambled DataType NodeIds when using multiple information models
**Describe the bug** <br />
When importing multiple information models, where DataTypes are defined across them and used in the topmost information model, the server will report incorrect Namespace Index for the DataTypes of Varaibles (for example).

As far as I could investigate the issue is in `common\xmlimporter.py::_to_migrated_nodeid` which is then calling `common\xmlimporter.py::_migrate_ns` for nodeids which are `aliases`. AFAI understood the alias are already in the server namespace index, therefore `common\xmlimporter.py::_migrate_ns` should not be called.

I have made a fix that resolves my problem, but I am not sure if it is the correct approach for this:

```python
   def _to_migrated_nodeid(self, nodeid: Union[ua.NodeId, None, str]) -> ua.NodeId:
        needs_migration, nodeid = self._to_nodeid(nodeid)
        if needs_migration: # added this check for migrating namespace conditionally
            nodeid = self._migrate_ns(nodeid)
        return nodeid

    def _to_nodeid(self, nodeid: Union[ua.NodeId, None, str]) -> bool, ua.NodeId:  # returns a pair of values, bool indicates need of ns migration
        if isinstance(nodeid, ua.NodeId):
            return True, nodeid
        if not nodeid:
            return True, ua.NodeId(ua.ObjectIds.String)
        if "=" in nodeid:
            return True, ua.NodeId.from_string(nodeid)
        if hasattr(ua.ObjectIds, nodeid):
            return True, ua.NodeId(getattr(ua.ObjectIds, nodeid))
        if nodeid in self.aliases:
            return False, self.aliases[nodeid]  # aliases do not need migration
        return True, ua.NodeId(getattr(ua.ObjectIds, nodeid))
```

**To Reproduce**<br />
Create 3 information models, each of them defining one custom DataType.
Create other information model which uses the 3 previously created information models; in this one create an ObjectType that contains one instance on each of the DataTypes from the other information models.

You may use the attached files [multiple_inf_models.zip](https://github.com/FreeOpcUa/opcua-asyncio/files/6701602/multiple_inf_models.zip) __ATTENTION__: in order for this information models to work it is necessary to apply the fix I suggest in https://github.com/FreeOpcUa/opcua-asyncio/issues/594#issuecomment-866763055 (Because all of the information models generated by current UaModeler version defines TypeDictionary with same BrowseName Name, but different BrowseName NamespaceIndex)

**Expected behavior**<br />
In the client the correct DataType for each of the variables is reported.

**Screenshots**<br />
This is what I get:
image

This is what I expect to get:
image

**Version**<br />
Python-Version:3.7 and 3.8 (at least, most likely all)
opcua-asyncio Version (e.g. master branch, 0.9): tested in 0.8.2 and in master.

jcbastosportela:
fixes #611 
also https://github.com/FreeOpcUa/opcua-asyncio/issues/594#issuecomment-866763055

__NOTE__ : Sorry for the previous PR. It was having changes from my fork's master where I applied https://github.com/FreeOpcUa/opcua-asyncio/pull/517#issue-602313624 because I need it. In this PR I remove those changes (sorry for the trailing of unrelated commits you see with this PR, I should not have changed my fork's master :( )

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -21,6 +21,7 @@ class XmlImporter:
         self.server = server
         self.namespaces: Dict[int, int] = {}  # Dict[IndexInXml, IndexInServer]
         self.aliases: Dict[str, ua.NodeId] = {}
+        self._unmigrated_aliases: Dict[str, str] = {} # Dict[name, nodeId string]
         self.refs = None
 
     async def _map_namespaces(self):
@@ -96,7 +97,8 @@ class XmlImporter:
         await self.parser.parse(xmlpath, xmlstring)
         self.namespaces = await self._map_namespaces()
         _logger.info("namespace map: %s", self.namespaces)
-        self.aliases = self._map_aliases(self.parser.get_aliases())
+        self._unmigrated_aliases = self.parser.get_aliases()        # these nodeids are not migrated to server namespace indexes
+        self.aliases = self._map_aliases(self._unmigrated_aliases)  # these nodeids are already migrated to server namespace indexes
         self.refs = []
         dnodes = self.parser.get_node_datas()
         dnodes = self.make_objects(dnodes)
@@ -286,8 +288,8 @@ class XmlImporter:
             return ua.NodeId.from_string(nodeid)
         if hasattr(ua.ObjectIds, nodeid):
             return ua.NodeId(getattr(ua.ObjectIds, nodeid))
-        if nodeid in self.aliases:
-            return self.aliases[nodeid]
+        if nodeid in self._unmigrated_aliases:
+            return ua.NodeId.from_string(self._unmigrated_aliases[nodeid])
         return ua.NodeId(getattr(ua.ObjectIds, nodeid))
 
     async def add_object(self, obj, no_namespace_migration=False):


oroulet:
OK the fixes might be correct, but I cannot mere 10 commits for something that should be do.
Just run `git reset --soft THELASTCOMIMTIDFROMMASTER` then make two commits, one for the browsenamefix and one for the xmlimport thing. use `git add -p` for example

jcbastosportela:
> OK the fixes might be correct, but I cannot mere 10 commits for something that should be two.
> Just run `git reset --soft THELASTCOMIMTIDFROMMASTER` then make two commits, one for the browsenamefix and one for the xmlimport thing. use `git add -p` for example

Done. Hope it is right this time. 😃

asyncua/server/address_space.py
```diff
@@ -260,7 +260,7 @@ def _add_node(self, item, user, check=True):
                     ]
                     and ref.IsForward
                 ):
-                    if item.BrowseName.Name == ref.BrowseName.Name:
+                    if item.BrowseName.Name == ref.BrowseName.Name and item.BrowseName.NamespaceIndex == ref.BrowseName.NamespaceIndex:
```
jcbastosportela, line 263: @oroulet I see that you have made the "same" (better) fix here in this PR https://github.com/FreeOpcUa/opcua-asyncio/pull/618#issue-683224939

This will probably be a merge conflict. Should I "uncommit" this file changes?

oroulet: yes I copied that code from you ;-) . you need to rebase and probably delete that commit. should be quick

oroulet: Sorry it was bad practice from my part to copy that but I was unsure how long it would take you to update that one

PR #618, opened 2021-07-04
let  TranslateBrowsePathToNodeId return several nodeids

oroulet:


oroulet:
@swamper123 OK with that one?

asyncua/server/address_space.py
```diff
@@ -162,21 +162,34 @@ def _translate_browsepath_to_nodeid(self, path):
...
-        target.RemainingPathIndex = 4294967295
-        res.Targets = [target]
+        target_nodeids = self._navigate(path.StartingNode, path.RelativePath.Elements)
+        if not target_nodeids:
+            res.StatusCode = ua.StatusCode(ua.StatusCodes.BadNoMatch)
+            return res
+        for nodeid in target_nodeids:
+            target = ua.BrowsePathTarget()
+            target.TargetId = nodeid
+            target.RemainingPathIndex = 4294967295
```
swamper123, line 172: What is this and why is it a hardcoded number?

asyncua/common/node.py
```diff
@@ -500,7 +500,8 @@ async def get_child(self, path):
         result = await self.server.translate_browsepaths_to_nodeids([bpath])
         result = result[0]
         result.StatusCode.check()
-        # FIXME: seems this method may return several nodes
+        if return_all:
```
swamper123, line 503: This should be normal practice, not an exclusive one. Otherwise `translate_browsepaths_to_nodeids` would be inefficient and the specs expect such a behaviour, that if multiple Nodes in the same hierarchy exist, all should be returned.

swamper123 (changes requested):

asyncua/server/address_space.py
oroulet: I have no ideas.... it was there. probably som max uint or similar

asyncua/common/node.py
oroulet: this is necessary for compatibility. Otherwise a lot of existing code will fail. Otherwise we need to add a new call and that is also messy.
One can always call directly translate_browsepaths_to_nodeid. 
get_child is jsut some high level python-asyncio invention

asyncua/server/address_space.py
swamper123: <html><body>


remainingPathIndex | Index | The index of the first unprocessed element in the RelativePath.This value shall be equal to the maximum value of Index data type if all elements were processed (see 7.13 for Index definition).
-- | -- | --



</body>
</html>

asyncua/common/node.py
swamper123: But undefined behaviour (which Node is now selected ?!) is also messy.

asyncua/server/address_space.py
oroulet: so that is probably what is done. we could add some doc

Issue #635, opened 2021-07-30
Server doesn't support index range for Array variables


jcbastosportela:
**Describe the bug** <br />
According to the standard the server should reply to a read of an array that defines `AttributeIds.IndexRange` with a sub array of the given range, however the server always replies with the full array.
see:
https://reference.opcfoundation.org/v104/Core/docs/Part4/7.24/
https://reference.opcfoundation.org/v104/Core/docs/Part4/7.22/

**To Reproduce**<br />
1. create a server providing an array
2. create a client that reads the `Value` attribute of the array giving also `IndexRange` attribute with a valid range

**Expected behavior**<br />
Client receives an array that matches the requested range

**Screenshots**<br />
![image](https://user-images.githubusercontent.com/15647115/127634516-6c2c46dc-e97f-45e1-8191-d8506aa78141.png)

**Version**<br />
Python-Version:3.8<br />
opcua-asyncio Version (e.g. master branch, 0.9):at least 0.8.2 and above

Check the server and client in the attachment.
[indexrange_issue.zip](https://github.com/FreeOpcUa/opcua-asyncio/files/6906646/indexrange_issue.zip)

I have identified a potential solution by changing:
- server/address_space.py::AttributeService::read()
- server/address_space.py::AddressSpace::read_attribute_value()

PR #636, opened 2021-07-30
Server support for read array with index range

jcbastosportela:
solves https://github.com/FreeOpcUa/opcua-asyncio/issues/635

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -1,4 +1,6 @@
 import asyncio
+import typing
+from asyncua.ua.uatypes import DataValue, NodeId, Variant
 import pickle
 import shelve
 import logging
@@ -47,7 +49,7 @@ class AttributeService:
         # self.logger.debug("read %s", params)
         res = []
         for readvalue in params.NodesToRead:
-            res.append(self._aspace.read_attribute_value(readvalue.NodeId, readvalue.AttributeId))
+            res.append(self._aspace.read_attribute_value(readvalue.NodeId, readvalue.AttributeId, readvalue.IndexRange))
         return res
 
     async def write(self, params, user=User(role=UserRole.Admin)):
@@ -686,7 +688,7 @@ class AddressSpace:
 
         self._nodes = LazyLoadingDict(shelve.open(path, "r"))
 
-    def read_attribute_value(self, nodeid, attr):
+    def read_attribute_value(self, nodeid:NodeId, attr:ua.AttributeIds, index_range: typing.Optional[str] = None) -> DataValue:
         # self.logger.debug("get attr val: %s %s", nodeid, attr)
         if nodeid not in self._nodes:
             dv = ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadNodeIdUnknown))
@@ -696,9 +698,29 @@ class AddressSpace:
             dv = ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid))
             return dv
         attval = node.attributes[attr]
+        value = ua.DataValue()
         if attval.value_callback:
-            return attval.value_callback()
-        return attval.value
+            value = attval.value_callback()
+        else:
+            value = attval.value
+        if not index_range is None:
+            index_range_list = index_range.split(':')
+            try:
+                low_idx = int(index_range_list[0])
+                if len(index_range_list) == 1:
+                    high_idx = int(index_range_list[0])+1
+                elif len(index_range_list) == 2:
+                    high_idx = int(index_range_list[1])+1 # python List[n:n] -> empty list, List[n:n+k] -> list of k elements; OPC UA see https://reference.opcfoundation.org/v104/Core/docs/Part4/7.22/
+                else:
+                    self.logger.info(f'Read with bad index range: {index_range}')
+                    dv = ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadIndexRangeInvalid))
+                    return dv
+                value = DataValue(Variant(value.Value.Value[low_idx:high_idx]))
+            except Exception as ex:
+                self.logger.warning(f'Read index range failed. IndexRange = {index_range}; Ex: {ex}')
+                dv = ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadIndexRangeInvalid))
+                return dv
+        return value
 
     async def write_attribute_value(self, nodeid, attr, value):
         # self.logger.debug("set attr val: %s %s %s", nodeid, attr, value)


oroulet:
Is that working? Nobody is supposed to use address space class directly...
Also add a test showing how to use that feature.
And move all string parsing at higher level. At address space level you can pass a tuple of indexes

jcbastosportela:
> Is that working? Nobody is supposed to use address space class directly...

Well, it works. I honestly don't thoroughly understand the inner of the project, more form the user perspective. However, after debugging and trying to follow the code I have came to the conclusion that when an external read (a read requested issued by a client) lands on the `address_space.py::AttributeService::read()`

`uaprocessor.py::_process_message()` calls the:
__`internal_session.py::InternalSession::read()` that calls:
____`internal_server.py::InternalServer::attribute_service::read` meaning `address_space.py::AttributeService::read()`

`InternalSession` holds an instance of `internal_server.py::InternalServer` which instantiates `address_space.py::AddressSpace` used to instantiate `address_space.py::AttributeService`

Now, if you ask me, I wonder why is the `InternalSession` calling the read via the `InternalServer::attribute_service` member's `read` method when `InternalServer` provides a method `InternalServer::read_attribute_value` that would call the `address_space.py::AddressSpace::read_attribute_value`.

Do you think this needs fixing also, or is this intentional and is like so for a reason. Maybe someone can help me in this point. 

> Also add a test showing how to use that feature.

I am not sure if I will know how to use the tests you have setup, but will give it a try.

> And move all string parsing at higher level. At address space level you can pass a tuple of indexes

I can do that.

Moreover, as far as I can see this functionality is also missing for the write operation, maybe can be addressed later in other PR if I can understand the "correct" fix for this.

jcbastosportela:
with this last changes address_space.py::AttributeService seems that is rendered unused. If this changes are okay maybe it must be removed. I tried writing tests, but it wasn't working correctly, I have to look after, for now maybe you can understand what I am trying to solve by looking the zip attached (server+xml, and the client)
[examples.zip](https://github.com/FreeOpcUa/opcua-asyncio/files/6910548/examples.zip)

asyncua/server/address_space.py
```diff
@@ -696,11 +710,35 @@ def read_attribute_value(self, nodeid, attr):
...
+        value = ua.DataValue()
         if attval.value_callback:
-            return attval.value_callback()
-        return attval.value
-
-    async def write_attribute_value(self, nodeid, attr, value):
+            value = attval.value_callback()
+        else:
+            value = attval.value
+        if not index_range is None:
```
oroulet, line 718: if index_range is not None if the advices syntax.
But I am rather sure you could write: `if index_range`

asyncua/server/address_space.py
```diff
@@ -696,11 +710,35 @@ def read_attribute_value(self, nodeid, attr):
             dv = ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid))
             return dv
         attval = node.attributes[attr]
+        value = ua.DataValue()
```
oroulet, line 713: also it looks like the logic is more complicated than necessary. That  local variable is useless.

asyncua/server/address_space.py
```diff
@@ -696,11 +710,35 @@ def read_attribute_value(self, nodeid, attr):
...
         if attval.value_callback:
-            return attval.value_callback()
-        return attval.value
-
-    async def write_attribute_value(self, nodeid, attr, value):
+            value = attval.value_callback()
```
oroulet, line 715: just return here insteas of making a local variable

asyncua/server/address_space.py
```diff
@@ -696,11 +710,35 @@ def read_attribute_value(self, nodeid, attr):
...
+        else:
+            value = attval.value
```
oroulet, line 717: same here

asyncua/server/address_space.py
oroulet: and that test should be ealier, so you do not need to overwrite a local value

jcbastosportela: In fact IMO if we don't make the test here (after knowing if the value is read via `attval.value_callback` or via `attval.value`) it makes the implementation more complex.

The check if just a range is wanted or full array wanted relies on having the full array already, specially because I don't understand when `attval.value_callback()` is called and how does it work; as far as I can see it is not used, I couldn't find any location on the code where it is set. If you or someone can confirm that it can be removed, than the code can be simplified.

So I think this implementation is safer, but eventually worst performing (we can make some lambdas for trying to minimize number of copies, but...). In case `value_callback` is used, in order to make the check before we would need to change the `value_callback` so it could also take a range.

Other option is to sacrifice performance only when the value is obtained via `value_callback` and having some code redundancy for range (maybe mitigated with some helper function). 

What do you recon to be the best alternative?

jcbastosportela: @oroulet I have some interest in having this feature fixed 😸 
Could you help me getting this PR in shape to get merged?

asyncua/common/ua_utils.py
```diff
@@ -14,6 +15,26 @@
...
 
+def ua_index_range_to_list_slice(ua_index_range:str) -> Tuple[int,int]:
+    """Converts an OPC UA NumericRange into a python tuple for list slicing
+    see: https://reference.opcfoundation.org/v104/Core/docs/Part4/7.22/
+    Args:
+        ua_index_range (str): The OPC UA NumericRange
+
+    Returns:
+        Tuple[int,int]: Equivalent python list slicing tuple with low and high bounds
+    """
```
oroulet, line 26: good method

asyncua/common/ua_utils.py
```diff
@@ -14,6 +15,26 @@
...
+    index_range_list = ua_index_range.split(':')
+    low_idx = int(index_range_list[0])
+    if len(index_range_list) == 1:
+        high_idx = int(index_range_list[0])+1
+    elif len(index_range_list) == 2:
+        high_idx = int(index_range_list[1])+1 # python List[n:n] -> empty list, List[n:n+k] -> list of k elements; OPC UA see https://reference.opcfoundation.org/v104/Core/docs/Part4/7.22/
```
oroulet, line 32: please do nt write so long lines. split it. And also I am sure linters complain about nothaving 2 spaces before #

asyncua/common/ua_utils.py
```diff
@@ -14,6 +15,26 @@
...
+    else:
+        raise Exception(f'Read with bad index range: {ua_index_range}')
```
oroulet, line 34: I thin ValueError is the correct excpetino to use in this case. Unless UA defines one for such case

asyncua/server/address_space.py
oroulet: I just had a second look. That variable is really completely useless. It is ALWAYS overwritten in one of the next 4 lines

asyncua/server/address_space.py
```diff
@@ -696,11 +710,35 @@ def read_attribute_value(self, nodeid, attr):
...
+        if not index_range is None:
+            if not attval.value.Value.is_array:
+                return ua.StatusCode(ua.StatusCodes.BadWriteNotSupported)
+            try:
+                value = DataValue(Variant(value.Value.Value[index_range[0]:index_range[1]]))
+            except Exception as ex:
```
oroulet, line 723: catching exception is usually a bad idea. In that case there are two things that can happen: the value is not an array or indexes are out of bound. Only catch these two things

asyncua/server/address_space.py
oroulet: fix that one: `if index_range is not None` is the advices syntax.

asyncua/server/internal_server.py
```diff
@@ -292,18 +294,34 @@ def unsubscribe_server_callback(self, event, handle):
...
-    async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
+    async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value, index_range:Optional[str]=None)->StatusCode:
         """
         directly write datavalue to the Attribute, bypassing some checks and structure creation
         so it is a little faster
         """
-        await self.aspace.write_attribute_value(nodeid, attr, datavalue)
-
-    def read_attribute_value(self, nodeid, attr=ua.AttributeIds.Value):
+        if not index_range is None:
```
oroulet, line 302: same here: `if index_range is not None`

asyncua/server/internal_server.py
```diff
@@ -292,18 +294,34 @@ def unsubscribe_server_callback(self, event, handle):
...
+            try:
+                py_list_index_range = ua_index_range_to_list_slice(index_range)
+            except Exception as ex:
```
oroulet, line 305: probably possible to only catch ValueError here

asyncua/server/internal_server.py
```diff
@@ -292,18 +294,34 @@ def unsubscribe_server_callback(self, event, handle):
...
+        else:
+            py_list_index_range = None
+        return await self.aspace.write_attribute_value(nodeid, attr, datavalue, py_list_index_range)
+
+    def read_attribute_value(self, nodeid:NodeId, attr=ua.AttributeIds.Value, index_range:Optional[str]=None)->DataValue:
         """
         directly read datavalue of the Attribute
         """
-        return self.aspace.read_attribute_value(nodeid, attr)  
+        if not index_range is None:
```
oroulet, line 316: same here

asyncua/server/internal_server.py
```diff
@@ -292,18 +294,34 @@ def unsubscribe_server_callback(self, event, handle):
...
+            try:
+                py_list_index_range = ua_index_range_to_list_slice(index_range)
+            except Exception as ex:
+                self.logger.warning(f'Parse index range failed. IndexRange = {index_range}; Ex: {ex}')
+                return ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadIndexRangeInvalid))
+        else:
+            py_list_index_range = None
```
oroulet, line 323: also that code is exactly the same in both read and write, put it in an internal method

asyncua/server/internal_session.py
```diff
@@ -102,29 +105,78 @@ def activate_session(self, params, peer_certificate):
...
             user = User()
         else:
             user = self.user
         await self.iserver.callback_service.dispatch(CallbackType.PreRead,
                                                      ServerItemCallback(params, None, user))
-        results = self.iserver.attribute_service.read(params)
+        results = [
+                    self.iserver.read_attribute_value(node_to_read.NodeId, node_to_read.AttributeId, node_to_read.IndexRange)
+                    for node_to_read in params.NodesToRead
+                ]
```
oroulet, line 126: why did you change passing all arguments to passing them one by one? you are proably taking away some arguments, and the loop was probably anyway inside the read() method in iserver

asyncua/server/server.py
```diff
@@ -663,15 +664,15 @@ async def load_enums(self) -> Coroutine:
         _logger.warning("Deprecated since spec 1.04, call load_data_type_definitions")
         return await load_enums(self)
 
-    async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
+    async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value, index_range:Optional[str]=None)->StatusCode:
```
oroulet, line 667: OK, this is supposed to be the deficient server side write method, there we should use python indexes, not string that need to be parsed

asyncua/server/server.py
```diff
@@ -663,15 +664,15 @@ async def load_enums(self) -> Coroutine:
...
         """
         directly write datavalue to the Attribute, bypasing some checks and structure creation
         so it is a little faster
         """
-        return await self.iserver.write_attribute_value(nodeid, datavalue, attr)
+        return await self.iserver.write_attribute_value(nodeid, datavalue, attr, index_range)
 
-    def read_attribute_value(self, nodeid, attr=ua.AttributeIds.Value):
+    def read_attribute_value(self, nodeid, attr=ua.AttributeIds.Value, index_range:Optional[str]=None)->DataValue:
```
oroulet, line 674: same here. string parsing is slow

asyncua/server/internal_server.py
```diff
@@ -292,18 +294,34 @@ def unsubscribe_server_callback(self, event, handle):
         """
         self.callback_service.removeListener(event, handle)
 
-    async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value):
+    async def write_attribute_value(self, nodeid, datavalue, attr=ua.AttributeIds.Value, index_range:Optional[str]=None)->StatusCode:
```
oroulet, line 297: same here, this is is supposed to be an efficient method. use indexes

PR #643, opened 2021-08-11
fix: robust stop of time_task of iserver

kozachynskyi:
Usage of loop and status of time_task is not required
It's better to use task.cancel(), because it might avoid deadlock here

I might found a bug, when iserver doesn't stop correctly, because ``await self.time_task`` is never awaited. This might be, because ``await self.current_time_node.write_value(datetime.utcnow())`` is hanging. In any case my example is stopping correctly, when I use this fix.

### DIFF
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -71,7 +71,6 @@ class InternalServer:
         )
         self.current_time_node = Node(self.isession, ua.NodeId(ua.ObjectIds.Server_ServerStatus_CurrentTime))
         self.time_task = None
-        self._time_task_stop = False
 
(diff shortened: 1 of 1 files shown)


oroulet:
Ok. Are you 100% sure using cancel if the advices method to stop a task from python development?
My own experience is that it is similar to ask the Java threading stuff where they tried 100 things and then went back saying using some variable is in fact the safest.
I have had many issues with that cancel() stuff failing...

kozachynskyi:
I'm not sure. Let me dig deeper and check everything.

kozachynskyi:
I don't think that it's the best way now, so I'll close the request

PR #663, opened 2021-09-17
External argument to server callbacks

Marco9412:
Added a field to ServerItemCallback class used by CallbackManager to distinguish events triggered by server itself from one from external clients.

### DIFF
--- a/asyncua/common/callback.py
+++ b/asyncua/common/callback.py
@@ -37,7 +37,7 @@ class Callback(object):
 
 
 class ServerItemCallback(Callback):
-    def __init__(self, request_params, response_params, is_internal=False, user=None):
+    def __init__(self, request_params, response_params, user=None, is_internal=False):
         self.request_params = request_params
         self.response_params = response_params
         self.event_comes_from_internal_session = is_internal
--- a/asyncua/common/callback.py
+++ b/asyncua/common/callback.py
@@ -37,10 +37,10 @@ class Callback(object):
 
 
 class ServerItemCallback(Callback):
-    def __init__(self, request_params, response_params, user=None, is_internal=False):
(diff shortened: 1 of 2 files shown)


asyncua/common/callback.py
```diff
@@ -37,9 +37,10 @@ def getName(self):
 
 
 class ServerItemCallback(Callback):
-    def __init__(self, request_params, response_params, user=None):
+    def __init__(self, request_params, response_params, user=None, is_internal=False):
```
oroulet, line 40: an where is that is_internal flag set?
As far as I remember the architecture there is no way to differentiate between a write from external or internal clients. We have something called InternalSession but a user can easily create more internal sessions wit different users.
What we could do is a Role InternalAdmin and maybe InternalUser, modify InternalSession to use that user (one liner), modify a bit address space to handle the InternalAdmin role and add a check on these users in your code

Marco9412: @oroulet I used the "external" parameter of the InternalSession object defined here [https://github.com/FreeOpcUa/opcua-asyncio/blob/2f0c9fc1d40252e08e64242ba0f65465dd077feb/asyncua/server/internal_session.py#L33](url), which is used to distinguish the "local" session used by the library itself from the external ones.
Anyway also your solution should work as well, so if you prefer I can implement it in a few days.

oroulet: ahh. great. I was looking for that boolean but did not find it. I see it makes sense. Maybe it is better than using users.
Så let's keep it as you have it now. although I thing it would be nice to avoid this "not self.externl" everywhere, canåt you use an is_external bool as far down as possible so we do not need to negate it everywhere?

Issue #689, opened 2021-10-17
is_array is False even if the variant is an array


ybozer:
**Describe the bug** <br />
"is_array" of the Variant is set to "False" for arrays. 

**To Reproduce**<br />
1. Add an array variable in server.
2. Read the array by using server.read_attribute_value() method.
3. Check the "is_array" attribute of the array.

**Expected behavior**<br />
The array's "is_array" attribute is True.

**Screenshots**<br />
![image](https://user-images.githubusercontent.com/92372842/137623565-dd30c2e0-6574-4bf0-887b-d461bafe8def.png)

![image](https://user-images.githubusercontent.com/92372842/137623634-d4a0b362-749a-459b-97f0-34ef1d2c635b.png)

**Version**<br />
Python-Version: 3.7.3<br />
opcua-asyncio Version (e.g. master branch, 0.9): master branch

AndreasHeine:
https://github.com/FreeOpcUa/opcua-asyncio/blob/a69566c5ae926233e85be3b77c4f5480f148614e/asyncua/ua/uatypes.py#L783

looks like something is broken since Dataclass update...

AndreasHeine:
https://github.com/FreeOpcUa/opcua-asyncio/blob/a69566c5ae926233e85be3b77c4f5480f148614e/asyncua/ua/uatypes.py#L842

if i use value = [0,1,2,3,4,5]
this looks wrong -> dims = [6]

AndreasHeine:
https://github.com/FreeOpcUa/opcua-asyncio/blob/a69566c5ae926233e85be3b77c4f5480f148614e/asyncua/ua/uatypes.py#L804

should be "none" i guess

oroulet:
I tried to add back the is_array for compatibility but it adds a lot if complicated if sentences.

oroulet:
one more attempt to fix this correctly. Basiclaly I did what @AndreasHeine had in his PR apart from that dims > 1 stuff. As far as I understand we do not need to set that dimension variable unless requested by user or if we have more than one dimension. In fact setting that dimension seems to break uaexpert. Please test

oroulet:
https://github.com/FreeOpcUa/opcua-asyncio/pull/717

oroulet:
test master I merged since it was broken

PR #690, opened 2021-10-17
bugfix: for #689

AndreasHeine:


oroulet:
we should really add a test for that before merging so we do not get that bug again

oroulet:
inspired by your work I wrote that one which tries to fix things much deeper. I am not 100% sure it is correct:
https://github.com/FreeOpcUa/opcua-asyncio/pull/691

AndreasHeine:
ok

PR #691, opened 2021-10-17
remove is_array argument from Variant, it is redundant, one can speci…

oroulet:
…fy Dimensions


AndreasHeine:
@oroulet breaking change on next release!

oroulet:
the problem is that I would like to avoid increasing main version number... not sure what to do

AndreasHeine:
dont break the api ^^

AndreasHeine:
add it to the release notes?

PR #696, opened 2021-10-22
Add low-level file system functionality

### LINKED ISSUE #695: Extension of the FileSystem functionality [enhancement]
In OPC UA, FileSystem functionalities are realized with variables and methods. The nature of the nodes used to map a remote file system is described in the OPC UA specification.

See:
OPC 10000-5: OPC Unified Architecture V1.04
Part 5: Information Model - Annex C (normative) File Transfer

Currently only the functionality to read small files is implemented in the Client. To be able to work with FileSystem in a reasonable way, at least the parts of the specification described as mandatory are required.

Since OPC UA FileSystem differs partly strongly from the python build-in way of working, it would be nice to have an OPC UA typical (low-level) and a python typical (high-level) interface.

CuriousCrook:
Closes #695 

This PR contains all FileSystem functions marked as mandatory in the OPC UA Spec.

See:
OPC 10000-5: OPC Unified Architecture V1.04
Part 5: Information Model - Annex C (normative) File Transfer

The FileType functionality "./client/ua_file_system.py - class UaFile" is not compatible to the "old" implementation in "./client/ua_file.py".
I decided to do this because the build-in python open-mode differs significantly from OPC UA open-mode. The open-mode written in the OPC UA spec. does not map unambiguously to the char-based open-mode in python build-in.

Maybe the old ua_file.py should be removed. I think that is justifiable, because ua_file.py::UaFile() was incomplete and buggy. For example, it reacted differently when used in the context manager or without. Switching to the new variant is very simple. You just have to replace the open-mode 'r' with 'asyncua.ua.OpenFileMode.Read.value'. Everything else was not supported by the old UaFile anyway.

In the next step I want to write a high-level FileSystem API. The goal of this additional interface should be to react as similar as possible to the build-in Python file handling. Besides that the high-level interface should also handle the splitting of large files.

I did the tests with a proprietary OPC UA server. This embeds a Linux file system in the OPC UA address space. PyTests are also available. If someone knows a good demo server with a file system, I can prepare autotests for it.

Relevant questions, tips and hints are welcome.

With kind regards
Curious Crook

### DIFF
--- a/asyncua/client/ua_file_transfer.py
+++ b/asyncua/client/ua_file_transfer.py
@@ -3,10 +3,7 @@ Low level implementation of OPC UA File Transfer.
 This module contains the mandatory functionality specified
 by the OPC Foundation.
 
-The Read() and Write() functions only support files up to about 16MB in size.
-Larger files must be transferred in several 'pieces'.
-
-See:
+See also:
 OPC 10000-5: OPC Unified Architecture V1.04
 Part 5: Information Model - Annex C (normative) File Transfer
 https://reference.opcfoundation.org/Core/docs/Part5/C.1/
@@ -44,7 +41,7 @@ class UaFile:
         :param open_mode: Open mode defined in C.2.1.
         :return: The file handle.
         """
-        _logger.debug(f"Request to open file {self._file_node} in mode:  {OpenFileMode}")
+        _logger.debug("Request to open file %s in mode: %s", self._file_node, OpenFileMode)
         open_node = await self._file_node.get_child("Open")
         arg1_mode = Variant(open_mode or self._open_mode, VariantType.Byte)
         self._file_handle = await self._file_node.call_method(open_node, arg1_mode)
@@ -54,7 +51,7 @@ class UaFile:
         Close is used to close a file represented by a FileType.
         When a client closes a file the handle becomes invalid.
         """
-        _logger.debug(f"Request to close file {self._file_node}")
+        _logger.debug("Request to close file %s", self._file_node)
         read_node = await self._file_node.get_child("Close")
         arg1_file_handle = Variant(self._file_handle, VariantType.UInt32)
         await self._file_node.call_method(read_node, arg1_file_handle)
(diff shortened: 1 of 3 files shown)


oroulet:
Is this finished? you still have a draft flag on that PR.

CuriousCrook:
> Is this finished? you still have a draft flag on that PR.

If it's okay that we don't have any tests for now, I'm done.

asyncua/client/ua_file_transfer.py
```diff
@@ -0,0 +1,255 @@
...
+        SetPosition is used to set the current position of the file handle.
+        :param position: The position to be set for the fileHandle in the file.
+        If a Read or Write is called it starts at that position.
+        If the position is higher than the file size the position is set to the end of the file.
+        """
+        _logger.debug("Request to set position in file %s", self._file_node)
+        set_position_node = await self._file_node.get_child("SetPosition")
+        arg1_file_handle = Variant(self._file_handle, VariantType.UInt32)
+        arg2_position = Variant(position, VariantType.UInt64)
+        return await self._file_node.call_method(set_position_node, arg1_file_handle, arg2_position)
```
oroulet, line 111: you could just write `return await self._file_node.call_method("SetPosition",  arg1_file_handle, arg2_position)` since anyway call_method will do a query to get the nodeid. 
But if these methods are going to be called many times, many it is worth caching the nodeid to avoid a network call?

for example in open() you could query for all children of that node.

oroulet:
> > Is this finished? you still have a draft flag on that PR.
> 
> If it's okay that we don't have any tests for now, I'm done. Is it?

what does it require on server side? just adding some methods to a node?
Should be easy to make a class taking a node and a file and translate from ua to file but ok...

CuriousCrook:
> > > Is this finished? you still have a draft flag on that PR.
> > 
> > 
> > If it's okay that we don't have any tests for now, I'm done. Is it?
> 
> what does it require on server side? just adding some methods to a node? Should be easy to make a class taking a node and a file and translate from ua to file but ok...

To be honest, I have no idea what to do on the server side.
I would expect from an OPC UA server toolkit that you can specify a path in the local file system and that the toolkit then maps this in the address space in accordance with the specification. But for me that sound like a lot of work.

asyncua/client/ua_file_transfer.py
CuriousCrook: Good idea. But then there would be a few more NodeIds that could be cached in a meaningful way. I'll make a suggestion.

PR #700, opened 2021-10-26
3x performance improvement for serialization

### LINKED ISSUE #692: message serialization is slow
Hello, and thank you for this great library !

  We use it on embedded hardware in an environment where performance matters, and after some profiling, we noticed that around 10% of the run time of our program is spent in the single function [`ua::ua_binary::struct_to_binary`](https://github.com/FreeOpcUa/opcua-asyncio/blob/b19881b91a3efe8f10eff9dfff213e822812e734/asyncua/ua/ua_binary.py#L227).
  
  I have a few ideas to improve performance of serialization in asyncua, and before I start writing code, I would like to know if you would be ready to merge pull requests that slightly increase the code complexity for performance gains.
  
  I am mainly thinking about two things: 
  
  - generating the serialization functions ahead of time instead of processing each dataclass field type every time an object instance is serialized 
  - reducing the amount of small bytes() object allocation and concatenation by pre-allocationg buffers and using struct.pack_into on them.

lovasoa:
Instead of re-inspecting the type metadata everytime an object is serialized,
create a single serialization function for each type and just call it when an instance is serialized

Fixes #692

### DIFF
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -72,6 +72,7 @@ class InternalServer:
         self.current_time_node = Node(self.isession, ua.NodeId(ua.ObjectIds.Server_ServerStatus_CurrentTime))
         self.time_task = None
         self._time_task_stop = False
+        self.match_discovery_source_ip: bool = True 
 
     async def init(self, shelffile=None):
         await self.load_standard_address_space(shelffile)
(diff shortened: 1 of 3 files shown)


lovasoa:
@oroulet : When do you think we can have a new release ? I would need the changes from this PR in my project...

oroulet:
I can release if you want

lovasoa:
Great thank you !

oroulet:
done. great if you confirm , there is not issues

lovasoa:
Thanks ! I updated to the latest version in my project, everything seems to work like before, and opcua serialization doesn't show up as high as before in performance  profiles 👍 
I'll keep you updated here if I notice anything that doesn't work the way it should after more testing.

PR #705, opened 2021-11-05
Reduction of network access

CuriousCrook:
Cache the nodes of potentially multiple-used FileType methods.


CuriousCrook:
@oroulet Here you will find the optimisation you suggested in #696.

PR #706, opened 2021-11-05
3x performance improvement in deserialization

lovasoa:
This pull request applies the same changes as https://github.com/FreeOpcUa/opcua-asyncio/pull/700 to the deserialization code, with similar speedups measured in my tests. As before, most of the deserialization time is now spent creating the objects, as opposed to introspecting the object types.

I also changed the timestamp object serialization function. The new function is ~7 times faster than the old one, resulting in a further 1.5x speedup in my serialization test.

### DIFF
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -8,11 +8,11 @@ import logging
 from typing import Any, Callable
 import uuid
 from enum import IntEnum, Enum, IntFlag
-from dataclasses import is_dataclass, fields, Field, MISSING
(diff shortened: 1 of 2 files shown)


lovasoa:
The test failure is `OSError: [Errno 98] error while attempting to bind on address ('127.0.0.1', 48516): address already in use`. It looks like a bug in the tests more than this PR...

oroulet:
@ZuZuD Looks like your ha_client has issues with python10. I fixed the obvious things, but looks like you still have some tasks crashing at closing of client. Can you have a look?l

lovasoa:
Thanks !

PR #707, opened 2021-11-05
iupdate CI python versions

oroulet:


oroulet:
@ZuZuD Looks like your ha_client has issues with python10. I fixed the obvious things, but looks like you still have some tasks crashing at closing of client. Can you have a look?l

PR #711, opened 2021-11-10
[HaClient] Fix test race condition

ZuZuD:
According to the logs of the recent test failure, there's a small window when we test the connection is established where we have a socket connected but don't have the session created yet, and this is throwing an exception when trying to disconnect because the renew loop is not yet created.

```
=================================== FAILURES ===================================
_________________________ TestHaManager.test_reconnect _________________________

self = <tests.test_ha_client.TestHaManager object at 0x7f8614996e30>
ha_client = <asyncua.client.ha.ha_client.HaClient object at 0x7f86158c9ea0>

    @pytest.mark.asyncio
    async def test_reconnect(self, ha_client):
        await ha_client.start()
        await wait_clients_socket(ha_client, UASocketProtocol.OPEN)
        for c in ha_client.get_clients():
>           await c.disconnect()

tests/test_ha_client.py:348: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
asyncua/client/client.py:261: in disconnect
    await self.close_session()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Client(opc.tcp://127.0.0.1:40399)

    async def close_session(self) -> Coroutine:
        """
        Close session
        """
>       self._renew_channel_task.cancel()
E       AttributeError: 'NoneType' object has no attribute 'cancel'

asyncua/client/client.py:516: AttributeError
------------------------------ Captured log call -------------------------------
INFO     asyncua.client.ha.ha_client:ha_client.py:503 Starting HaManager loop, checking every 1sec
INFO     asyncua.client.ha.reconciliator:reconciliator.py:83 Starting Reconciliator loop, checking every 1sec
INFO     asyncua.client.ha.reconciliator:reconciliator.py:152 [PASS] No configuration difference for healthy targets: set()
INFO     asyncua.client.ha.reconciliator:reconciliator.py:94 [TIME] Reconciliation: 0.00sec
INFO     asyncua.client.ha.ha_client:ha_client.py:549 Virtually reconnecting and resubscribing Client(opc.tcp://127.0.0.1:40399)
INFO     asyncua.client.client:client.py:259 disconnect
INFO     asyncua.client.client:client.py:242 connect
[...]

PR #716, opened 2021-11-11
[Safe Client disconnect]

ZuZuD:
Follow-up of #711 

Make sure calling disconnect is safe when connection isn't established or half-established (i.e: protocol/transport exist but session is not ready).

## Test1
Call to disconnect is safe when there's no existing connection
```
> python -m IPython
Python 3.10.0 (default, Oct 13 2021, 06:45:00) [Clang 13.0.0 (clang-1300.0.29.3)]

In [1]: import sys; sys.path.insert(0, "~/Documents/github/opcua-asyncio"); import asyncua; c = asyncua.Client(url="opc.tcp://localhost:4840",timeout=10
   ...: );

In [2]: await c.disconnect()
close_session but connection wasn't established
close_secure_channel was called but connection is closed

In [3]:
```
## Test2
Ensure that call to disconnect is safe when there's no session established.
To simulate this, I add delay to the server internal_session on session creation. I check that the client connects and disconnects from the server logs. Finally, the client doesn't raise any error.
```
> python -m IPython
Python 3.10.0 (default, Oct 13 2021, 06:45:00) [Clang 13.0.0 (clang-1300.0.29.3)]
In [1]: import sys; sys.path.insert(0, "~/Documents/github/opcua-asyncio"); import asyncua; c = asyncua.Client(url="opc.tcp://localhost:4840",timeout=10
   ...: ); import threading; import asyncio;

In [2]:

In [2]: async def connect(c):
   ...:     await c.connect()
   ...:

In [3]: async def disco(c):
   ...:     await c.disconnect()
   ...:

In [4]: L = await asyncio.gather(connect(c), disco(c))
close_session but connection wasn't established
close_secure_channel was called but connection is closed

In [5]: quit;
```

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -513,11 +513,12 @@ class Client:
         """
         Close session
         """
-        self._renew_channel_task.cancel()
-        try:
-            await self._renew_channel_task
-        except Exception:
-            _logger.exception("Error while closing secure channel loop")
+        if self._renew_channel_task:
+            self._renew_channel_task.cancel()
+            try:
+                await self._renew_channel_task
+            except Exception:
+                _logger.exception("Error while closing secure channel loop")
         return await self.uaclient.close_session(True)
 
     def get_root_node(self):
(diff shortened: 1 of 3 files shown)


ZuZuD:
Unable to repro the test failure locally, it might be sorted out once rebased on #711.

oroulet:
can you reabase and I will merge

Issue #655, opened 2021-09-01
Creating a server with OPC 40077 fails, due to importer errors


GoetzGoerisch:
**Describe the bug** <br />
While importing the nodeset [40083 GeneralTypes](https://github.com/OPCFoundation/UA-Nodeset/blob/v1.04/PlasticsRubber/GeneralTypes/1.03/Opc.Ua.PlasticsRubber.GeneralTypes.NodeSet2.xml) the importer fails with:

```
Importing companion spec. XML...
('Error no extension class registered ', 'PIDParametersDataType', NodeId(Identifier=3023, NamespaceIndex=7, NodeIdType=<NodeIdType.Numeric: 2>))
```

As per discussion with @AndreasHeine we might need to create structs on the fly, while importing.

**To Reproduce**<br />
Steps to reproduce the behavior incl code:
Implement server as per: https://github.com/umati/Sample-Server-asyncio/pull/40 
with `logging.basicConfig(level=logging.DEBUG)`

**Expected behavior**<br />
The the companion spec is imported and instanciated correctly.

**Screenshots**<br />
The PR above.

**Version**<br />
Python-Version: **3.9**<br />
opcua-asyncio Version (e.g. master branch, 0.9): **0.9.90**

KUZJurk:
For me the Error is still present when trying to import on sync or on async server.
I attached a log file to this post for the importing procedure.

__Version__

Python-Version: __3.7.9__

opcua-asyncio Version: __0.9.92_

[GeneralTypes_Errorlog.txt](https://github.com/FreeOpcUa/opcua-asyncio/files/7941129/GeneralTypes_Errorlog.txt)
_

swamper123:
This issue is still up-to-date:

```
Exception: ('Error no extension class registered ', 'PIDParametersDataType', NodeId(Identifier=3023, NamespaceIndex=4, NodeIdType=<NodeIdType.Numeric: 2>))
```

This DataType was not added to the `extension_objects_by_typeid` while importing, which .

Do you have any clue what could be the cause  @oroulet ?

schroeder-:
@swamper123 maybe this recent patch helps here: #947.

GoetzGoerisch:
Thank you @schroeder- 
Tried with the latest commit: https://github.com/umati/Sample-Server-asyncio/pull/40/commits/f8da71ec2e6e823aef4222832aee3724f59422a5

Still no luck.
```
[...]

PR #719, opened 2021-11-12
Changes to xmlimporter to load customn data types when needed

janfelixklein:
This is a fix I made concerning:

https://github.com/FreeOpcUa/python-opcua/issues/1368

It allows the xmlimporter to use customn data types that are imported in the same sequence when being used in variables. This was necessary for the I40AAS companion spec found here:

https://github.com/OPCFoundation/UA-Nodeset/tree/v1.04/I4AAS

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -355,7 +355,11 @@ class XmlImporter:
         raise Exception("Error no alias found for extension class", name)
 
     def _make_ext_obj(self, obj):
-        extclass = self._get_ext_class(obj.objname)
+        try:
+            extclass = self._get_ext_class(obj.objname)
+        except Exception:
+            await self.server.load_data_type_definitions()      # load new data type definitions since a customn class should be created
+            extclass = self._get_ext_class(obj.objname)
         args = {}
         for name, val in obj.body:
             if not isinstance(val, list):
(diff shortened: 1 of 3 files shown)


oroulet:
make sense. If it works we can merge (Looks like some tests failed)

asyncua/common/xmlparser.py
```diff
@@ -17,7 +17,10 @@ def ua_type_to_python(val, uatype_as_str):
     """
     Converts a string value to a python value according to ua_utils.
     """
-    return string_to_val(val, getattr(ua.VariantType, uatype_as_str))
+    try:
+        return string_to_val(val, getattr(ua.VariantType, uatype_as_str))
+    except:
+        return string_to_val(val, getattr(ua, uatype_as_str))
```
oroulet, line 23: better do a check than using try/catch. Exception handling is very very slow so it should not happen for normal parsing

asyncua/common/xmlimporter.py
```diff
@@ -354,8 +354,12 @@ def _get_ext_class(self, name: str):
...
         raise Exception("Error no alias found for extension class", name)
 
-    def _make_ext_obj(self, obj):
-        extclass = self._get_ext_class(obj.objname)
+    async def _make_ext_obj(self, obj):
+        try:
+            extclass = self._get_ext_class(obj.objname)
+        except Exception:
+            await self.server.load_data_type_definitions()      # load new data type definitions since a customn class should be created
+            extclass = self._get_ext_class(obj.objname)
```
oroulet, line 362: try/catch here is fine because this is a very special case. although if we could detect the issue before it would be nice

GoetzGoerisch:
relates to #655

asyncua/common/xmlimporter.py
janfelixklein: I get your point but I currently do not really see any better possibility in an earlier sequence...

PR #722, opened 2021-11-15
fix in _create_custom_type

AndreasHeine:
-fix prop name because of dataclass prop names
-add modelingrule (otherwise the props will not be instantiated)

### DIFF
--- a/asyncua/server/server.py
+++ b/asyncua/server/server.py
@@ -565,14 +565,16 @@ class Server:
             datatype = None
             if len(prop) > 2:
                 datatype = prop[2]
-            await custom_t.add_property(
-                idx, prop[0], ua.get_default_value(prop[1]), varianttype=prop[1], datatype=datatype)
+            prop_node = await custom_t.add_property(
+                idx, prop[0], ua.get_default_value(prop[1].VariantType), varianttype=prop[1].VariantType, datatype=datatype)
+            await prop_node.set_modelling_rule(True)
         for variable in variables:
             datatype = None
             if len(variable) > 2:
                 datatype = variable[2]
-            await custom_t.add_variable(
-                idx, variable[0], ua.get_default_value(variable[1]), varianttype=variable[1], datatype=datatype)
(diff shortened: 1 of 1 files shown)


AndreasHeine:
i guess we should add typing!?

AndreasHeine:
@oroulet 

would like to change the tests aswell (API-Change)!

it makes more sence to let a variant decide the initial value and have the option to predefine a default value by the user!?

asyncua/server/server.py
```diff
@@ -532,6 +532,10 @@ async def create_custom_object_type(self,
                                         properties=None,
                                         variables=None,
                                         methods=None) -> Coroutine:
+        """
+        properties: [iterable(<name>, <ua.Variant>, <datatype>), ...]
+        variables: [iterable(<name>, <ua.Variant>, <datatype>), ...]
+        """
```
oroulet, line 538: I did not even know these arguments existed. Why can't we add variable and properties in a new call by adding children to the created type node? That is much more flexible and does not take much longer to type

AndreasHeine: dont know! stumbled over it and realised that it does not work that well ^^

oroulet: maybe better flag is as depcrecated then

AndreasHeine: sorry for the delay!
flagging it as depricated will not solve the issue... if you use it you can create a type but it has no modelling rules so our instantiate method will just instantiate the objectnode!
the only thing i added is that i like to pass a list of iterable(name, Variant, Datatype) so the default value for that type can be added via the Variant!
in my opinion a nice feature and i would like to keep it ;)

oroulet: ok

oroulet: but add a small test to make sure it works. Add something in test_common..py

AndreasHeine: sure! need to make some changes on the existing tests aswell as some for that specific case!

PR #728, opened 2021-11-18
added sync client methods for getting root, server, and objects nodes

jk23909:


oroulet:
in fact I was considering not adding these methods. there are kind of deprecated. nodes are available as Client.nodes.root Client.nodes.server, etc.... see examples

jk23909:
yeah I found the conventions changed after I submitted the request. Closing now.

PR #739, opened 2021-11-25
XML import/export add more supported valuetypes

schroeder-:
Allow XMLElement, ExpandedNodeId, StatusCode and QualifiedName for xmlimport and export.


oroulet:
thanks

PR #755, opened 2021-12-15
Fix misplaced error message

MarSchif:
Parsing custom enumerations no longer logs an error in case of success. Same as in parsing of custom structures.


oroulet:
thanks. I saw that one and wondered where it was coming form

PR #761, opened 2022-01-04
Introduce type hinting into address_space.py for online discussion

ratara:
Hi everybody,
I started with issue #594 but got stuck in the code because I had to double check each call to understand parameters and whats going on. I started to add some type hinting for each discovery and at the end I was adding type hinting instead of fixing the issue.

Please let me know, if your are interested in code quality improvements. - Feel free to stop it here. But for me, things are getting much better now and this was only the start of this journey.. :-)

While swimming through the code i got some points (see added Fixme/ question comments) I would like to ask the community.

1. The access to _nodes look ugly and complicated to me. Is there a reason why we do it this way?
2. ua.StatusCode is typed as UInt32 but the constructor is called with int. Is this UInt32 data type forced by the standard? - I assume yes, I would like to think about an type hinting correct form.

Thanks!

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -508,16 +508,16 @@ class NodeManagementService:
 
     def _add_node_attr(
         self,
-        item: __TYPE_ATTRIBUTES,
+        attributes: __TYPE_ATTRIBUTES,
         nodedata: NodeData,
         name: str,
         vtype: VariantType = None,
         add_timestamps: bool = False,
         is_array: bool = False
     ):
-        if item.SpecifiedAttributes & getattr(ua.NodeAttributesMask, name):
+        if attributes.SpecifiedAttributes & getattr(ua.NodeAttributesMask, name):
             dv = ua.DataValue(
-                ua.Variant(getattr(item, name), vtype, is_array=is_array),
+                ua.Variant(getattr(attributes, name), vtype, is_array=is_array),
                 SourceTimestamp=datetime.utcnow() if add_timestamps else None,
             )
             nodedata.attributes[getattr(ua.AttributeIds, name)] = AttributeValue(dv)
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -15,7 +15,7 @@ from asyncua.ua.attribute_ids import AttributeIds
 if TYPE_CHECKING:
     from typing import Callable, Dict, List, Union
     from asyncua.ua.uaprotocol_auto import AddNodesItem
-    from asyncua.ua.uatypes import ExtensionObject, NodeId, DataValue, NumericNodeId
+    from asyncua.ua.uatypes import ExtensionObject, NodeId, DataValue
     from asyncua.ua.uaprotocol_auto import (
         BrowsePath, BrowseParameters, RelativePathElement, BrowseResult,
         BrowseDescription, ReadParameters, WriteParameters, ReferenceDescription
@@ -153,14 +153,14 @@ class ViewService(object):
             oktypes.remove(ua.NodeId(ua.ObjectIds.HasSubtype))
         return ref2 in oktypes
 
-    def _get_sub_ref(self, ref: NodeId):
-        res = []
+    def _get_sub_ref(self, ref: NodeId) -> List[NodeId]:
+        res: List[NodeId] = []
         nodedata = self._aspace[ref]
         if nodedata is not None:
-            for ref in nodedata.references:
-                if ref.ReferenceTypeId.Identifier == ua.ObjectIds.HasSubtype and ref.IsForward:
-                    res.append(ref.NodeId)
-                    res += self._get_sub_ref(ref.NodeId)
+            for ref_desc in nodedata.references:
+                if ref_desc.ReferenceTypeId.Identifier == ua.ObjectIds.HasSubtype and ref_desc.IsForward:
+                    res.append(ref_desc.NodeId)
+                    res += self._get_sub_ref(ref_desc.NodeId)
         return res
 
     def _suitable_direction(self, direction: ua.BrowseDirection, isforward: bool) -> bool:
@@ -577,28 +577,28 @@ class AddressSpace:
 
     def __init__(self):
         self.logger = logging.getLogger(__name__)
-        self._nodes: Dict[NumericNodeId, NodeData] = {}
+        self._nodes: Dict[NodeId, NodeData] = {}
         self._datachange_callback_counter = 200
         self._handle_to_attribute_map = {}
         self._default_idx = 2
         self._nodeid_counter = {0: 20000, 1: 2000}
 
-    def __getitem__(self, nodeid: NumericNodeId) -> NodeData:
+    def __getitem__(self, nodeid: NodeId) -> NodeData:
         return self._nodes.__getitem__(nodeid)
 
-    def get(self, nodeid: NumericNodeId) -> Union[NodeData, None]:
+    def get(self, nodeid: NodeId) -> Union[NodeData, None]:
         return self._nodes.get(nodeid, None) # Fixme This is another behaviour than __getitem__ where an exception is thrown, right?
 
-    def __setitem__(self, nodeid: NumericNodeId, value: NodeData):
+    def __setitem__(self, nodeid: NodeId, value: NodeData):
         return self._nodes.__setitem__(nodeid, value)
 
-    def __contains__(self, nodeid: NumericNodeId) -> bool:
+    def __contains__(self, nodeid: NodeId) -> bool:
         return self._nodes.__contains__(nodeid)
 
-    def __delitem__(self, nodeid: NumericNodeId):
+    def __delitem__(self, nodeid: NodeId):
         self._nodes.__delitem__(nodeid)
 
-    def generate_nodeid(self, idx: Union[int, None] = None):
+    def generate_nodeid(self, idx: Union[int, None] = None) -> NodeId:
         if idx is None:
             idx = self._default_idx
         if idx in self._nodeid_counter:
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -11,14 +11,15 @@ from functools import partial
 from typing import TYPE_CHECKING
 
 from asyncua.ua.attribute_ids import AttributeIds
+from asyncua.ua.uaprotocol_auto import Index
 
 if TYPE_CHECKING:
     from typing import Callable, Dict, List, Union
-    from asyncua.ua.uaprotocol_auto import AddNodesItem
     from asyncua.ua.uatypes import ExtensionObject, NodeId, DataValue
     from asyncua.ua.uaprotocol_auto import (
-        BrowsePath, BrowseParameters, RelativePathElement, BrowseResult,
-        BrowseDescription, ReadParameters, WriteParameters, ReferenceDescription
+        AddNodesItem, BrowsePath, BrowseParameters, RelativePathElement, BrowseResult,
+        BrowseDescription, ReadParameters, WriteParameters, ReferenceDescription,
+        BrowsePathResult, DeleteNodesParameters
     )
 
 from asyncua import ua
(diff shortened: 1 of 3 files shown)


asyncua/server/address_space.py
```diff
@@ -1,22 +1,36 @@
+from __future__ import annotations
```
oroulet, line 1: are you sure you need that one?

usually I just write `from typing import Dict, tuple, etc....`

asyncua/server/address_space.py
```diff
@@ -39,18 +54,19 @@ def __str__(self):
 
 
 class AttributeService:
-    def __init__(self, aspace: "AddressSpace"):
+    # FIXME: What is the purpose of this class?
```
oroulet, line 57: what is that comment? that class implement "AttrbuteService" from the specification. Hard to make it more clear

asyncua/server/address_space.py
oroulet: or does it do something else?

asyncua/server/address_space.py
```diff
@@ -39,18 +54,19 @@ def __str__(self):
...
+    def __init__(self, aspace: AddressSpace):
         self.logger = logging.getLogger(__name__)
-        self._aspace: "AddressSpace" = aspace
+        self._aspace: AddressSpace = aspace
 
-    def read(self, params):
+    def read(self, params) -> List[DataValue]:
```
oroulet, line 62: would be nice to add typing to params. I started in some classes,

asyncua/server/address_space.py
```diff
@@ -39,18 +54,19 @@ def __str__(self):
...
         # self.logger.debug("read %s", params)
-        res = []
+        res: List[DataValue] = []
         for readvalue in params.NodesToRead:
             res.append(self._aspace.read_attribute_value(readvalue.NodeId, readvalue.AttributeId))
         return res
 
-    async def write(self, params, user=User(role=UserRole.Admin)):
+    async def write(self, params, user: User = User(role=UserRole.Admin)):
```
oroulet, line 69: in that case type of user was obvious but not `params`

asyncua/server/address_space.py
```diff
@@ -74,11 +90,12 @@ async def write(self, params, user=User(role=UserRole.Admin)):
 
 
 class ViewService(object):
-    def __init__(self, aspace: "AddressSpace"):
+    # FIXME: What is the purpose of this class?
```
oroulet, line 93: same as over

asyncua/server/address_space.py
```diff
@@ -554,7 +571,7 @@ def __init__(self):
         self._default_idx = 2
         self._nodeid_counter = {0: 20000, 1: 2000}
 
-    def __getitem__(self, nodeid):
+    def __getitem__(self, nodeid): # question: Why do we use this 'fancy' python stuff? - It looks pretty complicated and is confusing for me.
```
oroulet, line 574: this is rally basic python object method. No it is not complicated. that is the same in most object oriented language

asyncua/server/address_space.py
ratara: I added annotations because someone added stringified type hints to resolve forward declarations for types.This workaround is fixed by annotations automatically and will be standard somewhere in the future.

asyncua/server/address_space.py
```diff
@@ -598,7 +615,7 @@ def generate_nodeid(self, idx=None):
     def keys(self):
         return self._nodes.keys()
 
-    def empty(self):
+    def empty(self): # FIXME: missleading name empty, better clear or reset, but could be braking change
```
oroulet, line 618: good point

oroulet:
typing is welcome. please continue your work, but try to not do too many changes at once. work a few files at a time

oroulet:
> The access to _nodes look ugly and complicated to me. Is there a reason why we do it this way?
starting with undersore because it is private and protected by a lock. Or has the lock been removed? We might still have some threads around, I do not remember

asyncua/server/address_space.py
ratara: A link to that specification would be nice then or a short summary.

asyncua/server/address_space.py
oroulet: OK. I saw it now. then maybe we should protect that line with some " if python version < 3.xx: from future import annotations" ?

asyncua/server/address_space.py
oroulet: you are welcome. Add a link

asyncua/server/address_space.py
oroulet: would be nice to rename. simply to a rep in code to see where it is called. I am guessing it is never called and could even be removed

asyncua/server/address_space.py
ratara: I take a look at best practices for futures.. This can't be a new topic.

asyncua/server/address_space.py
ratara: The type was obvious to us but not to the linter. 🙈
I would try to introduce the concept "lean on the linter".🤗 It helps a lot before refactorings or new features.

asyncua/server/address_space.py
ratara: Done

asyncua/server/address_space.py
ratara: Never mind. It is just my personal opinion. I was surprised by this "magic method" and that is in most cases a bad sign. A simple method to access private data would do the same job but would be easier to understand for all non-pythonic developers.
I will type hint this, this should provide enough hint for others.

ratara: Now i discovered a duplicated method get(). This method is exactly what i meant.
I have to think about that. I wont change anything in this branch, just type hinting and maybe very simple rafactorings.
Trying to stay focused.:-)

ratara:
@oroulet 
Bildschirmfoto vom 2022-01-05 21-55-59
Type hinting reveals a problem here. BrowsePathResult claims that its attribute TargetId is of type ExpandedNodeId. In fact the parent class is assigned here which looks correct. In reality (unit tests) NodeId instances and NumericNodeId instances are assigned.
I tend to change the type hinting of BrowsePathResult::TargetId to NodeId. I need some acknowledgement about this because of the following assignment in BrowsePathResult, i do not fully understand:
grafik
Although the default is ExpandedNodeId, it is in reality NodeId or any child. - Correct?

asyncua/server/address_space.py
ratara: Done

oroulet:
uaprotocol_auto.py is generated from the specification, so an error there would be surprising (they have error though).
In current code NodeId and ExpandedNodeId are exchanged almost randomly. write a FIXME: when you see cases with potential errors. NumericNodeId is a NodeId so that one i correct

asyncua/server/address_space.py
```diff
@@ -1,58 +1,84 @@
...
+    def __str__(self) -> str:
         return f"AttributeValue({self.value})"
 
     __repr__ = __str__
 
 
 class NodeData:
-    def __init__(self, nodeid):
+    """
+    This class implements a node in the address space.
```
oroulet, line 49: better -> The class is internal to asyncua and holds all the information about a Node

asyncua/server/address_space.py
```diff
@@ -1,58 +1,84 @@
...
+
 from .users import User, UserRole
 
 _logger = logging.getLogger(__name__)
 
 
 class AttributeValue(object):
-    def __init__(self, value):
+    """
+    This class implements the value of an attribute.
```
oroulet, line 34: The class holds the value(s) of an attribute and callbacks

ratara:
> uaprotocol_auto.py is generated from the specification, so an error there would be surprising (they have error though). In current code NodeId and ExpandedNodeId are exchanged almost randomly. write a FIXME: when you see cases with potential errors. NumericNodeId is a NodeId so that one i correct

I did that in the last commits. I guess using the parent class (NodeId) as type hint is the right thing to do but I will check the link to the specification for this. If it is unclear, I leave it for now with the FIXME comment.

asyncua/server/address_space.py
```diff
@@ -455,45 +506,62 @@ def _delete_reference(self, item, user):
             self._delete_unique_reference(item, True)
         return self._delete_unique_reference(item)
 
-    def _add_node_attr(self, item, nodedata, name, vtype=None, add_timestamps=False, is_array=False):
+    def _add_node_attr(
+        self,
+        item: __TYPE_ATTRIBUTES,
```
oroulet, line 511: you could also change the name`item`here. it seems quite wrong

asyncua/server/address_space.py
ratara: annotations is supports with Python 3.7. This code project does not support python version prior to 3.7. Therefor we have no backward compatibility issues.
Annotation will be mandantory with 3.11, see https://docs.python.org/3/library/__future__.html
They wont delete the feature from __future__ so it should not brake anything.
I recommend to keep the import simple and to remove it, if this project minimum version gets 3.11.

asyncua/server/address_space.py
```diff
@@ -1,58 +1,94 @@
...
-import logging
-import collections
-from datetime import datetime
 from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime
 from functools import partial
+from typing import TYPE_CHECKING
+
+from asyncua.ua.attribute_ids import AttributeIds
+from asyncua.ua.uaprotocol_auto import Index
```
oroulet, line 14: you should import from ua_protocol.py and not uaprotocol_auto. maybe uaprotocol_auto.py should be rename to `_uaprotocol_auto.py`

asyncua/server/address_space.py
```diff
@@ -1,58 +1,94 @@
...
+
+if TYPE_CHECKING:
+    from typing import Callable, Dict, List, Union
+    from asyncua.ua.uatypes import NodeId, DataValue, VariantType
+    from asyncua.ua.uaprotocol_auto import (
```
oroulet, line 19: same here. import from uaprotocol simply

asyncua/server/address_space.py
```diff
@@ -1,58 +1,94 @@
...
+        VariableTypeAttributes, VariableAttributes, ObjectTypeAttributes
+    )
+    __TYPE_ATTRIBUTES = Union[    
+        DataTypeAttributes,
+        ReferenceTypeAttributes,
+        VariableTypeAttributes,
+        VariableAttributes,
+        ObjectTypeAttributes,
+        ObjectAttributes
+    ] # FIXME Check, if there are missing attribute types. Another way to solves this would a base class for all attribute classes in uaprotocol_auto.py
```
oroulet, line 33: that is possible. or define that one in uaprotocol_hand.py

asyncua/server/address_space.py
ratara: I renamed it, no unit test broke. Maybe users use this method. Their code has to be changed.
I also replaced the assignment of an empty dictionary with the more precise method "clear()" of dict.

asyncua/server/address_space.py
ratara: I'll do.

asyncua/server/address_space.py
ratara: To be honest, my IDE chose that import.  I will fix that.

asyncua/server/address_space.py
ratara: Done

asyncua/server/address_space.py
ratara: @oroulet There is no file ua_protocol.py or uaprotocol.py. Can u give me a hint?

asyncua/server/address_space.py
ratara: I chose the type alias because i do not want to change any functionality. Introducing a whole new class hierarchy would be a great change compared to the listing of used types.
I will check at the end, if i can find a list of attributes but even if this list is incomplete, it is better than before. The next person sees a type conflict, if he calls the method with another attribute.

asyncua/server/address_space.py
ratara: I think i git it now. U meant the namespace ua

ratara:
I think i am done.

Next task would be to reduce the type hinting errors but this involves some refactoring. Therefor i suggest to do it in another branch afterwards.
Most type problems occur by assigning standrad python data types, like bool, to opc ua data types, like Boolean. There is a border between opc ua spec and python data types. To solve this, we need some discussion first.

Please merge, if you like the changes.

PR #764, opened 2022-01-12
Add UML for address space to prepare future tasks

ratara:
Note: I had to drop the original PR because I had to rebase my branch with master first and got into some trouble. :-)

**Content**

After introducing type hinting to address space, I discovered some data type issues. I would like to introduce some design documentation about address space to enable a discussion how to solve/refactor the issues.

Step 1: Add UML overview about address space ( in progress )
Step 2: Lets discuss what we can do on basis of these UML diagrams.
Step 3: Merge UML overview into master
Step 4: add new branch and resolve the discussed issues

**Plant UML:**
I thought about how to document and although I work a lot with Enterprise Architect, this is not a good solution here. I chose Plant UML because:
- It is a known free vscode extension that can be run without any license.
- It works offline.
- Changes to the documentation can be tracked in git because the UML is generated from a text file and we have no binary blob.

**Open questions:**
- I am open for ideas where to store the UML content. My first idea is side-by-side with the code but this can confuse. Maybe a sub folder is better?

Any thoughts?

### DIFF
--- a/asyncua/server/address_sapce.puml
+++ b/asyncua/server/address_sapce.puml
@@ -1,0 +1,141 @@
+@startuml Class diagram address_space
+NodeId <|-- ExpandedNodeId
+NodeId <|-- StringNodeId
+NodeId <|-- GuidNodeId
+NodeId <|-- ByteStringNodeId
+NodeId <|-- NumericNodeId
+NodeId <|-- FourByteNodeId
+NodeId <|-- TwoByteNodeId
+
+NodeId o-- NodeIdType
+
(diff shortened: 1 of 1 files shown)


ratara:
This is a snapshot of the current generated UML diagram. I post it to visualise the possible benefit. Please do not review the diagram quality. I am creating the big picture now and will split the information afterwards in smaller pieces.

grafik

oroulet:
The problem with UML documentation is that it can get very quickly outdated... but that software has been realitvely stable for 10 years so maybe it is fine

PR #765, opened 2022-01-12
fix hardcoded field serializer for Encoding field

moongladestorm:
The serializer for the Encoding field is hardcoded at the moment to "Byte". This prevented the use of Optionals and Unions as the OPCUA v104 specification expects it to be UInt32 in both cases, even tough the field type of the Encoding field was set correctly to UInt32 (see https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.7/ and https://reference.opcfoundation.org/v104/Core/docs/Part6/5.2.8/). This can simply be fixed by using the field serializer also for the encoding field and should still work correctly for Encoding field with type "Byte".

### DIFF
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -277,18 +277,18 @@ def create_dataclass_serializer(dataclazz):
         in enumerate(filter(lambda f: type_is_union(f.type), data_fields))
     ]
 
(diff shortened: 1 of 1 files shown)


oroulet:
tests seems to pass so that should be correct sic eyou only modified the serializer

PR #771, opened 2022-01-16
Build state machines from type definition

SanobLuj:
To simplify the creation of a state machine I added an initial implementation to build a state machine based on its type definition. Currently an existing node of a state machine type needs to be passed to the constructor. The `StateMachineClass` will then browse the type definition of this node to add the states, transitions, and causes defined by the type. 

For executing a state, each `State` object provides an `on_entry` and `on_exit` method. These are executed on each state change for the current state and the next one. These methods can be overwritten to execute a certain behaviour.

Base classes for the `ProgramStateMachine`, `FileTransferStateMachine`, and `ExclusiveLimitStateMachine` have been added. However, none of these are fully implemented. The `ProgramStateMachine` is missing some functionality as it is described by the specification but can be used in a simple form (see example). About the `FileTransferStateMachine` and the `ExclusiveLimitStateMachine` I'm not sure how these are intenden to be used. At least there are some references missing, I guess coming from the `TemporaryFileTransferType` and a `ConditoinAlarmType`. 

In addition to the state machine a `ParameterSet` implementation has been added. A parameter set is a flat list of parameters/variables. These are used e.g. within the [device specification](https://reference.opcfoundation.org/v104/DI/v102/docs/5.3/) to organize the parameters of a device. Optionally the values of the parameters in the set can be subscribed. 

There is the example [`server_typed_state_machine.py`](https://github.com/SanobLuj/opcua-asyncio/blob/typed_state_machines/examples/server_state_machine_type.py) using a  [custom nodeset](https://github.com/SanobLuj/opcua-asyncio/blob/typed_state_machines/examples/StateMachine.Example.NodeSet2.xml) with several predefined state machines and parameter sets. 

grafik

There is still some more work needed. However, would be glad if you guys can tell me what you think.

### DIFF
--- a/asyncua/common/parameter_set.py
+++ b/asyncua/common/parameter_set.py
@@ -1,0 +1,153 @@
+from asyncore import dispatcher
+from asyncua import Node, Server, ua
+from asyncua.common.callback import CallbackType, CallbackService, ServerItemCallback
+
+class ParameterSet:
+    """
+    Parmeter set for easy access of parameters
+ 
+    A parameter set is an object only containing variables. 
+    This class can be used for e.g. devices or state machines.
+
+    @param node: the node to the object representing the parameter set
+    @param subscribe: subscribe to value changes of the parameters
+    @param notifier: method to be called when a value in the parameter set changes 
+    @param source: server or client object as source of the information
+    """
+    def __init__(self, node : Node, subscribe=False, notifier=None, source=None, interval=100):
+        self._parameters = {}
+        self._node = node
+        self._source = source
+        self.name = ''
+        self._notify_data_change = notifier
+        self._subscribe_data_change = subscribe
+        self._parameter_nodes = []
+        self._subscribe = subscribe
+        self._parameter_ids = []
+        self._subscription_interval = interval
+        self._subscription = None
+
+    async def init(self):
+        # Get the ParameterSet name
+        bn = await self._node.read_browse_name()
+        self.name = bn.Name
+
+        # Browse ParameterSet object
+        parameters = await self._node.get_children(refs=33)
+        for p in parameters:
+            bn = await p.read_browse_name()
+            val = await p.read_value()
+            # Add the parameter to parameter dictionary 
+            self._parameters[bn.Name] = {'Name': bn.Name, 'Default': val, 'Node': p, 'Value': val} # TODO: add unit and range information 
+            self._parameter_nodes.append(p)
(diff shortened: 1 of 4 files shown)


oroulet:
I have no ideas what that stuff is about. @AndreasHeine  isn'it the kind of things you are working on?

asyncua/common/parameter_set.py
```diff
@@ -0,0 +1,153 @@
...
+            typedefinition = await node.read_type_definition()
+            if value != None: 
+                s = '|{:^30}|{:^20}|'.format(name, value)
+            else: 
+                nan = 'nan'
+                s = f'|{name:^30}|{nan:^20}|'
+            s += '\n-----------------------------------------------------'
+            print(s)
+
+        print()
```
AndreasHeine, line 153: ?

AndreasHeine:
@oroulet sure i implemented them as far as i needed them ... the rest i left open for reference! @SanobLuj is probably the first one using it else then me ^^

on the first look it appearance fine, the only thing i am not sure about is that internal/serverside subscription

SanobLuj:
The original intention was to use this for implementing skills. Here are a few references describing the skill concept: 
- [Plattform I40 - Describing Capabilities of I40 Components](https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Capabilities_Industrie40_Components.pdf?__blob=publicationFile&v=3)
- [OPC UA skills  based on the program type using open62541](https://github.com/opcua-skills)
- [SkillUp framework with skills based on the PackML state machine using milo](https://github.com/aljoshakoecher/skill-up)

However, I thought it could be helpful for implementing all kinds of state machines that are defined by its type. I will make additional changes and give some more explanation soon.  

@AndreasHeine  the parameter set class could be used with a server or a client (`source`). That is why I have used the subscription there. Probably not the best idea to be used with the server  ^^

asyncua/common/parameter_set.py
SanobLuj: Better not? It's just for prettier prtinting ^^

AndreasHeine: f-string would be better!

AndreasHeine:
@oroulet after a while and recap what i did ^^ the FiniteStatmachineType is abstract... so it should not be possible to instantiate it only subtypes... maybe we need to add a check for that in instantiate-method!? some refactoring is required as well...

Issue #774, opened 2022-01-19
issue: Custom Structures fail with import_xml


kozachynskyi:
**Describe the bug** <br />
If user creates custom structures with structure104 and one of fields has dimensions, than it works as expected. However, if one exports XML and starts a new server that imports that custom structure, client cannot write lists and has following error:

```
File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 288, in list_to_binary
    pack = [to_binary(uatype, el) for el in val]
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 288, in <listcomp>
    pack = [to_binary(uatype, el) for el in val]
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 277, in to_binary
    return struct_to_binary(val)
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 253, in struct_to_binary
    packet.append(to_binary(uatype, val))
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 271, in to_binary
    return pack_uatype(vtype, val)
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 177, in pack_uatype
    return extensionobject_to_binary(value)
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 457, in extensionobject_to_binary
    body = struct_to_binary(obj)
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 253, in struct_to_binary
    packet.append(to_binary(uatype, val))
  File "/home/ukrman/.pyenv/versions/3.9.5/envs/optipal/lib/python3.9/site-packages/asyncua/ua/ua_binary.py", line 271, in to_binary
    return pack_uatype(vtype, val)
[...]

PR #775, opened 2022-01-19
fix: ArrayDimensions of Structures xml_importer

kozachynskyi:
Closes: https://github.com/FreeOpcUa/opcua-asyncio/issues/774

I think that after importing of XML with custom structures, ArrayDimension attribute is set as a String, rather than List.Thus, in the issue 774, server writes `1 is not a instance of "list"!` (1 here is ArrayDimension), and send DataType Node to client with None attribute of ArrayDimension. Thus client loads `await client.load_data_type_definitions() `  without any infromation about Dimension.

This is my fix to this problem, not 100% sure though. I ran pytest on version 0.9.92 with and without a fix, and it yelled same result.

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -596,7 +596,7 @@ class XmlImporter:
             f.IsOptional = field.optional
             if f.IsOptional:
                 optional = True
-            if field.arraydim == "" or field.arraydim is None:
+            if field.arraydim is None:
                 f.ArrayDimensions = field.arraydim
             else:
                 f.ArrayDimensions = [int(i) for i in field.arraydim.split(",")]
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -596,7 +596,10 @@ class XmlImporter:
             f.IsOptional = field.optional
             if f.IsOptional:
                 optional = True
-            f.ArrayDimensions = field.arraydim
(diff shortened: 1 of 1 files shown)


asyncua/common/xmlimporter.py
```diff
@@ -596,7 +596,10 @@ def _get_sdef(self, obj):
             f.IsOptional = field.optional
             if f.IsOptional:
                 optional = True
-            f.ArrayDimensions = field.arraydim
+            if field.arraydim == "" or field.arraydim is None:
```
oroulet, line 599: looks like the type of ArayDimensions is int32 (so also List[int] since all variables can be list in ua.
So you should not allow `""` as a possible value. Replace it with None.
Can you fix and I will merge

kozachynskyi:
Final commit is not what you asked for, but it does a following:

opcua exporter exports any structure field by using https://github.com/FreeOpcUa/opcua-asyncio/blob/bbb7155040263131e47dce7456b6420f8264c663/asyncua/common/xmlexporter.py#L234

 thus XML importer will now just do the opposite operation. However if the operation is not possible, it will proceed by doing what it did before. Taking into account, that no one so far had issues with it  before, sound like a good solution. What you think @oroulet ?

oroulet:
I prefered the original the version. your latest commit will only delay the detection of the error. 
If ArrayDimension is set to wrong value we should crash as soon as possible, not forward the problem

oroulet:
I merged the two first commits

PR #782, opened 2022-01-31
limit size of not-acknowledged results

SamLyuubc:
I experienced a memory leak [issue](https://github.com/FreeOpcUa/opcua-asyncio/discussions/750) with the library earlier, when there  are too many not acknowledged publish results stored.

I think it would make sense to limit the number of not acknowledged result we stored.

### DIFF
--- a/asyncua/server/internal_subscription.py
+++ b/asyncua/server/internal_subscription.py
@@ -10,7 +10,6 @@ from asyncua import ua
 from .monitored_item_service import MonitoredItemService
 from .address_space import AddressSpace
 
-
 class InternalSubscription:
     """
     Server internal subscription.
@@ -34,6 +33,7 @@ class InternalSubscription:
         self._triggered_events: Dict[int, List[ua.EventFieldList]] = {}
         self._triggered_statuschanges: list = []
         self._notification_seq = 1
+        self._no_acks_limit = 500
         self._not_acknowledged_results: Dict[int, ua.PublishResult] = {}
         self._startup = True
         self._keep_alive_count = 0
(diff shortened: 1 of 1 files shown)


asyncua/server/internal_subscription.py
```diff
@@ -5,11 +5,30 @@
...
+        self._check_size_limit()
+
+    def __setitem__(self, key, value):
+        OrderedDict.__setitem__(self, key, value)
+        self._check_size_limit()
+
+    def _check_size_limit(self):
+        if self.size_limit is not None:
+            while len(self) > self.size_limit:
+                self.popitem(last=False)
```
oroulet, line 31: that seems quite a lot of lines, instead of 2 lines: definine a class member: `self._no_acks_limit = xxx`and then `while len(self.xxxx): self.xxx.pop_item()` when that dict is getting updated

SamLyuubc: That would also work. I was thinking of a more reusable class in the first place. I will make the change.

asyncua/server/internal_subscription.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, data: ua.CreateSubscriptionResult, aspace: AddressSpace,
         self._triggered_events: Dict[int, List[ua.EventFieldList]] = {}
         self._triggered_statuschanges: list = []
         self._notification_seq = 1
-        self._not_acknowledged_results: Dict[int, ua.PublishResult] = {}
+        self._no_acks_limit = 500
+        self._not_acknowledged_results: OrderedDict[int, ua.PublishResult] = OrderedDict()
```
oroulet, line 38: As far as I know dict are ordered in C Python. Was there another reason to use OrderedDict?

SamLyuubc: I was not aware that dict are sorted now. You are correct, but for python < 3.7, dictionary are not ordered..  Should we also consider users with lower version of python? If not I will change it back.

PR #784, opened 2022-02-02
clean up NodeIde String when importing custom structs

oroulet:


asyncua/common/structures104.py
```diff
@@ -271,6 +271,9 @@ def __str__(self):
 
 async def _recursive_parse(server, base_node, dtypes, parent_sdef=None, add_existing=False):
     for desc in await base_node.get_children_descriptions(refs=ua.ObjectIds.HasSubtype):
+        if desc.NodeId.NodeIdType == ua.NodeIdType.String:
+            # Some PLC create names not compatible with Python syntax
+            desc.NodeId.Identifier = clean_name(desc.NodeId.Identifier)
```
oroulet, line 276: does not work that node is readonly. need to replace it

oroulet:
OK. this is not a good idea. we want to acept string with quote or space as NodeId

oroulet:
one more attempt

oroulet:
but still probably not a solution...

oroulet:
last commit might be correct, but I have no ideas how to test it

swamper123:
> last commit might be correct, but I have no ideas how to test it

Make 2 tests and throw some strings into it. (e.g. [via @pytest.mark.parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html)). 
One with a good stings (which shouldn't be changed) and other strings with "bad content", which should be changed. re.match() may do most of the job then.

tdpu:
Thanks, @oroulet we will test this with our Siemens PLC and let you know how it works.

oroulet:
@tdpu did you test that?

oroulet:
dos not work. solved differently in anther PR

PR #806, opened 2022-02-16
Fixed Parametername varianttype -> VariantType

jrast:
Fixed parameter named from lowercase to CameCase in statemachine.install.

### DIFF
--- a/asyncua/common/statemachine.py
+++ b/asyncua/common/statemachine.py
@@ -127,7 +127,7 @@ class StateMachine(object):
                 self._last_transition_transitiontime_node = await self._last_transition_node.add_property(
                     0, 
                     "TransitionTime", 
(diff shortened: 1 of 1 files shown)


jrast:
Note: This fix would also be applied if #771 would be merged. See https://github.com/FreeOpcUa/opcua-asyncio/pull/771/files#diff-f005748099837bd8fb9db1b8adc0c32b4dda5614b1505d6f7e7b686c60fcd11dL130

AndreasHeine:
well spotted! another leftover from switching to DataClasses ;)
https://github.com/FreeOpcUa/opcua-asyncio/blob/033427b6a5bdf1a4a512e4976a410b6cc6291406/asyncua/ua/uatypes.py#L771

PR #825, opened 2022-02-26
support unions datatypes

schroeder-:
Support for unions. 
Add xmlimport/export tests for structs with optional fields.

generation via struct104:

```
await new_struct(server, idx, "MyUnionStruct", [
        new_struct_field("MyString", ua.VariantType.String),
        new_struct_field("MyInt64", ua.VariantType.Int64),
    ], is_union=True)

```
example of the generate code:

```
@dataclass
class MyUnionStruct(ua.UaUnion):

    '''
    MyUnionStruct structure autogenerated from StructureDefinition object
    '''

    data_type = ua.NodeId.from_string("ns=4;i=1")

    Encoding: ua.Byte = field(default=0, repr=False, init=False, compare=False)    
    Value: Union[None, ua.String,ua.Int64] = field(default=None, init=False)       

    @property
    def MyString(self) -> Optional[ua.String]:
        if self.Encoding == 1:
            return self.Value
        return None

    @MyString.setter
    def MyString(self, value: ua.String) -> None:
        self.Value = value
        self.Encoding = 1



    @property
    def MyInt64(self) -> Optional[ua.Int64]:
        if self.Encoding == 2:
            return self.Value
        return None

    @MyInt64.setter
    def MyInt64(self, value: ua.Int64) -> None:
        self.Value = value
        self.Encoding = 2
```

usage:

```
my_union = ua.MyUnionStruct()
my_union.MyInt64 = 555
print(my_union.MyInt64)
```


oroulet:
I did not even read the spec about Union but aren't union supposed to return the same value as two or several type? Here is looks like an Union is just a storage type containing either one type or the other. But maybe this is what ua unions are?

oroulet:
anyway looks like you know what you are doing, coe looks OK  and all tests pass so I am merging

schroeder-:
Unions are either Null or one of the fields. They are classic C-Style unions.  Here is the c repr.

```
struct MyUnionStruct{
     uint8_t Encoding;
     union{
         char MyString[32];
         int64_t MyInt64;
      } Value;
};
```

Will add xmlimport and export next.

oroulet:
you are right.

PR #826, opened 2022-02-27
basic pubsub

schroeder-:
**First of this pr is just a preview, to gather feedback about the api design. Pubsub is really complex and a lot of things are not implemented or handled correctly.**

Points of interests are [examples/pubsub/publisher_simple.py](https://github.com/FreeOpcUa/opcua-asyncio/compare/master...schroeder-:basic_pubsub?expand=1#diff-d33e0f82b40fea6895ec27c51e19758ee83a61c3b737eaf4ddc726a777c7e01e) and [examples/pubsub/publisher_simple.py](https://github.com/FreeOpcUa/opcua-asyncio/compare/master...schroeder-:basic_pubsub?expand=1#diff-98736bfc9f6baeea47ebb6abe6e6bd74bfb080eecec5e6cb058f88731cac5fc6).
Running both examples creates an publishing server on port 4840 with a folder "PublisherDemoNodes" with variables. This variables are send via uadp-udp, to the subscriber. On port 4841 the subscriber server has a folder "PublisherDemoNodes" where the variables revived via pubsub are mirrored. You can now change the variables on the publisher (Port 4840) and the values will be mirrored on the subscriber.

# TODO
- add a lot more test after api is finalized.
- add tests against open62541 for interop testing.
- refresh information model.
- provide more methods to modify pubsub elements.

# implemented features
## Messagetypes:
- [x] Uadp (Binary)
- [ ] Json
## Transportlayer
- [x] UDP
- [ ] Ethernet
- [ ] Mqtt
- [ ] Ampq
## Informationmodel
- [x] readonly Informationmodel
- [ ] import/export via UABinaryFile
- [ ] add methods to modify informationmodel

### DIFF
--- a/examples/pubsub/subscriber_standalone.py
+++ b/examples/pubsub/subscriber_standalone.py
@@ -1,0 +1,88 @@
+"""
+Example creating a standalone subscriber that recives an Int32, String, Bool and ArrayInt16 matching the pubsubpublisher standalone example
+"""
+import asyncio
+import logging
+from typing import List
+from asyncua import ua, pubsub
+from asyncua.pubsub.udp import UdpSettings
+from dataclasses import dataclass
+
+# This Parameter must match the Publisher Settings!
+@dataclass
+class PubSubCFG:
+    PublisherID: ua.Variant = ua.Variant(ua.UInt16(1))
+    WriterId: ua.UInt16 = ua.UInt16(1)
+    DataSetWriterId: ua.UInt16 = ua.UInt16(32)
+    Url: str = "opc.udp://239.0.0.1:4840"
+
+
+CFG = PubSubCFG()
+
+
+def create_meta_data():
+    # This is the description of the incoming data
+    dataset = pubsub.DataSetMeta.Create("Simple")
+    dataset.add_field(pubsub.DataSetField.CreateScalar("Int32", ua.VariantType.Int32))
+    dataset.add_field(pubsub.DataSetField.CreateScalar("String", ua.VariantType.String))
+    dataset.add_field(pubsub.DataSetField.CreateScalar("Bool", ua.ObjectIds.Boolean))
+    dataset.add_field(
+        pubsub.DataSetField.CreateArray("ArrayInt16", ua.ObjectIds.Double)
+    )
+    return dataset
+
+
+class OnDataRecived:
+    """
+    This is called when a dataset is recived
+    """
+
+    async def on_dataset_recived(
+        self, _meta: pubsub.DataSetMeta, fields: List[pubsub.DataSetValue]
+    ):
+        print("Got Msg:")
+        for f in fields:
+            print(f"{f.Name} -> {f.Value}")
+
+    async def on_state_change(self, meta: pubsub.DataSetMeta, state: ua.PubSubState):
+        """Called when a DataSet state changes"""
+        print(f"State changed {meta.Name} - {state.name}")
+
+
+async def main():
+    handler = OnDataRecived()
+    logging.basicConfig(level=logging.INFO)
+    metadata = create_meta_data()
+    # Configure ReaderGroup and DataSetReader this must match the WriterGroup/DataSetWriter configured on the publisher
+    reader = pubsub.ReaderGroup.new(
+        name="ReaderGroup1",
+        enable=True,
+        reader=[
+            pubsub.DataSetReader.new(
+                CFG.PublisherID,
+                CFG.WriterId,
+                CFG.DataSetWriterId,
+                metadata,
+                name="SimpleDataSetReader",
+                subscriped=handler,
+                enabled=True,
+            )
+        ],
+    )
+    # Create a connection
+    # The PublisherId here is 2 but could be any number because we are just subscribing
+    connection = pubsub.PubSubConnection.udp_udadp(
+        "Subscriber Connection1 UDP UADP",
+        ua.UInt16(2),
+        UdpSettings(Url=CFG.Url),
+        reader_groups=[reader],
+    )
+    ps = pubsub.PubSub.new(connections=[connection])
+    # Run the connection
+    async with ps:
+        while 1:
+            await asyncio.sleep(5)
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ examples/history.db
 /.settings/
 /venv/
 .eggs*
+.vscode/settings.json
+.coverage
--- a/asyncua/pubsub/__init__.py
+++ b/asyncua/pubsub/__init__.py
@@ -1,0 +1,10 @@
+
+from .dataset import *
+from .writer import *
+from .connection import *
+from .pubsub import *
+from .connection import *
+from .writer import *
+from .reader import *
+from .udp import UdpSettings
+from .subscriped_dataset import *
--- a/asyncua/pubsub/connection.py
+++ b/asyncua/pubsub/connection.py
@@ -1,0 +1,298 @@
+'''
+    Connection which sends/generates and recives/handles pubsub message
+    over the network
+"""
+from __future__ import annotations
+from typing import List, Optional, Union, TYPE_CHECKING
+from asyncua.ua.status_codes import StatusCodes
+from asyncua.ua.uaprotocol_auto import ReaderGroupDataType, WriterGroupDataType
+from ..server.server import Server
+from ..pubsub.information_model import PubSubInformationModel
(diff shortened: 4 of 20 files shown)


examples/pubsub/subscriber_standalone.py
```diff
@@ -0,0 +1,74 @@
...
+}
+
+
+def create_meta_data():
+    # This is the description of the incoming data
+    dataset = pubsub.DataSetMeta.Create("Simple")
+    dataset.add_field(pubsub.DataSetField.CreateScalar("Int32", ua.VariantType.Int32))
+    dataset.add_field(pubsub.DataSetField.CreateScalar("String", ua.VariantType.String))
+    dataset.add_field(pubsub.DataSetField.CreateScalar("Bool", ua.ObjectIds.Boolean))
+    dataset.add_field(pubsub.DataSetField.CreateArray("ArrayInt16", ua.ObjectIds.Double))
```
oroulet, line 25: Maybe we can make the API more pythonic:

```
@dataclass MyDataSet(pubsub.DataSetMeta):
        MyInt32: ua.Int32
         MysTring ua.String
         MyArrayInt16: List[Int16]
         
return MyDataSet
```

examples/pubsub/subscriber_standalone.py
```diff
@@ -0,0 +1,74 @@
...
+from asyncua import ua, pubsub
+from asyncua.pubsub.udp import UdpSettings
+
+# This Parameter must match the Publisher Settings!
+CFG = {
+    'PublisherID': ua.Variant(ua.UInt16(1)),
+    'WriterId': ua.UInt16(1),
+    'DataSetWriterId': ua.UInt16(32),
+    'url': "opc.udp://239.0.0.1:4840"
+}
```
oroulet, line 16: not sure what the constraints are but that one could also be a dataclass so the editor can suggest field names, etc...

examples/pubsub/subscriber_standalone.py
```diff
@@ -0,0 +1,74 @@
...
+    # Register Handler to get the DataSetResults
+    cfg = pubsub.ReaderGroupDataType("ReaderGroup1", True)
+    reader = pubsub.ReaderGroup(cfg)
+    cfg = pubsub.DataSetReaderDataType("SimpleDataSetReader", True, CFG['PublisherID'], CFG['WriterId'], CFG['DataSetWriterId'])
+    dsr = pubsub.DataSetReader(cfg)
+    dsr.set_subscriped(handler)
+    metadata = create_meta_data()
+    await dsr.set_meta_data(metadata)
+    await reader.add_dataset_reader(dsr)
+    return reader
```
oroulet, line 39: generally I think the current api has many "add", "create", "set" methods and function. I would like if possible to use more data structure (dataclass in pyhon) if possible

examples/pubsub/subscriber_standalone.py
```diff
@@ -0,0 +1,74 @@
...
+
+async def main():
+    handler = OnDataRecived()
+    logging.basicConfig(level=logging.INFO)
+    ps = pubsub.PubSub()
+    con = create_connection()
+    reader = await create_reader(handler)
+    await con.add_reader_group(reader)
+    await ps.add_connection(con)
+    async with ps:
```
oroulet, line 69: maybe you are reusing the ua wording, but I think that code is quite complicated. Don't we always need a connection? why not creating it on automatically then. Also do we need to "create" a reader? Can't we "add" many readers? 
Also we could have a default group or create one one the fly if not already done when adding a "reader".

Careful that I did not read spec so I am only commenting from a very naive guessing. I may be wrong. We must be careful not to limit the users

examples/pubsub/subscriber_standalone.py
schroeder-: I think about it, because i already simplified creating datasets. Because it's backed by https://reference.opcfoundation.org/v104/Core/docs/Part14/6.2.2/#6.2.2.1.3 which contains 9 fields. Also we have to convert to and from address space.

oroulet: OK. then maybe your solution is the best we can do. I suppose you have a struct in the background

oroulet: just make sure there is a way to specify all parameters if needed

schroeder-: I tought about the api, to make it more conform to the rest of the code.
```
    dataset.add_scalar("Int32", ua.VariantType.Int32)
    dataset.add_scalar("String", ua.VariantType.String)
    dataset.add_scalar("Bool", ua.ObjectIds.Boolean)
    dataset.add_array("ArrayInt16", ua.ObjectIds.Double)
    dataset.add_field(FieldMetaData(Name="Custom", ArrayDimensions=[3, 3], DataSetFieldId=Guid("936DA01F-9ABD-4D9D-80C7-02AF85C822A8")))
    dsf = DataSetField()
    dsf.Name = "PyObject"
    dsf.BuildInType = ua.ObjectIds.UInt16
    dataset.add_field(dsf)
```
or this 
```
    dsf = DataSetField()
    dsf.Name = "PyObject"
    dsf.BuildInType = ua.ObjectIds.UInt16
    dsfields = [
        pubsub.make_scalar_metafield("Int32", ua.VariantType.Int32),
        pubsub.make_scalar_metafield("String", ua.VariantType.String),
        pubsub.make_scalar_metafield("Bool", ua.ObjectIds.Boolean),
        pubsub.make_array_metafield("ArrayInt16", ua.ObjectIds.Double)
        DataSetField(FieldMetaData(Name="Custom", ArrayDimensions=[3, 3], DataSetFieldId=Guid("936DA01F-9ABD-4D9D-80C7-02AF85C822A8")))
    ]
    dataset = pubsub.DataSetMeta.Create("Simple", dsfields)
```

examples/pubsub/subscriber_standalone.py
schroeder-: I can upgrade the constructors of all class to except the reader/writer/datasets..., as optional parameter to make it more pretty. 
Pubsub has a complex model with named objects: [specs](https://reference.opcfoundation.org/v104/Core/docs/Part14/9.1.1/), so creating default ones is tricky. 
In theory a client can create all objects via address space, import via config file(UABinaryFile) or read the configuration form a opc ua server, so defaults can cause conflicts.

examples/pubsub/subscriber_standalone.py
oroulet: I think the first one is more readble. also can't you write:

```
dataset = pubsub.DataSetMeta.Create("Simple")
dataset.add_scalar("Int32", ua.VariantType.Int32)
    dataset.add_scalar("String", ua.VariantType.String)
    dataset.add_scalar("Bool", ua.ObjectIds.Boolean)
    dataset.add_array("ArrayInt16", ua.ObjectIds.Double)
    dataset.add_field(FieldMetaData(Name="Custom", ArrayDimensions=[3, 3], DataSetFieldId=Guid("936DA01F-9ABD-4D9D-80C7-02AF85C822A8")))
    dataset.add_field(
        DataSetField(
            Name = "PyObject"
            BuildInType = ua.ObjectIds.UInt16
            )
        )
```
Also I have been using Rust opcua lately. the author there require user to use the ua objects directly BUT add several high level constructors. This is typicall Rust but that also would work fine for opcua in python https://github.com/locka99/opcua/blob/master/types/src/node_id.rs#L66 . . then I suppose the code would look something like: (athugh I did not get the difference between DataSetField and FieldMetaData)

```Python
dataset = pubsub.DataSetMeta("Simple")  # or DataSetMeta.new("Simple")
dataset.Fields.append(DataSetField.as_scalar("Int32", ua.VariantType.Int32)
dataset.Fields.append(DataSetField.as_scalar("String", ua.VariantType.String)
dataset.Fields.append((FieldMetaData(Name="Custom", ArrayDimensions=[3, 3], DataSetFieldId=Guid("936DA01F-9ABD-4D9D-80C7-02AF85C822A8")))
dataset.Fields.append(
        DataSetField(
            Name = "PyObject"
            BuildInType = ua.ObjectIds.UInt16
            )
        )

```

schroeder-: I wrapped all opc ua datastructs for pubsub that are used, in this case DataSetField  wraps FieldMetaData, with custom classes. This to implement the logic and provide easier to configure classes. Also data structs and their changes need to be mirrored in the address space if we want to support all use cases. For FieldMetaData this is maybe a bit of overkill.

schroeder-:
So took me sometime to address the api changes.

alexchandel:
Can we get a rebase and merge of this (even as an unstable API)?  PubSub is now used by CODESYS, and it is being adopted by a couple other big vendors.

alexchandel:
@schroeder-  Could you rebase this so we can work on it?

alexchandel:
@oroulet Would you mind releasing a new version (1.1.9?) of asyncua, since #1859 and #1879 added basic pubsub support?

PR #842, opened 2022-03-23
Prevent abstract Types from beeing instantiated

AndreasHeine:


asyncua/common/instantiate_util.py
```diff
@@ -12,13 +14,22 @@
...
+async def instantiate(parent, node_type, nodeid: ua.NodeId=None, bname: Union[str, ua.QualifiedName]=None, dname: ua.LocalizedText=None, idx: int=0, instantiate_optional: bool=True):
     """
     instantiate a node type under a parent node.
     nodeid and browse name of new node can be specified, or just namespace index
     If they exists children of the node type, such as components, variables and
     properties are also instantiated
     """
+    abstract = await is_abstract(node_type)
+    if abstract:
+        raise ValueError(f"Abstract Type with NodeId: {node_type.nodeid} cant be instantiated!")
```
oroulet, line 31: Can't we raise some version of UaError? We can call that method from a client and then we are expected to return som 
UaError

AndreasHeine: sure! any suggestion?

AndreasHeine:
```python
raise ua.UaError(f"Abstract Type with NodeId: {node_type.nodeid} cant be instantiated!")
```

oroulet: We should look into list of errors from spec

AndreasHeine: its not really a statuscode error... its more a violation of part 3

oroulet: one of these I suppose: https://reference.opcfoundation.org/v104/Core/docs/Part4/5.7.2/

oroulet: Bad_InvalidArgument maybe...

AndreasHeine: i would say UaError!
probably we need to try/catch instantiate in AddNode-Call and translate it to a OperationLevel ResultCode?

oroulet: you could raise a ValueError and then catch it in AddNode but I do not see any issue in raising UaError immediatly so we do not need to catch things later. It is goind to be the same use case either from client or Server side. If opcua had a UaValuError it would be nice ;-)
As you want

oroulet (changes requested):

tests/test_common.py
```diff
@@ -1010,6 +1010,17 @@ async def test_instantiate_string_nodeid(opc):
...
 
+async def test_instantiate_abstract(opc):
+    finit_statemachine_type = opc.opc.get_node("ns=0;i=2771")  # IsAbstract=True
+    try:
+        node = await instantiate(opc.opc.nodes.objects, finit_statemachine_type, bname="2:TestFiniteStateMachine")
+    except Exception as e:
+        if isinstance(e, ua.UaError):
+            assert 1
+            return
+    assert 0
```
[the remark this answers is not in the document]
AndreasHeine: legit point! didn't know there is a context manager for that...

PR #857, opened 2022-04-07
fix: read_attribute can optional not raise an exception

### LINKED ISSUE #856: Client should be able to read Tag with a Bad Quality
Hi everyone thanks a lot for your good job and this nice python library, i know this should be a discussion but i sadly am not able to start one the start button stays light green in my case.

We decide to use the library to automate some system test we currently run manually, a test case we have that works with another opcua client dictates that a tag is writen with a bad quality and then is read back to verify this works sadly this throws the following error
```
./env/lib/python3.8/site-packages/asyncua/ua/uatypes.py", line 318, in check
    raise UaStatusCodeError(self.value)
```
I checked the master code since i am using 0.9.92 and this check and throw is also there but in line 303.
Then I took a look at https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/ua/uatypes.py and found this  cryptic message in the check function that causes this 'Use the is_good() method if you do not want an exception.' and to be honest i dont see a way this could work as.

For me it seems like a without reason restriction and i have some ideas on how this restriction could be removed i would be happy to provide a pull request if you agree with me.

Sample code to reproduce this.

```
from asyncua.sync import Client
from asyncua import ua

client = Client('url')
try:
    client.connect()
    value = client.get_node('ns=1;i=9')
    variant= ua.Variant("_", ua.VariantType.Boolean)
    dataValue = ua.DataValue(variant)
    object.__setattr__(datavalue, 'StatusCode_', ua.StatusCode(ua.StatusCodes.Bad))
    value .set_value(dataValue)
    result = value .read_attribute(ua.AttributeIds.Value)
    print(result)
finally:
    client.disconnect()
```

Falital:
* Previously a client could not read values with bad status.
* This functionality can be toggled on and off.
* Added tests.

Fixes #856

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -179,13 +179,13 @@ class Node:
 
     get_value = read_value  # legacy compatibility
 
-    async def read_data_value(self):
+    async def read_data_value(self, raise_on_bad_status=True):
         """
         Get value of a node as a DataValue object. Only variables (and properties) have values.
         An exception will be generated for other node types.
         DataValue contain a variable value as a variant as well as server and source timestamps
         """
-        return await self.read_attribute(ua.AttributeIds.Value)
+        return await self.read_attribute(ua.AttributeIds.Value, None, raise_on_bad_status)
 
     async def write_array_dimensions(self, value):
         """
@@ -286,7 +286,7 @@ class Node:
         result = await self.server.write(params)
         return result
 
-    async def read_attribute(self, attr, indexrange=None):
+    async def read_attribute(self, attr, indexrange=None, raise_on_bad_status=True):
         """
         Read one attribute of a node
         attributeid is a member of ua.AttributeIds
@@ -300,7 +300,8 @@ class Node:
         params = ua.ReadParameters()
         params.NodesToRead.append(rv)
         result = await self.server.read(params)
-        result[0].StatusCode.check()
+        if raise_on_bad_status:
+            result[0].StatusCode.check()
         return result[0]
 
     async def read_attributes(self, attrs):
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -317,8 +317,6 @@ class StatusCode:
     def check(self):
         """
         Raises an exception if the status code is anything else than 0 (good).
-
-        Use the is_good() method if you do not want an exception.
         """
         if not self.is_good():
             raise UaStatusCodeError(self.value)
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ examples/history.db
 /.settings/
 /venv/
 .eggs*
+.coverage
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ A set of command line tools also available: https://github.com/FreeOpcUa/opcua-a
 * `uacall `(call method of a node)
 * `uasubscribe `(subscribe to a node and print datachange events)
 * `uaclient `(connect to server and start python shell)
-* `uaserver `(starts a demo OPC UA server)  
+* `uaserver `(starts a demo OPC UA server)
   `tools/uaserver --populate --certificate cert.pem --private_key pk.pem`
 
 How to generate certificate: https://github.com/FreeOpcUa/opcua-asyncio/tree/master/examples/generate_certificate.sh
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -109,6 +109,37 @@ async def test_multiple_read_and_write_value(server, client):
     with pytest.raises(ua.uaerrors.BadUserAccessDenied):
         await client.write_values([v1, v2, v_ro], [4, 5, 6])
 
+async def test_read_and_write_status_check(server, client):
+    f = await server.nodes.objects.add_folder(3, 'read_and_write_status_check')
+    v1 = await f.add_variable(3, "a", 1)
+    await v1.set_writable()
+
+    testValue = 1
+    testStatusCode = ua.StatusCode(ua.StatusCodes.Bad)
+
+    # set value StatusCode to Bad
+    variant = ua.Variant(testValue, ua.VariantType.Int64)
+    dataValue = ua.DataValue(variant, StatusCode_=testStatusCode)
+    await v1.set_value(dataValue)
+
+    # check that reading the value generates an error
+    # with raise_on_bad_status set to True as default
+    with pytest.raises(ua.UaStatusCodeError):
+        val = await v1.read_data_value()
+
+    # check that reading the value does not generate an error
+    # with raise_on_bad_status set to False
+    val = await v1.read_data_value(False)
+    assert val.Value.Value == testValue, "Value expected " \
+        + str(val) + ", but instead got " + str(testValue)
+    assert val.StatusCode_ == testStatusCode, "StatusCode expected " \
+        + str(val.StatusCode_) + ", but instead got " + str(testStatusCode)
+
+    # check that reading the value generates an error
+    # with raise_on_bad_status set to True
+    with pytest.raises(ua.UaStatusCodeError):
+        val = await v1.read_data_value(True)
+
 async def test_browse_nodes(server, client):
     nodes = [
         client.get_node("ns=0;i=2267"),
--- a/tests/test_subscriptions.py
+++ b/tests/test_subscriptions.py
@@ -1,3 +1,4 @@
+import sys
 import asyncio
 import pytest
 from copy import copy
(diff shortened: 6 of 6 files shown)


asyncua/common/node.py
```diff
@@ -47,6 +47,8 @@ class Node:
     def __init__(self, server, nodeid):
         self.server = server
         self.nodeid = None
+        # this flag indicates that an exception should not be raised when an attribute is read or written.
+        self.noStatusCheckThrow = False
```
oroulet, line 51: should_follow_pep8

asyncua/common/node.py
```diff
@@ -264,6 +266,20 @@ def set_read_only(self):
         """
         return self.set_writable(False)
 
+    def dont_throw_on_bad_status(self):
```
oroulet, line 269: PEP8 or general Pythoin conventions says not getter/Setter when class property can be used (users can set diurectly Node.not_status_code_check

asyncua/common/node.py
```diff
@@ -300,7 +316,7 @@ async def read_attribute(self, attr, indexrange=None):
         params = ua.ReadParameters()
         params.NodesToRead.append(rv)
         result = await self.server.read(params)
-        result[0].StatusCode.check()
+        result[0].StatusCode.check(self.noStatusCheckThrow)
```
oroulet, line 319: not checking status code when writting does not really make sense. Especially when the status code is not returned to the user

public.key
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
oroulet, line 1: what is that? please clean up commit. Only one set of connected changes per commit

run-tests.sh
```diff
@@ -1,2 +1,19 @@
-#!/bin/sh
```
oroulet, line 1: might considere inclusing these things but make a separate PR

asyncua/common/node.py
```diff
@@ -300,7 +316,7 @@ async def read_attribute(self, attr, indexrange=None):
...
         return result[0]
 
     async def read_attributes(self, attrs):
```
oroulet, line 322: I still think it would be better to add a parameter to write_attribute and only suppress status code check for read_data_value. All other methods do not return the status code so they must be informed is somehting wrong happend. The few people who want status code and have bad values can use "read_data_value" or read_attribute

asyncua/common/node.py
Falital: Done

asyncua/common/node.py
Falital: Removed

asyncua/common/node.py
Falital: Removed

public.key
Falital: Removed

run-tests.sh
Falital: Removed

asyncua/common/node.py
Falital: If you restrict this only to read_data_value, you still get an excpetion when you want to call a function that does not return a status code. So instead of getting the requested attribute you get an exception, which i think should not be the way this works.

oroulet: As far as i know all these other methods are just convenience methods. And raising an exception in case of Error is the Python normal case. Something else will create very fancy situation like receiving None when reading an uint32

Falital: This functionality is disabled by default so if the user has activated this he will handle thia in his code. In the use case you describe he will still get an uint32 value, as is now there is no way to get a uint32 with a status othet than good.

asyncua/ua/uatypes.py
```diff
@@ -314,13 +314,15 @@ def __post_init__(self):
...
+        Raises an exception if noException is false and the status code
+        is anything else than 0 (good).
 
-        Use the is_good() method if you do not want an exception.
+        Set no_read_status_check_throw on the node object if you do not want an
+        exception while calling read_attribute functions.
         """
-        if not self.is_good():
+        if not noException and not self.is_good():
             raise UaStatusCodeError(self.value)
```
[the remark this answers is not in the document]
Falital: I agree, restructure the code to call the function conditionally.

asyncua/common/node.py
```diff
@@ -300,7 +300,8 @@ async def read_attribute(self, attr, indexrange=None):
         params = ua.ReadParameters()
         params.NodesToRead.append(rv)
         result = await self.server.read(params)
-        result[0].StatusCode.check()
+        if not dont_raise_on_bad_status:
+            result[0].StatusCode.check()
         return result[0]
```
oroulet, line 305: here we are returning the entire datavalue incl the tstaus code so this is fine to allow not to raise. Same with read_data_value. but for all oter methods we cannot add that option, without getting a broken behaviour

asyncua/common/node.py
```diff
@@ -300,7 +300,8 @@ async def read_attribute(self, attr, indexrange=None):
         params = ua.ReadParameters()
         params.NodesToRead.append(rv)
         result = await self.server.read(params)
-        result[0].StatusCode.check()
+        if not dont_raise_on_bad_status:
```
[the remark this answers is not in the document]
Falital: Changed to "if dont_raise_on_bad_status is not False:"

asyncua/common/node.py
```diff
@@ -247,13 +247,13 @@ async def set_writable(self, writable=True):
...
-        dv = await self.read_attribute(attr)
+    async def set_attr_bit(self, attr, bit, dont_raise_on_bad_status=False):
+        dv = await self.read_attribute(attr, None, dont_raise_on_bad_status)
         val = ua.ua_binary.set_bit(dv.Value.Value, bit)
         await self.write_attribute(attr, ua.DataValue(ua.Variant(val, dv.Value.VariantType)))
 
-    async def unset_attr_bit(self, attr, bit):
-        dv = await self.read_attribute(attr)
+    async def unset_attr_bit(self, attr, bit, dont_raise_on_bad_status=False):
+        dv = await self.read_attribute(attr, None, dont_raise_on_bad_status)
```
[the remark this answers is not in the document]
Falital: With this change functionality is added, nothing gets changed or removed. Code that work before will continue to work with the posible to not raise an execption if the status is not good when an attribute is read.

asyncua/common/node.py
Falital: With this change functionality is added, nothing gets changed or removed. Code that work before will continue to work with the posible to not raise an execption if the status is not good when an attribute is read. No bhaviour is broken with this change.

asyncua/common/node.py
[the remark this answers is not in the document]
Falital: My bad, Changed to "if dont_raise_on_bad_status is False:"

asyncua/common/node.py
jrast: Just as a note from my experience: We added a similar flag in some of our device drivers. This flag was missuesed by some users which wanted to get rid of the exceptions, thinking that they could handle errors in some other way. Fact is: they can't as no error code is returned. In newer driver we provide the flag just for those methods which return a object which provides access to the error code so actual custom error handling can be implemented. So I totaly agree with the statement of @oroulet.

Falital: In a closed system it makes sense to guard the user, but it does not make sense to remove functionality in a library or dont allow something that the protocol allows because the user might misused it. The user in this case knows exactly what he is doing and if you dont offer it he will probably work around it.

asyncua/common/node.py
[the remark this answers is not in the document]
Falital: This flag can now only be used in the read_data_value method. I still think it makes no sense to restrict this only to this method, a library should be flexible enought to cover all user cases.

asyncua/common/node.py
Falital: This flag can now only be used in the read_data_value method. I still think it makes no sense to restrict this only to this method, a library should be flexible enought to cover all user cases.

Falital:
@oroulet it seems tests/test_subscriptions.py::test_subscription_overload hangs in version 3.7.x, i tested this localy with https://hub.docker.com/layers/python/library/python/3.7.12-slim-bullseye/images/sha256-199b55e4e6ec1d4ea81c49c321afbc3cea4140a72e0ebfb8278794d8ef541877?context=explore but there it works. So it has something to do with github.

asyncua/common/node.py
```diff
@@ -179,13 +179,13 @@ async def read_value(self):
...
 
-    async def read_data_value(self):
+    async def read_data_value(self, dont_raise_on_bad_status=False):
         """
         Get value of a node as a DataValue object. Only variables (and properties) have values.
         An exception will be generated for other node types.
         DataValue contain a variable value as a variant as well as server and source timestamps
         """
-        return await self.read_attribute(ua.AttributeIds.Value)
+        return await self.read_attribute(ua.AttributeIds.Value, None, dont_raise_on_bad_status)
```
oroulet, line 188: Looks good now, but could you use a positive option name? double degations are hard for humans It is much easier to understand "raise_on_bad_status=True" than "dont_raise_on_bad_status=False"

Falital: Changed to raise_on_bad_status=True

oroulet:
thanks

Falital:
Happy to contribute

PR #859, opened 2022-04-12
fix: read_attribute can optional not raise an exception

oroulet:
alternative solution. It is cleaner in my mind, but changes the behaviour of the library so maybe this is a bad idea...


Falital:
I integraded this in my pull request.

Issue #430, opened 2021-02-02
Server's CertificateUserManager does'n work


che2116:
**Describe the bug** <br />
The server returns error: BadUserAccessDenied, when client tries to connect to the server with valid certificate and private key.

**To Reproduce**<br />
1. Run server-with-encryption.py from examples directory.
2. Run UaExpert (v.1.5.1) as an OPC UA client.
3. In the UaExpert add new connection in the authentication settings select the certificate (examples/certificates/peer-certificate-example-1.der) and the private key (examples/certificates/peer-private-key-example-1.pem).
4. Try to connect to the server.
5. Server returns error BadUserAccessDenied

**Expected behavior**<br />
The server must establish a connection with client.

**Version**<br />
Python-Version: 3.8.5 <br />
opcua-asyncio Version: master branch

ZuZuD:
I gave a try to `server-with-encryption.py` and it works like a charm from the asyncua client:

```
import sys
import asyncio
from asyncua.crypto import security_policies
from asyncua import ua, Client


c = Client(url="opc.tcp://127.0.0.1:4840");
cert = "../examples/certificates/peer-certificate-example-1.der"
key = "../examples/certificates/peer-private-key-example-1.pem"
await c.set_security(
        security_policies.SecurityPolicyBasic256Sha256,
        cert,
        key,
        mode=ua.MessageSecurityMode.SignAndEncrypt
)
await c.connect()
print(await c.get_node("ns=0;i=20001").read_browse_name())
print(await c.get_node("ns=0;i=20002").read_browse_name())
print(await c.get_node("ns=0;i=20002").read_value())
```

It prints
```
QualifiedName(0:MyObject)
QualifiedName(0:MyVariable)
0.2
```

And from the server side:
```
[...]
INFO:asyncua.server.internal_session:Created internal session ('127.0.0.1', 39150)
INFO:asyncua.server.internal_session:Create session request
INFO:asyncua.server.internal_server:get endpoint
INFO:asyncua.server.uaprocessor:Activate session request (None)
INFO:asyncua.server.internal_session:activate session
INFO:asyncua.server.internal_session:Activated internal session ('127.0.0.1', 39150) for user User(role=<UserRole.User: 3>, name='test_user')
INFO:asyncua.server.uaprocessor:Read request (User(role=<UserRole.User: 3>, name='test_user'))
[...]

PR #861, opened 2022-04-19
Fix User Authentication with certificates from UAExpert. Relates to #430

TobiasNils:
UAExpert sends its application certificate as client certificate in the initial `CreateSessionRequest`.
The user certificate is sent as `ua.X509IdentityToken` in the request body and has to be looked up there.


oerpli:
If I am interested in this fix should I install from the current master or are there plans for a release in the near future that will include this PR?

PR #862, opened 2022-04-22
Server DataValue StatusCode handling

schroeder-:
Currently it is not possible to only set the StatusCode of a DataValue for a Node, because the type will be checked.

This fails with BadTypeMissmatch
```
dvar = ua.DataValue(None, StatusCode_=ua.StatusCode(ua.StatusCodes.BadDeviceFailure))
await v.write_value(dvar)
```

Also if the StatusCode is bad, the Value of the DataValue should be set to Null. See: https://reference.opcfoundation.org/v104/Core/docs/Part4/7.7.1/


AndreasHeine:
guess you need to pass a variant instead of none... type info is part of the Variant!

schroeder-:
Also Variant(ua.Null()) does not work. Also you don't have to provide a Value, it is optional. Before this pr you always had to provide a matching Variant to set the StatusCode, which complicates the code.

AndreasHeine:
this does not work -> ua.Variant(None, ua.VariantType.Double) ?

schroeder-:
With this Pr, the Value is set to Null and the type is not checked if the Statuscode is bad. So it doens't matter.

AndreasHeine:
i like it if it is closer to spec and userfriendly!

asyncua/server/address_space.py
```diff
@@ -780,8 +780,19 @@ async def write_attribute_value(self, nodeid: ua.NodeId, attr: ua.AttributeIds,
...
+            # https://reference.opcfoundation.org/v104/Core/docs/Part4/7.7.1/
+            # If the StatusCode indicates an error then the value is to be ignored and the Server shall set it to null.
+            value = ua.DataValue(
+                ua.Variant(ua.Null, ua.VariantType.Null),
+                value.StatusCode,
+                value.ServerTimestamp,
+                value.SourceTimestamp,
+                value.ServerPicoseconds,
+                value.SourcePicoseconds
+            )
```
oroulet, line 793: there is a Dataclass.replace method instead of all that code

asyncua/server/address_space.py
```diff
@@ -780,8 +780,19 @@ async def write_attribute_value(self, nodeid: ua.NodeId, attr: ua.AttributeIds,
         attval = node.attributes.get(attr, None)
         if attval is None:
             return ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid)
-
-        if not self._is_expected_variant_type(value, attval, node):
+        if value.StatusCode is not None and not value.StatusCode.is_good():
```
oroulet, line 783: so what you are doing here is forcing value to None when a user write a dv with a bad statuscode. I do not like chaging what users do, that will lead to surprises... but if this is really what the spec wants....

schroeder-: Maybe we can relax that and just skip the type check for the Value? Because it should be possible to just set the StatusCode without knowing what ValueType is excepted.

oroulet: the spec really seems to specify we should set it to null so OK

PR #866, opened 2022-04-26
Fixed code generator for structures and enums (fixes #810)

### LINKED ISSUE #810: AttributeError: module 'asyncua.ua' has no attribute 'ParameterResultDataType'
I'm having trouble reading an ExtensionObject from an Server running on <1.04. The Server runs on a B&R PLC.
When calling load_type_definition() I get folllowing error:

```
Deprecated since spec 1.04, call load_data_type_definitions


@dataclass
class FetchResultDataType:

    
    FetchResultDataType structure autogenerated from xml
    




@dataclass
class TransferResultErrorDataType:

    
    TransferResultErrorDataType structure autogenerated from xml
    

    Status:ua.Int32 = 0
    Diagnostics:ua.DiagnosticInfo = ua.DiagnosticInfo()



@dataclass
class TransferResultDataDataType:

    
    TransferResultDataDataType structure autogenerated from xml
    

    SequenceNumber:ua.Int32 = 0
    EndOfResults:ua.Boolean = True
    ParameterDefs:List[ua.ParameterResultDataType] = field(default_factory=list)

Failed to execute auto-generated code from UA datatype: 

@dataclass
class TransferResultDataDataType:

    
    TransferResultDataDataType structure autogenerated from xml
    

    SequenceNumber:ua.Int32 = 0
    EndOfResults:ua.Boolean = True
    ParameterDefs:List[ua.ParameterResultDataType] = field(default_factory=list)
  File "C:\Users\Admin\AppData\Local\Programs\Python\Python38-32\lib\site-packages\asyncua\common\structures.py", line 281, in _generate_python_class
    exec(code, env)
  File "<string>", line 4, in <module>
  File "<string>", line 12, in TransferResultDataDataType
AttributeError: module 'asyncua.ua' has no attribute 'ParameterResultDataType'
```

Reading this issue https://github.com/FreeOpcUa/opcua-asyncio/issues/509 it seems like the error has been fixed but I still get it. 
I'm running the current master (0.9.92)

jrast:
This is a fix for the the issue reported in #810. The fix consists of two parts:

1. Updated Code generator: The generated code uses quoted type annotations for type hints in dataclasses and default factories for field defaults for fields of custom types. Both changes are required to prevent dependency issues between the different structures.
2. Updated dataclass serializers and deserializers: Minor changes to the serializer and deserializer where necessary to handle the quoted type annotations of the generated classes


asyncua/common/structures.py
```diff
@@ -83,13 +83,15 @@ class {self.name}:
...
+            uatype = f"'ua.{sfield.uatype}'"
             if sfield.array:
                 uatype = f"List[{uatype}]"
             if uatype == 'List[ua.Char]':
                 uatype = 'String'
-            code += f"    {sfield.name}:{uatype} = {sfield.value}\n"
-        print(code)
+            uavalue = sfield.value
+            if isinstance(uavalue, str) and uavalue.startswith("ua."):
+                uavalue = f"field(default_factory=lambda: {uavalue})"
```
oroulet, line 93: ok. so by using the factory you are delaying the evaluation of the type. Is that your entire point?
so if we have an issue we are going to discover it later and not at import time... I am a bit afraid of that. although  I see if will help in some situations

jrast: Yes, the evaluation is delayed until instanciation of a struct of this type. However, without dependency resolution between the structs prior to generating code for them this seems the only way to go for me. Or am I missing something?

oroulet: nope. I think you are right.
We also have the same issue with structres104.py I suppose?

oroulet: nice ;-)

oroulet:
ok. fine we can merge. Can you rebase? Looks like there are some conflicts since I merged a few other MR

asyncua/common/structures.py
jrast: I'll have a look into the code of structures104 to see if it's possible to reuse some of the functionality from there. If there is already something there which works I would prefere this solution.

jrast: Actualy I don't think the DataTypeSorter works as intended for nested structures as the `__lt__` function only looks at it's own dependencies and not at dependencies of the dependencies.

jrast:
Sorry, I messed up rebasing... Trying to fix this...

asyncua/common/structures.py
oroulet: might be...

PR #868, opened 2022-04-26
reject not supported identity tokens

schroeder-:
prevents unsupported identity tokens to create a session.
For example reject anonymous connection if not set via set_security_IDs, currently this was possible and was a security issue.
See: https://github.com/FreeOpcUa/python-opcua/issues/1457#issuecomment-1109502689_

### DIFF
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -10,7 +10,7 @@ from struct import unpack_from
 import os
 import logging
 from urllib.parse import urlparse
-from typing import Coroutine
+from typing import Coroutine, Tuple
 
 from asyncua import ua
 from .user_managers import PermissiveUserManager, UserManager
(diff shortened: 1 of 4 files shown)


asyncua/server/internal_server.py
```diff
@@ -352,3 +353,9 @@ def check_user_token(self, isession, token):
             password = password.decode('utf-8')
 
         return user_name, password
+
+    def get_supported_tokens(self) -> Tuple[type]:
+        return self._supported_tokens
+
+    def set_supported_tokens(self, token_classes: Tuple[type]) -> None:
+        self._supported_tokens = token_classes
```
oroulet, line 361: code looks ok. but do we need these setter and getter methods? they do not add any value, you could as well make `supported_tokens`public.
Also github complains about missing newline at end of file here

oroulet:
nice and clean

PR #833, opened 2022-03-06
enable mypy

vruge:
- mirror change from mypy
- ignore mypy errors
  - better to fix issues step by step
  
## rough summary of the errors ( not complete )

```python
# I'm not find stubs package for mypy
from sortedcontainers import SortedDict  # type: ignore

# Optinal[String]
ProductUri: String = None  # type: ignore

# "frozen" argument must be True or False. I guess it's false posetive 
disable_error_code = misc

# Name "SubHandler" already defined (possibly by an import)  [no-redef]
class SubHandler(SubHandler):  # type: ignore
# asyncua/ua/uaprotocol_auto.py:2579: error: Name "Union" already defined (possibly by an import)  [no-redef]
# asyncua/common/connection.py:12: error: Name "InvalidSignature" already defined (possibly by an import)  [no-redef]
class Union:  # type: ignore


# Cannot find implementation or library stub for module named "IPython"
from IPython import embed  # type: ignore

# Argument 1 to "NodeId" has incompatible type "int"; expected "Union[Int32, String, Guid, ByteString]"
# "FourByteNodeId" has incompatible type "int"; expected "Union[Int32, String, Guid, ByteString]"
disable_error_code = arg-type


# asyncua/client/ha/reconciliator.py:384: error: "Reconciliator" has no attribute "hook_add_to_map_error"  [attr-defined]
# asyncua/client/ha/reconciliator.py:411: error: "Reconciliator" has no attribute "hook_add_to_map"; maybe "add_to_map"?  [attr-defined]
# asyncua/client/ha/reconciliator.py:432: error: "Reconciliator" has no attribute "hook_del_from_map"; maybe "del_from_map"?  [attr-defined]
disable_error_code = attr-defined


# asyncua/ua/uaerrors/_base.py:39: error: Need type annotation for "_subclasses" (hint: "_subclasses: Dict[<type>, <type>] = ...")  [var-annotated]
# asyncua/server/address_space.py:42: error: Need type annotation for "datachange_callbacks" (hint: "datachange_callbacks: Dict[<type>, <type>] = ...")  [var-annotated]
disable_error_code = var-annotated


# asyncua/ua/uatypes.py:311: error: Incompatible types in assignment (expression has type "int", variable has type "UInt32")  [assignment]
# asyncua/ua/uatypes.py:380: error: Incompatible types in assignment (expression has type "int", variable has type "Union[Int32, String, Guid, ByteString]")  [assignment]
[...]

PR #873, opened 2022-04-30
Update Nodeset to 1.04.10

schroeder-:
- Reading Locale in Nodeset fixed.
- Fix model comparison, comparing against 1.04.10 failed
- UaBinary: Allow fields in struct with "AllowSubtypes" 
- Handle DataSetFieldFlags and other Flags with Length != 4 Byte


schroeder-:
Ok I run in to mypy issues. There where some ignore hints places with #833 in asyncua/ua/uaprotocol_auto.py that are removed when generating the new types.

oroulet:
Arrg i was afraid of that MR. I should have seen that. I can revert if we need to

schroeder-:
Maybe we can work around this, I will look into it.

schroeder-:
Currently its only the Union class, i added a list for mypy ignore classes.

schroeder-:
@[oroulet](https://github.com/oroulet) is there anything that blocks the merge? I need the changes for my pubsub branch.

oroulet:
Sorry forgot that one

PR #897, opened 2022-05-20
add test showing new bug

oroulet:


oroulet:
@schroeder- looks like your commit form a few days ago introduces a regression that was not catched by our tests. see my commit in this PR. I am a bit lost in what is happening here, so if you have an idea it would be nice https://github.com/FreeOpcUa/opcua-asyncio/commit/2b64f0f16fe9d27d62321f00aa40182c673d8ac8

schroeder-:
Will look into it

schroeder-:
See #898

oroulet:
thanks @schroeder- !

PR #900, opened 2022-05-22
Add support for pre 1.04 optionsets

schroeder-:
based on https://github.com/FreeOpcUa/python-opcua/pull/1451 from @deeTEEcee


schroeder-:
Had somethings wrong, correct it. Made the code simpler.

PR #906, opened 2022-06-05
EventFilter: handle objects in EventTypes

### LINKED ISSUE #905: Add event support for nested objects (NodeClass.Object)
I see that in https://github.com/FreeOpcUa/opcua-asyncio/pull/415 we added support for variables with properties but it looks like we're still missing `NodeClass.Object` in that filter. These can then contain their own properties, variables, or nested objects.

schroeder-:
If a EventType has a object as child, add it's variables and properties to the eventfilter. Also handles variables that contain nested variables or objects.
Fixes #905

### DIFF
--- a/asyncua/common/events.py
+++ b/asyncua/common/events.py
@@ -1,9 +1,11 @@
 import copy
-
+from typing import Dict, List, TYPE_CHECKING
 from asyncua import ua
 import asyncua
 from ..ua.uaerrors import UaError
 from .ua_utils import get_node_subtypes
+if TYPE_CHECKING:
+    from asyncua.common.node import Node
 
 
 class Event:
@@ -142,10 +144,7 @@ async def get_filter_from_event_type(eventtypes):
     return evfilter
 
 
-async def append_new_attribute_to_select_clauses(attribute, select_clauses, already_selected, parent_variable):
-    browse_path = []
-    if parent_variable:
-        browse_path.append(await parent_variable.read_browse_name())
+async def _append_new_attribute_to_select_clauses(attribute: "Node", select_clauses: List[ua.SimpleAttributeOperand], already_selected: Dict[str, str], browse_path: List[ua.QualifiedName]):
     browse_path.append(await attribute.read_browse_name())
     string_path = '/'.join(map(str, browse_path))
     if string_path not in already_selected:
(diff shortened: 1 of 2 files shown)


oroulet:
We should add tests for that kind of things tou

asyncua/common/events.py
```diff
@@ -142,34 +144,42 @@ async def get_filter_from_event_type(eventtypes):
...
 
-async def append_new_attribute_to_select_clauses(attribute, select_clauses, already_selected, parent_variable):
-    browse_path = []
-    if parent_variable:
-        browse_path.append(await parent_variable.read_browse_name())
+async def _append_new_attribute_to_select_clauses(attribute: "Node", select_clauses: List[ua.SimpleAttributeOperand], already_selected: Dict[str, str], browse_path: List[ua.QualifiedName]):
     browse_path.append(await attribute.read_browse_name())
     string_path = '/'.join(map(str, browse_path))
     if string_path not in already_selected:
         already_selected[string_path] = string_path
```
deeTEEcee, line 151: Shouldn't `already_selected` just be using the node identifier instead? Why are we using browser path here?

schroeder-: I guess because filters work on browse_names not on nodeids?

deeTEEcee: Yeah that makes sense. i just saw that it's required in the spec. Rather strange to me considering this quote from their [docs](https://reference.opcfoundation.org/v104/Core/docs/Part3/5.2.4/): 

> Unlike NodeIds, the BrowseName cannot be used to unambiguously identify a Node. Different Nodes may have the same BrowseName.

asyncua/common/events.py
```diff
@@ -142,34 +144,42 @@ async def get_filter_from_event_type(eventtypes):
...
-        select_clauses.append(op)    
+        select_clauses.append(op)
+
+
+async def _select_clause_from_childs(child: "Node", select_clauses: List[ua.SimpleAttributeOperand], already_selected: Dict[str, str], browse_path: List[ua.QualifiedName]):
+    for property in await child.get_properties():
+        await _append_new_attribute_to_select_clauses(property, select_clauses, already_selected, [*browse_path])
+    for variable in await child.get_variables():
+        await _append_new_attribute_to_select_clauses(variable, select_clauses, already_selected, [*browse_path])
+        await _select_clause_from_childs(variable, select_clauses, already_selected, browse_path + [await variable.read_browse_name()])
```
deeTEEcee, line 163: Question: Will "variables" actually have nested objects or do they only contain properties? (Just to make sure I have the definition right, properties are leaf nodes that hold values)

schroeder-: See: https://reference.opcfoundation.org/v104/Core/docs/Part3/7.8/

PR #913, opened 2022-06-06
mypy: enable union-attr check

vruge:


asyncua/client/ha/reconciliator.py
```diff
@@ -10,6 +10,7 @@
 from typing import TYPE_CHECKING, Dict, Set, Union, List, Optional
 from sortedcontainers import SortedDict  # type: ignore
 from asyncua import ua, Client
+from asyncua.sync import Subscription
```
oroulet, line 13: Check TYPE_CHECKING before importintg stuff only for typeckeching otherwise it cost time and increases current and future problems with circular imports.
Also I may have missed that issue in your other PRs but you should use that everywhere 
```

if typing.TYPE_CHECKING:
    import expensive_mod
    ```

File: asyncua/ua/uatypes.py [hunk 2]
```diff
@@ -950,7 +950,7 @@ class DataValue:
     data_type = NodeId(25)
 
     Encoding: Byte = field(default=0, repr=False, init=False, compare=False)
-    Value: Optional[Variant] = None
+    Value: Variant = field(default_factory=Variant)
```
schroeder-, line 953: This change is incorrect, Value in DataValue is optional!

schroeder-: I am not sure, but would say it is Optional so it should be optional, that a user knows it can be None.

see: https://reference.opcfoundation.org/v104/Core/docs/Part4/7.7.1/


> If the StatusCode indicates an error then the value is to be ignored and the Server shall set it to null.

oroulet: Sorry I missed that one. Also I do not have time to udnerstand this now, so I propose to revert that PR

asyncua/common/statemachine.py
```diff
@@ -287,8 +287,8 @@ def __init__(self, server: Server=None, parent: Node=None, idx: int=None, name:
         if name is None:
             self._name = "FiniteStateMachine"
         self._state_machine_type = ua.NodeId(2771, 0)
-        self._available_states_node: Optional[Node] = None
-        self._available_transitions_node: Optional[Node] = None
+        self._available_states_node: Node = None
+        self._available_transitions_node: Node = None
```
oroulet, line 291: I also missed that one. you cannot make a variable as type Node and then set it to None!!!!
@vruge

asyncua/common/statemachine.py
```diff
@@ -88,22 +88,22 @@ def __init__(self, server: Server=None, parent: Node=None, idx: int=None, name:
...
         self._idx = idx
         self._optionals = False
-        self._current_state_node: Optional[Node] = None
+        self._current_state_node: Node = None
         self._current_state_id_node = None
         self._current_state_name_node = None
         self._current_state_number_node = None
         self._current_state_effective_display_name_node = None
-        self._last_transition_node: Optional[Node] = None
+        self._last_transition_node: Node = None
```
oroulet, line 101: and same

asyncua/common/statemachine.py
```diff
@@ -88,22 +88,22 @@ def __init__(self, server: Server=None, parent: Node=None, idx: int=None, name:
         self.locale = "en-US"
         self._server = server
         self._parent = parent
-        self._state_machine_node: Optional[Node] = None
+        self._state_machine_node: Node = None
```
oroulet, line 91: and same

asyncua/client/ua_client.py
```diff
@@ -249,7 +251,7 @@ def __init__(self, timeout: float = 1):
         self._subscription_callbacks = {}
         self._timeout = timeout
         self.security_policy = ua.SecurityPolicy()
-        self.protocol: Optional[UASocketProtocol] = None
+        self.protocol: UASocketProtocol = None
```
oroulet, line 254: same here too

PR #916, opened 2022-06-07
avoid import only for type checking

vruge:


oroulet:
seems to be a conflict...

PR #917, opened 2022-06-07
[WIP] guarded get for optional value

vruge:


vruge:
is probably not needed, see:
https://github.com/FreeOpcUa/opcua-asyncio/pull/913#discussion_r891189245

schroeder-:
This pr is needed see #928

PR #919, opened 2022-06-08
speed up eventfilter creation

schroeder-:
Reduced network transactions:
- instead of 3 browse request (get_properties, get_variables, get_objects) use 1.
- no read_browse_name 
- refactored logic into 1 function

The addition to Node is needed because in events.py you can't import node because of circular dependency.

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -725,3 +725,9 @@ class Node:
         await self.server.unregister_nodes([self.nodeid])
         self.nodeid = self.basenodeid
         self.basenodeid = None
+
+    def init_child_node(self, nodeid: ua.NodeId):
+        """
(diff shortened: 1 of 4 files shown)


asyncua/common/node.py
```diff
@@ -725,3 +725,9 @@ async def unregister(self):
         await self.server.unregister_nodes([self.nodeid])
         self.nodeid = self.basenodeid
         self.basenodeid = None
+
+    def init_child_node(self, nodeid: ua.NodeId):
```
oroulet, line 729: the created node is not a child. Maybe rename to new_node(nodeid) and make it static so it can also be called from not instantiated class

oroulet:
seems to need rebase again...

schroeder-:
Rebased the pr

PR #923, opened 2022-06-09
XMLExport: make export values optional

schroeder-:
Allow disable of exporting values from variables.
Normally you don't need the values in the nodeset. As side effect it reduces the time need to export a nodeset and the xml size.

Because of the the incompleteness of the xml encoding I would disable the variable export per default.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -645,11 +645,12 @@ class Client:
         importer = XmlImporter(self)
         return await importer.import_xml(path, xmlstring)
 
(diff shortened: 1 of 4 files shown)


wvalcke:
Hi Alexander,

Thanks for the fast feedback, i also solved it by setting explicitly the value to None like for example
`await varnode.write_value(ua.Variant(None, ua.VariantType.ExtensionObject))`

I looked into the code you made your changes into, and i saw it did not generate the Value element if the stored value was None in the node variable.

Thanks.

PR #927, opened 2022-06-12
enable mypy arg-type

vruge:


schroeder-:
Sry you can not change anything in asyncua/ua/uaprotocol_auto.py, the content is auto generated from the specs.

Issue #926, opened 2022-06-10
DataType syntax


om327:
Hello, 

I've been working on integrating some asyncio servers with LabVIEW's SystemLink OPC UA plug-in. 

Following a couple of service requests we've noticed that there is a slight difference how DataType is represented. In the older opcua-python library this is represended as, for example 'String' or 'Double'.  The asyncio version seems to display DataType as an ENUM for example i=12 (string). Is there a reason for this change and is it possibel tto enforce a particular format for DataType as it seems to prevent some clients from connecting?

Many thanks!

schroeder-:
There isn't any change, DataTypes are always NodeIds, there are no Enums behind it. Maybe you have a minimal example?

om327:
I've tried to use the same commands for both implimentations but i'm still getting the same DataType difference when using the asyncio (DataType i=11) rather than DataType=Double in the client application. In UaExpert is doens't seem to make a difference and in both instances DataType is shown as Double.

python-opcua
```
import time
from opcua import ua, Server

if __name__ == "__main__":

    server = Server()
    server.set_endpoint("opc.tcp://0.0.0.0:4840/freeopcua/server/")

    uri = "http://examples.freeopcua.github.io"
    idx = server.register_namespace(uri)

    myobj = server.nodes.objects.add_object(idx, "MyObject")
    myvar = myobj.add_variable(idx, "MyVariable", 6.7)
    myvar.set_writable()

    server.start()
    
    try:
        count = 0
        while True:
            time.sleep(1)
            count += 0.1
            myvar.set_value(count)
    finally:
        server.stop()
``` 

opcua-ascyncio
```
import asyncio
from asyncua import ua, Server

async def main():
    server = Server()
    await server.init()
    server.set_endpoint('opc.tcp://0.0.0.0:4840/freeopcua/server/')

    uri = 'http://examples.freeopcua.github.io'
    idx = await server.register_namespace(uri)

    myobj = await server.nodes.objects.add_object(idx, 'MyObject')
    myvar = await myobj.add_variable(idx, 'MyVariable', 6.7)
    await myvar.set_writable()
    async with server:
        while True:
            await asyncio.sleep(1)
            new_val = await myvar.get_value() + 0.1
            await myvar.write_value(new_val)


if __name__ == '__main__':
    asyncio.run(main())
```

oroulet:
[...]

PR #937, opened 2022-06-16
return error on DataTypeDefinition empty

schroeder-:
See #926


om327:
Hi, I recieved a new SystemLink licence from NI this morning and tested using the minimal-server. DataType now appears as Double and can be subscribed by the client! Thank you!

image

Issue #945, opened 2022-06-22
XML generated reference seems incorrect for Datatype in nested struct


wvalcke:
Using the example server-custom-structures-and-enums.py we have 2 struct definitions like

    snode1, _ = await new_struct(server, idx, "MyStruct", [
        new_struct_field("MyBool", ua.VariantType.Boolean),
        new_struct_field("MyUInt32List", ua.VariantType.UInt32, array=True),
    ])
    snode3, _ = await new_struct(server, idx, "MyNestedStruct", [
        new_struct_field("MyStructArray", snode1, array=True),
        new_struct_field("UUID", ua.VariantType.String),
    ])

snode3 uses a field of type snode1. When looking at the generated XML code we have the following:

```
  <UADataType NodeId="ns=1;i=1" BrowseName="1:MyStruct">
    <DisplayName>MyStruct</DisplayName>
    <Description>MyStruct</Description>
    <References>
      <Reference ReferenceType="HasSubtype" IsForward="false">i=22</Reference>
      <Reference ReferenceType="HasEncoding">ns=1;i=2</Reference>
    </References>
    <Definition Name="MyStruct">
      <Field Name="MyBool" DataType="i=1" />
      <Field Name="MyUInt32List" DataType="i=7" ValueRank="0" ArrayDimensions="1" />
    </Definition>
  </UADataType>
  <UADataType NodeId="ns=1;i=5" BrowseName="1:MyNestedStruct">
    <DisplayName>MyNestedStruct</DisplayName>
    <Description>MyNestedStruct</Description>
    <References>
      <Reference ReferenceType="HasSubtype" IsForward="false">i=22</Reference>
      <Reference ReferenceType="HasEncoding">ns=1;i=6</Reference>
    </References>
    <Definition Name="MyNestedStruct">
      <Field Name="MyStructArray" DataType="ns=2;i=1" ValueRank="0" ArrayDimensions="1" />
      <Field Name="UUID" DataType="i=12" />
    </Definition>
  </UADataType>
```

The MyStruct data type has node id : ns=1;i=1. The MyNestedStruct data type has node id ns=1;i=5. 
If we look at the definition of the field MyStructArray, we see it refers to DatatType ns=2;i=1, but this node does not exists. Importing this into UA modeler for example gives an error. I think the DataType reference should be ns1;i=1.

Used  asyncua 0.9.4, python 3.9.9

schroeder-:
[...]

PR #948, opened 2022-06-23
Use XmlExporter _node_to_string

gstvg:
instead of Node.to_string 
_node_to_string correctly maps namespaces idxs


gstvg:
fixes  #945

Issue #950, opened 2022-06-25
HA client tests seems to fail on python 3.10


oroulet:
@ZuZuD the HA client seems broken again: https://github.com/FreeOpcUa/opcua-asyncio/runs/7048622278?check_suite_focus=true

not sure what is happening, all merged requests I merged passed (As far as I remember)

schroeder-:
Tests pass again. Maybe a issue with python version or some network problems in the test runner.

PR #951, opened 2022-06-27
Fixed XmlExporter raising BadAttributeIdInvalid on optional fields

smorokin:
Hi,

for my usecase (B&R PLC as OPCUA server) `XmlExporter` was raising some `BadAttributeIdInvalid` errors on fields that were optional in the first place. 
I suspect that some time ago `raise_on_bad_status=True` was added as a default to `Node.read_attribute` and that caused different behavior here, since in almost all of the cases the the value is checked before working with it.


smorokin:
No idea why the test with python3.10 fails on `tests/test_ha_client.py::TestHaClient::test_reconnect`.

schroeder-:
The failing test is a open issue: #950

Issue #961, opened 2022-07-12
Server import_xml performance downgrade from 0.8.4 to 0.9.94


jsbruglie:
Hello! Following the discussion on Gitter I figured I should summarize my findings an add an example to reproduce the problem.
I've observed that when moving from asyncua 0.8.4 to the latest 0.9.9X the servers take substantially longer to start.

**To Reproduce**<br />

I included a minimal server as well as an example server import file here: [example.zip](https://github.com/FreeOpcUa/opcua-asyncio/files/9090982/example.zip).
The script itself just initializes a server, imports data from a file, starts and stops.

```python
import asyncio
import pathlib
from time import time
from asyncua import Server

cur_dir = pathlib.Path(__file__).parent.resolve()
example_import_xml = cur_dir / "example.xml"

async def main():
    start = time()
    server = Server()
    await server.init()
    server.set_endpoint('opc.tcp://0.0.0.0:4841/freeopcua/server/')
    time_init = time()
    await server.import_xml(path=example_import_xml)
    time_import_xml = time()
    await server.start()
    await server.stop()
    stop = time()

    duration_init = time_init - start
    duration_import_xml = time_import_xml - time_init
    total = stop - start
    print("init {:.3f} s, import {:.3f} s, total {:.3f} s".format(duration_init, duration_import_xml, total))

if __name__ == "__main__":
    asyncio.run(main())
```

**Benchmarks**<br />

|        | init (sec) | import (sec) | total (sec) |
|--------|------------|--------------|-------------|
| 0.8.4  | 1.203      | 2.063        | 3.266       |
| 0.9.94 | 2.359      | 119.655      | 123.062     |

It seems that `import_xml` is now taking close to 60 times more than previously.
I'll add the profiling results in a comment soon.

**Version**<br />
Python-Version:3.9.10 (but also reproduced on 3.10)<br />
opcua-asyncio Version (e.g. master branch, 0.9): 0.9.94 (master)

swamper123:
Hey ho,

I haven't got a solution for your problem, but I want that you keep some things in your mind:

1. Since Jul 8, 2020, many things happened in this repo. Even with the xml_import.
2. XML in combination with Python is mostly slow. Also we switched from lxml to builtin xml.
[...]

PR #963, opened 2022-07-12
Fix XML import bottlenecks

elmovw-evac:
This PR fixes performance issues with XML imports, making it possible to actually use large files exported from software like TIA Portal.

**I have not properly tested the actual results of `_add_missing_reverse_references()` - I've only made assumptions and it might depend on some fixes I have in my personal fork.**

**I also have not run any unit tests.**

## Performance results
I tested this on a humongous 130 000 line nodeset that was directly exported from Siemens's TIA portal. The file needed a separate fix for sorting dependent data types correctly, and I should make another PR for that.

### Opc.Ua.Di.NodeSet2.xml
The nodeset depends on OPC Foundation's Opc UA DI Nodeset, so it is imported first:

| Task | `master` | `elmo/xmlimport-perf` |
| --- | --- | --- |
| `_sort_nodes_by_parentid()` | 73 ms | 27 ms |
| `_add_missing_reverse_references()` | 4.1 s | 0.22 s |
| Total time in `import_xml()` | 4.7 s | 0.66 s |

### The humongous TIA Portal export nodeset

| Task | `master` | `elmo/xmlimport-perf` |
| --- | --- | --- |
| `_sort_nodes_by_parentid()` | **128 s** | 0.99 s |
| `_add_missing_reverse_references()` | **N/A > 1 hour** | 10.3 s |
| Total time in `import_xml()` | **N/A** | 24.1 s |

The original implementation of `_add_missing_reverse_references` seems to clog up the program and I gave up after giving it **over 1 hour to run**. The humongous nodeset has ~44x the amount of data, but `_sort_nodes_by_parentid` takes ~1700x the time (roughly $O(n^2)$ ). Assuming that the `master` version of `_add_missing_reverse_references` is equally slow, I'd need to give it **2 hours** to complete.

### `_sort_nodes_by_parentid`

Many expensive `in` -operators are used against `list` objects - I simply swapped them to `set`:s, which provide constant-time lookup. The original implementation seems to have quadratic complexity, as it takes ~1 700x the time to run over ~40x data.

### `_add_missing_reverse_references`

I took the liberty to remove the 3-level for loop with redundant calls to `node.get_references()` by caching the references in a `dict`. I figured that it is not strictly necessary to add reverse type definition references to every node in existence - I might be wrong and I have not properly tested the results of this method.


jsbruglie:
I tried this version on my test suite and the tests pass, reduces total execution time from 19 minutes to 30 seconds (!), as the bottleneck is the starting up 3 separate OPC UA servers from XML import files.
However, my tests are really simple and mostly care about the structure of the server, no fancy reference magic happening there.

jsbruglie:
Thank you @elmovw-evac and @oroulet, I believe this fixes https://github.com/FreeOpcUa/opcua-asyncio/issues/961!

PR #965, opened 2022-07-13
structures104: sort type dependencies correctly

### LINKED ISSUE #941: asyncua.common.xmlimporter: failure adding node NodeData
Hello,

I have an OPC.xml dumped from a Siemens s7 I would like to load, so I can simulate it and work on it.
When I try to load it looks okay except for that one warning:

```
...
INFO:asyncua.common.xmlimporter:Importing xml node (QualifiedName(NamespaceIndex=2, Name='REMOTE'), NodeId(Identifier=3066, NamespaceIndex=2, NodeIdType=<NodeIdType.Numeric: 2>)) as (QualifiedName(NamespaceIndex=2, Name='REMOTE') NodeId(Identifier=3066, NamespaceIndex=2, NodeIdType=<NodeIdType.Numeric: 2>))
WARNING:asyncua.common.xmlimporter:failure adding node NodeData(nodeid:NodeId(Identifier=3066, NamespaceIndex=2, NodeIdType=<NodeIdType.Numeric: 2>))
WARNING:asyncua:"The parent node id does not to refer to a valid node."(BadParentNodeIdInvalid)
INFO:asyncua:Starting server!
WARNING:asyncua.server.server:Endpoints other than open requested but private key and certificate are not set.
INFO:asyncua.server.internal_server:starting internal server
INFO:asyncio:<Server sockets=(<asyncio.TransportSocket fd=6, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 4840)>,)> is serving
INFO:asyncua.server.binary_server_asyncio:Listening on 127.0.0.1:4840
DEBUG:asyncua.server.server:OPC UA Server(opc.tcp://127.0.0.1:4840) server started
```

But when I inspect the contents on the server with the opcua-client I cannot find any of the defined nodes/variables etc.

**To Reproduce**<br />
My XML is imported like this:
```
async def main():
    server = Server()
    await server.init()
    await server.import_xml("OPC.xml")
    server.set_endpoint('opc.tcp://127.0.0.1:4840')
    async with server:
        while True:
            await asyncio.sleep(1)
```

In the xml node 3065 got parsed just fine but 3066 doesn't have a references node.
Maybe that causes the warning and cancels the parsing altogether?
```
  <UADataType NodeId="ns=1;i=3065" BrowseName="1:ANY">
    <DisplayName>ANY</DisplayName>
    <References>
      <Reference ReferenceType="HasSubtype">ns=1;i=3066</Reference>
      <Reference ReferenceType="HasSubtype" IsForward="false">i=3</Reference>
    </References>
  </UADataType>
  <UADataType NodeId="ns=1;i=3066" BrowseName="1:REMOTE">
    <DisplayName>REMOTE</DisplayName>
  </UADataType>
```

And the opcua-client also shows 2 warnings which may be related?
```
asyncua.client.client - WARNING - Deprecated since spec 1.04, call load_data_type_definitions')
asyncua.client.client - WARNING - Deprecated since spec 1.04, call load_data_type_definitions')
```

**Expected behavior**<br />
I expect to see the same nodes/variables as defined in the s7.

**Version**<br />
Python-Version: 3.10
opcua-asyncio Version: master
OPC.xml generated by Generator Product="TIA Portal" Edition="STEP 7 - Server Interface Generation" Version="V1.0.0"

elmovw-evac:
This should fix #941, and other cases of importing XML nodesets with custom data structures.

The original problem was that each node would only be compared with its direct dependencies, instead of considering the whole type dependency chain.

### DIFF
--- a/asyncua/common/structures104.py
+++ b/asyncua/common/structures104.py
@@ -5,7 +5,7 @@ import uuid
 import logging
 import re
 import keyword
-from typing import Union, List, TYPE_CHECKING, Tuple, Optional, Any, Dict
+from typing import Union, List, TYPE_CHECKING, Tuple, Optional, Any, Dict, Set
 from dataclasses import dataclass, field
 
 from asyncua import ua
@@ -290,7 +290,11 @@ async def _generate_object(name, sdef, data_type=None, env=None, enum=False, opt
 
 
 class DataTypeSorter:
-    def __init__(self, data_type, name, desc, sdef):
+    dtype_index: Dict[ua.NodeId, 'DataTypeSorter'] = {}
+    referenced_dtypes: Set[ua.NodeId] = set()
+
+    def __init__(self, data_type: ua.NodeId, name: str,
+                 desc: ua.ReferenceDescription, sdef: ua.StructureDefinition):
         self.data_type = data_type
         self.name = name
         self.desc = desc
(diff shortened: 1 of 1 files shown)


oroulet:
looks good. but it would be nice with an test for that if you had one so we check it works and we make sure it does not break so easily

oroulet:
merging since this is good, but we really should have tests

PR #985, opened 2022-08-03
do not overwrite custom RequestHeader

brubbel:
https://reference.opcfoundation.org/v105/Core/docs/Part4/7.33/
https://reference.opcfoundation.org/v105/Core/docs/Part4/7.1/

> The AdditionalParametersType parameter is used as value of the additionalHeader field of the RequestHeader and ResponseHeader parameters. It allows Clients and Servers to pass additional named parameters with Service requests or responses. These named parameters may be defined by the OPC UA specification, a companion specification or be specific to a vendor implementation.

Spec v1.05 allows a custom `AdditionalHeader` in the `RequestHeader` of e.g. `ReadRequest` or `HistoryReadRequest`.
This PR allows to provide these parameters in the `RequestHeader` without being overwritten by the `send_request` method of `UaClient`.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -114,13 +114,13 @@ class UASocketProtocol(asyncio.Protocol):
         :param message_type: UA Message Type (optional)
         :return: Future that resolves with the Response
         """
-        request.RequestHeader = self._create_request_header(timeout)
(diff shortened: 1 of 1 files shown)


brubbel:
merged.

Issue #977, opened 2022-07-28
BadSessionIdInvalid after 30s


jmtatsch:
**Describe the bug** <br />
Connect to an S7 with opcua-client, client attempts to negotiate a timeout of an hour but gets only 30s.
`asyncua.client.client - WARNING - Requested session timeout to be 3600000ms, got 30000ms instead')`
After 30s client gets BadSessionIdInvalid and crashes.

```
asyncua.client.ua_client.UASocketProtocol - WARNING - ServiceFault (BadSessionIdInvalid, diagnostics: DiagnosticInfo(SymbolicId=None, NamespaceURI=None, Locale=None, LocalizedText=None, AdditionalInfo=None, InnerStatusCode=None, InnerDiagnosticInfo=None)) from server received  in response to ReadRequest')
uawidgets.utils - ERROR - "The session id is not valid."(BadSessionIdInvalid)')
Traceback (most recent call last):
  File "/home/jtatsch/.local/lib/python3.8/site-packages/opcua_widgets-0.6.1-py3.8.egg/uawidgets/utils.py", line 21, in wrapper
    result = func(self, *args)
  File "/home/jtatsch/.local/lib/python3.8/site-packages/opcua_client-0.8.4-py3.8.egg/uaclient/mainwindow.py", line 437, in _update_actions_state
    if node.read_node_class() == ua.NodeClass.Method:
  File "/home/jtatsch/.local/lib/python3.8/site-packages/asyncua-0.9.94-py3.8.egg/asyncua/sync.py", line 94, in wrapper
    result = self.tloop.post(aio_func(*args, **kwargs))
  File "/home/jtatsch/.local/lib/python3.8/site-packages/asyncua-0.9.94-py3.8.egg/asyncua/sync.py", line 52, in post
    return futur.result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/home/jtatsch/.local/lib/python3.8/site-packages/asyncua-0.9.94-py3.8.egg/asyncua/common/node.py", line 144, in read_node_class
    result = await self.read_attribute(ua.AttributeIds.NodeClass)
  File "/home/jtatsch/.local/lib/python3.8/site-packages/asyncua-0.9.94-py3.8.egg/asyncua/common/node.py", line 302, in read_attribute
[...]

Issue #974, opened 2022-07-27
Raise exception of _renew_channel_task correctly


elboerto:
**Describe the bug** <br />
`async def _renew_channel_loop(self)` and the related task created in `async def create_session(self):` does not raise Exceptions correctly. If any exceptions occur within the _renew_channel_loop task, the client stops working but does not crash.

**To Reproduce**<br />
Throw an exception in the "while True" loop of `_renew_channel_loop(self)`

**Expected behavior**<br />
The client should crash which would enable the user to handle it. The exception of the created task `_renew_channel_task` should be raised by e.g. calling `_renew_channel_task.result()`

**Screenshots**<br />
![image](https://user-images.githubusercontent.com/31927927/181220378-7deaae12-b7d9-4adf-acab-676dae6e5a15.png)

**Version**<br />
Python-Version: 3.9.13
opcua-asyncio Version (e.g. master branch, 0.9): master branch, 0.9.94

oroulet:
PR welcome if you have an idea for an API

schroeder-:
Fixed in #986

PR #585, opened 2021-06-02
[Subscription creation] adjust MaxKeepAliveCount based on RevisedPublishingInterval

ZuZuD:
## Problem statement 
As discussed in #557, the MaxKeepAliveCount has to be adjusted once the server replies with a different `RevisedPublishingInterval` value. 

This change adds the methods to update the parameters of existing subscriptions. Hence, we can capture the `CreateSubscriptionResult`, re-calculate the `MaxKeepAliveCount` when needed and update the sub accordingly.

## Testing
Let's set a PublishInterval of 1ms for a server which only supports 100ms and above, then check that the `MaxKeepAliveCount` value is recalculated and the subscription is modified:
```
2021-06-02 14:19:47,596 - asyncua.client.ua_client.UaClient - DEBUG - create_subscription
2021-06-02 14:19:47,596 - asyncua.client.ua_client.UASocketProtocol - DEBUG - Sending: CreateSubscriptionRequest(TypeId=FourByteNodeId(Identifier=787, NamespaceIndex=0, NodeIdType=<NodeIdType.FourByte: 1>), RequestHeader_=RequestHeader(AuthenticationToken=NodeId(Identifier=b'\x9e\xebjY\x10\xf0\x88ej)\xdc pf\xd1\xec\xbd\xf0W.I\x9c\x84pUH\xf0U"\xa4\xdf\x8c', NamespaceIndex=0, NodeIdType=<NodeIdType.ByteString: 5>), Timestamp=datetime.datetime(2021, 6, 2, 21, 19, 47, 596143), RequestHandle=4, ReturnDiagnostics=0, AuditEntryId=None, TimeoutHint=4000, AdditionalHeader=ExtensionObject(TypeId=NodeId(Identifier=0, NamespaceIndex=0, NodeIdType=<NodeIdType.TwoByte: 0>), Body=None)), Parameters=CreateSubscriptionParameters(RequestedPublishingInterval=1, RequestedLifetimeCount=10000, RequestedMaxKeepAliveCount=2700000, MaxNotificationsPerPublish=10000, PublishingEnabled=True, Priority=0))
2021-06-02 14:19:47,720 - asyncua.client.ua_client.UaClient - INFO - create_subscription success SubscriptionId 48
[...]

PR #986, opened 2022-08-04
[Client] Add watchdog for connection status

schroeder-:
This adds a new task to pull the connection status, which will prevent a session timeout as of #977. Is also used in the next step to inform all subscriptions about a connection lose. 
Also add a hook to check if this task an `_renew_channel_loop` thrown a exception. See #974.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -63,7 +63,8 @@ class Client:
         self.secure_channel_timeout = 3600000  # 1 hour
         self.session_timeout = 3600000  # 1 hour
         self._policy_ids = []
-        self.uaclient: UaClient = UaClient(timeout, self.check_connection)
+        self.uaclient: UaClient = UaClient(timeout)
+        self.uaclient.pre_request_hook = self.check_connection
         self.user_certificate = None
         self.user_private_key = None
         self._server_nonce = None
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -29,11 +29,7 @@ class Client:
     use UaClient object, available as self.uaclient
     which offers the raw OPC-UA services interface.
     """
-
-    _username = None
-    _password = None
-
-    def __init__(self, url: str, timeout: float = 4):
+    def __init__(self, url: str, timeout: float = 4, watchdog_intervall: float = 10):
         """
         :param url: url of the server.
             if you are unsure of url, write at least hostname
@@ -41,6 +37,8 @@ class Client:
         :param timeout:
             Each request sent to the server expects an answer within this
             time. The timeout is specified in seconds.
+        :param watchdog_intervall:
+            The time between checking if the server is still alive. The timeout is specified in seconds.
         Some other client parameters can be changed by setting
         attributes on the constructed object:
         See the source code for the exhaustive list.
@@ -65,7 +63,7 @@ class Client:
         self.secure_channel_timeout = 3600000  # 1 hour
         self.session_timeout = 3600000  # 1 hour
         self._policy_ids = []
-        self.uaclient: UaClient = UaClient(timeout)
+        self.uaclient: UaClient = UaClient(timeout, self._check_tasks)
         self.user_certificate = None
         self.user_private_key = None
         self._server_nonce = None
@@ -74,7 +72,9 @@ class Client:
         self.max_messagesize = 0  # No limits
         self.max_chunkcount = 0  # No limits
         self._renew_channel_task = None
+        self._watch_task = None
         self._locale = ["en"]
+        self._watchdog_intervall = watchdog_intervall
 
     async def __aenter__(self):
         await self.connect()
@@ -421,11 +421,39 @@ class Client:
             _logger.warning("Requested session timeout to be %dms, got %dms instead", self.secure_channel_timeout, response.RevisedSessionTimeout)
             self.session_timeout = response.RevisedSessionTimeout
         self._renew_channel_task = asyncio.create_task(self._renew_channel_loop())
+        self._watch_task = asyncio.create_task(self._watchdog_loop())
         return response
+
+    async def _check_tasks(self):
+        # Check if a background task has finished and if a exception is thrown rethrow it with result
+        # use half the session timeout or the supply value from constructor
+        if self._renew_channel_task is not None:
+            if self._renew_channel_task.done():
+                await self._renew_channel_task
+        if self._watch_task is not None:
+            if self._watch_task.done():
+                await self._watch_task
+
+    async def _watchdog_loop(self):
+        """
+        Checks if the server is alive
+        """
+        timeout = min(self.session_timeout / 1000 / 2, self._watchdog_intervall)
+        try:
+            while True:
+                await asyncio.sleep(timeout)
+                # @FIXME handle state change
+                _ = await self.nodes.server_state.read_value()
+
+        except asyncio.CancelledError:
+            pass
+        except Exception:
+            _logger.exception("Error in watchdog loop")
+            raise
 
     async def _renew_channel_loop(self):
         """
-        Renew the SecureChannel before the SessionTimeout will happen.
+        Renew the SecureChannel before the SecureChannelTimeout will happen.
         In theory we could do that only if no session activity
         but it does not cost much..
         """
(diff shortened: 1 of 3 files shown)


asyncua/client/client.py
```diff
@@ -406,11 +410,39 @@ async def create_session(self):
...
+        # Check if a background task has finished and if a exception is thrown rethrow it with result
+        # use half the session timeout or the supply value from constructor
+        if self._renew_channel_task is not None:
+            if self._renew_channel_task.done():
+                await self._renew_channel_task
+        if self._watch_task is not None:
+            if self._watch_task.done():
+                await self._watch_task
+
+    async def _watchdog_loop(self):
```
oroulet, line 426: maybe it it possible to give that method a more precise name? watchdog is very generic. `monitor_server_task` for example?

asyncua/client/client.py
```diff
@@ -406,11 +410,39 @@ async def create_session(self):
...
+        except asyncio.CancelledError:
+            pass
+        except Exception:
+            _logger.exception("Error in watchdog loop")
+            raise
+
     async def _renew_channel_loop(self):
         """
-        Renew the SecureChannel before the SessionTimeout will happen.
+        Renew the SecureChannel before the SecureChannelTimeout will happen.
```
oroulet, line 445: I do not remember, do we check session timeout? but anyway with that watcdog task I suppose we never go into timeout

asyncua/client/client.py
```diff
@@ -406,11 +410,39 @@ async def create_session(self):
...
 
+    async def _check_tasks(self):
+        # Check if a background task has finished and if a exception is thrown rethrow it with result
+        # use half the session timeout or the supply value from constructor
+        if self._renew_channel_task is not None:
+            if self._renew_channel_task.done():
+                await self._renew_channel_task
+        if self._watch_task is not None:
+            if self._watch_task.done():
+                await self._watch_task
```
oroulet, line 424: I like your solution of a brackground task and a method to check status an await task. Good work.
A few suggestions:
* A user application might have several threads running so we may need the exceptoin to be raised several times in several threads. Thus we probably should save any exception happening here and make sure it raise at every further call of that method
* Also can you make that mehtod public? it might be useful for users who want to check connection manually

asyncua/client/ua_client.py
```diff
@@ -143,12 +145,15 @@ async def send_request(self, request, timeout: Optional[float] = None, message_t
         Returns response object if no callback is provided.
         """
         timeout = self.timeout if timeout is None else timeout
+        if self._before_request_hook:
+            # This will propagade exceptions from background tasks to the libary user before calling a request which will
+            # timeout then.
+            await self._before_request_hook()
```
oroulet, line 151: Does that work? Now you are checking the connection just before trying to use it. Aren't you going to discover the connection right away snow anyway?

ZuZuD:
I'm not sure **not** allowing the session to timeout is the right default approach here. The traditional way of keeping a session active is to create a subscription, thereafter you transparently rely on the protocol KeepAlive mechanism and the session isn't supposed to time out (see #585 for the related work and references on this).

If I understand this patch correctly, the problem you're trying to solve with it is to achieve the same behaviour but without a subscription. If that's indeed the case, then I don't think this corner case warrants increasing the complexity of our client's library.

schroeder-:
I know that a subscription is the best way to keep a session alive. But we received some bug reports that didn't use subscriptions and had session timeouts. Because some servers limit the amount of subscriptions, I don't want to create a subscription just for session keep alive.
Maybe we can make it optional just for people that don't want to use subscriptions?

jmtatsch:
>The traditional way of keeping a session active is to create a subscription, thereafter you transparently rely on the protocol KeepAlive mechanism and the session isn't supposed to time out (see #585 for the related work and references on this).
I know very little about opcua and its traditions but to subscribe to for example the servers clock is certainly **the worst way** to keep the session alive. Now I don't know the resolution of the clock but giving the client every nano second an unused timestamp to fetch is certainly keeping everyone and the line busy, certainly more then one request every 30s.

oroulet:
From what I have seen a few years back, all clients where doing hard polling every seconds to keep session alive. I also kind of remember the spec advicing to do polling on server status node, but I could not find anything now when looking for it...
So I do not have anything against having a task doing exactly that.
Also the main advantage I see to the proposed approach is the possibility to offers to clients a simple and easily discoverable method like get_connection_status() than returns a connection status immediately, thus allowing all threads of a user application to check connection state instead of having implement some polling themselv.

oroulet:
could also help with that kind of things btw: https://github.com/FreeOpcUa/python-opcua/issues/521

ZuZuD:
> I know that a subscription is the best way to keep a session alive. But we received some bug reports that didn't use subscriptions and had session timeouts. Because some servers limit the amount of subscriptions, I don't want to create a subscription just for session keep alive. Maybe we can make it optional just for people that don't want to use subscriptions?

That would seem odd to have server implementations where ```max_subscription < max_session```. What I meant was that I believe for most of the use-cases we got, we could recommend the user to create a subscription (even an empty one) to work around this.

Also, there's a [KeepAlive task](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ha/ha_client.py#L424-L474) already available with the HaClient which covers most of what's suggested here. If we want to go down this road, that would probably make sense to consolidate these efforts.

oroulet:
> Also, there's a [KeepAlive task](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/ha/ha_client.py#L424-L474) already available with the HaClient which covers most of what's suggested here. If we want to go down this road, that would probably make sense to consolidate these efforts.

I am rather sure the HD client is overkill for most applications, but some features from there could be backported. That keepalive task for example

asyncua/client/client.py
schroeder-: The server checks session timeout, session timeout is reseted every time any service is used by the client. Its just a typo because SecureChannelTimeout was meant.

asyncua/client/client.py
schroeder-: the exception is rethrown each time someone calls this method.

asyncua/client/ua_client.py
schroeder-: I thought this way the original exception is thrown, not just that the connection is broken.

oroulet: that is a valid point

tests/test_connections.py
```diff
@@ -28,3 +29,17 @@ async def test_max_connections_1(opc):
...
+    port = find_free_port()
+    srv = Server()
+    await srv.init()
+    srv.set_endpoint(f'opc.tcp://127.0.0.1:{port}')
+    await srv.start()
+    async with Client(f'opc.tcp://127.0.0.1:{port}', timeout=0.5, watchdog_intervall=1) as cl:
+        await srv.stop()
+        await asyncio.sleep(2)
+        with pytest.raises(ConnectionError):
+            await cl.get_namespace_array()
```
oroulet, line 45: can you test direction the check_connection() method here?
Ad call it several times and make sure the exception is raised every time

oroulet: as a kind of documentation and check the functionnnality

oroulet:
looks good for me. Has anyone any idea to even improve it?

asyncua/client/ua_client.py
```diff
@@ -244,22 +249,24 @@ class UaClient:
     In this Python implementation  most of the structures are defined in
     uaprotocol_auto.py and uaprotocol_hand.py available under asyncua.ua
     """
-    def __init__(self, timeout: float = 1):
+    def __init__(self, timeout: float = 1, before_request_hook: Callable[[], Awaitable[None]] = None):
```
oroulet, line 252: I am wondering if this is best as a class argument or if it could simply be a public property that users can overwrite

cl = UaClient()
cl.pre_request_hook = whatever()
cl.post_request_hook = whatever()

run it
...

also I thing the word `pre`and `post`is more common than `before` but this is not important

oroulet: Same for UaSocketPRotocol. That is a bit less code. less discoverable but this is really internal stuff

schroeder-: update the interface to properties and changed the name to pre_request_hook

PR #1001, opened 2022-08-18
Always close session

cziebuhr:
Properly close session and cancel _renew_channel_task if connect() fails in activate_session(). This avoids the need for calling disconnect() after connect() failed, leading to better call semantics.

Unfortunately, CloseSessionRequest wasn't allowed to be called in an inactivated session (it is now).
For compatibility with older asyncua servers, BadUserAccessDenied gets ignored in CloseSessionResponse.

This also cleans up some tests, which needed to call disconnect() but expected it to fail.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -261,15 +261,20 @@ class Client:
             await self.open_secure_channel()
             try:
                 await self.create_session()
+                try:
+                    await self.activate_session(username=self._username, password=self._password, certificate=self.user_certificate)
+                except Exception:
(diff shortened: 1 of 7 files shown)


cziebuhr:
Had to do a second fix: disconnect_socket() is a nonblocking operation, which doesn't wait till socket is closed. So calling disconnect() right after, leads to the situation where UaClient.protocol.state is not yet set to CLOSED and close_secure_channel() tries to write to the transport, which is already closed, leading to an error.

PR #1009, opened 2022-08-23
add Integer and UInteger types

schroeder-:
For completness, these datatypes are abstract! If they are used in structs, abstract types are either Variant or ExtensionObjects.
If they only contain basic types (int16, float, double..) they are Variants.
https://github.com/FreeOpcUa/opcua-client-gui/issues/87

### DIFF
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -927,6 +927,13 @@ def get_shape(mylist):
     return dims
 
 
+# For completness, these datatypes are abstract!
+# If they are used in structs, abstract types are either Variant or ExtensionObjects.
+# If they only contain basic types (int16, float, double..) they are Variants
+UInteger = Variant
+Integer = Variant
+
+
 @dataclass(frozen=True)
 class DataValue:
     """


asyncua/ua/uatypes.py
```diff
@@ -927,6 +927,13 @@ def get_shape(mylist):
     return dims
 
 
+# For completness, these datatypes are abstract!
+# If they are used in structs, abstract types are either Variant or ExtensionObjects.
+# If they only contain basic types (int16, float, double..) they are Variants
+UInteger = Variant
+Integer = Variant
```
oroulet, line 934: Is that correct? All data types are defined as 

```
Class UInt32(int):
    pass
```
    but Integer should be a Variant type? Looks strnage but there are so many strange things in ua....

oroulet: @schroeder- ?

schroeder-: Sry I forget about it. After digging I am not sure if Variant is correct or ExtensionObject. I only found mention that abstract classes should be ExtensionObjects.
But to your point a abstract class can have an unknown content, that is why either Variant or a ExtenstionObject must be used. Otherwise the decoder can't know how long the data is. 
I will look in the specs to find the correct handling.

oroulet: OK. I see, then we should also add Number, Float, etc,... if it is allowed to use them in structs

schroeder-: I don't know if it is allowed, but there are servers in the wild that use them. 
Will add them if I know which type they should be.

oroulet: I could not find anything about using these insie a struct but maybe you have better sources...

PR #1012, opened 2022-08-25
fixed an UnboundLocalError

smorokin:
Variable is set in try block, but would be appended to the list after the except block


smorokin:
I don't think that failing test has something to do with my change.

smorokin:
@oroulet  thank you!

Issue #1014, opened 2022-08-29
Illegal access to Address Space


artfire52:
**Bug** <br />
With the example of server-with-encryption, it is possible to acces Address Space without encryption and authentication.

**To Reproduce**<br />
To gain an access to Address Space, we have to send hello message, create a secure channel with security none and create a session (activate the session is not required). After we can try to read and write nodes with regular Read and Write Request. 

**Expected behavior**<br />
The expected behaviour would be to require authentication in the application layer (ie session) to provide access to Address Space.

**Version**<br />
Python-Version:Python 3.8.10

opcua-asyncio Version (e.g. master branch, 0.9):
branch master, commit https://github.com/FreeOpcUa/opcua-asyncio/commit/54e54fa6b1383849fac4039f0e9c06902454297c (last commit 29/08/2022)

PR #1015, opened 2022-08-31
check if session is activated

schroeder-:
All services that require a active session are now check. Before it was possible to access these services without authentication, because this is done in activate session. See #1014


oroulet:
rebase so we can check fi it passes then we merge

Issue #1013, opened 2022-08-29
Denial of Service


artfire52:
**The bug** 
It is possible to create a denial of service by sending a malformed packet. The server will be trap in an infinite loop and consume lot of memory.  I create an issue instead of sending mail following instruction from  [How to report a security issue? #902 ](https://github.com/FreeOpcUa/opcua-asyncio/issues/902)

**Reproduce**<br />
To reproduce the denial of service, you have to send a packet with a null size.

**Explanation**<br />
The function ```data_received``` in file binary_server_asyncio.py (opcua-asyncio/asyncua/server/binary_server_asyncio.py)  continue to work on the same buffer. 
```python
def data_received(self, data):
        self._buffer += data
        # try to parse the incoming data
        while self._buffer:
            try:
                buf = Buffer(self._buffer)
                try:
                    header = header_from_binary(buf)
                except NotEnoughData:
                    logger.debug('Not enough data while parsing header from client, waiting for more')
                    return
                if len(buf) < header.body_size:
                    logger.debug('We did not receive enough data from client. Need %s got %s', header.body_size,
                                 len(buf))
                    return
                # we have a complete message
                self.messages.put_nowait((header, buf))
                self._buffer = self._buffer[(header.header_size + header.body_size):]
            except Exception:
                logger.exception('Exception raised while parsing message from client')
                return

```
The value of ```(header.header_size + header.body_size)``` is equal to zero. In the function ```header_from_binary``` in (opcua-asyncio/asyncua/ua/ua_binary.py), the bod_size is equal to zero at the beginning (due to our malformed packet). At the end ```hdr.header_size``` is equal to 12 and ```hdr.body_size``` is equal to -12.

```python
def header_from_binary(data):
    hdr = ua.Header()
    hdr.MessageType, hdr.ChunkType, hdr.packet_size = struct.unpack("<3scI", data.read(8))
    hdr.body_size = hdr.packet_size - 8
    if hdr.MessageType in (ua.MessageType.SecureOpen, ua.MessageType.SecureClose, ua.MessageType.SecureMessage):
        hdr.body_size -= 4
        hdr.ChannelId = Primitives.UInt32.unpack(data)
        hdr.header_size = 12
    return hdr
```
So in the function ```data_received```, the buffer is never updated, then we are trapped in an infinite loop.

**Version**<br />
Python-Version:Python 3.8.10<br />
opcua-asyncio Version (e.g. master branch, 0.9):
[...]

PR #1016, opened 2022-08-31
Server: prevent denial of service from malformed packet

schroeder-:
See #1013. Prevent packets with invalid content from tricking the internal buffer and trigger a denial of service.

### DIFF
--- a/asyncua/server/binary_server_asyncio.py
+++ b/asyncua/server/binary_server_asyncio.py
@@ -67,12 +67,12 @@ class OPCUAProtocol(asyncio.Protocol):
                 except NotEnoughData:
                     # a packet should at least contain a header otherwise it is malformed (8 or 12 bytes)
                     logger.debug('Not enough data while parsing header from client, empty the buffer')
-                    self._buffer = b''
+                    self.transport.close()
                     return
                 if header.header_size + header.body_size <= header.header_size:
                     # malformed header prevent invalid access of your buffer
                     logger.error(f'Got malformed header {header}')
-                    self._buffer = b''
+                    self.transport.close()
                 else:
                     if len(buf) < header.body_size:
                         logger.debug('We did not receive enough data from client. Need %s got %s', header.body_size,
--- a/asyncua/server/binary_server_asyncio.py
+++ b/asyncua/server/binary_server_asyncio.py
@@ -65,15 +65,22 @@ class OPCUAProtocol(asyncio.Protocol):
                 try:
                     header = header_from_binary(buf)
                 except NotEnoughData:
-                    logger.debug('Not enough data while parsing header from client, waiting for more')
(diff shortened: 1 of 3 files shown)


tests/test_connections.py
```diff
@@ -1,10 +1,10 @@
 # coding: utf-8
 import asyncio
 import pytest
-
-from asyncua import Client, Server
+import asyncio
+from asyncua import Client, Server, ua
 from asyncua.ua.uaerrors import BadMaxConnectionsReached
-
+import struct
```
oroulet, line 7: import struct must be in first import group at the top

asyncua/server/binary_server_asyncio.py
```diff
@@ -65,15 +65,22 @@ def data_received(self, data):
                 try:
                     header = header_from_binary(buf)
                 except NotEnoughData:
-                    logger.debug('Not enough data while parsing header from client, waiting for more')
+                    # a packet should at least contain a header otherwise it is malformed (8 or 12 bytes)
+                    logger.debug('Not enough data while parsing header from client, empty the buffer')
+                    self._buffer = b''
```
oroulet, line 70: that class handle one connection so one client, if we receive data that do not even have a header, aren't we fucked and can as well close connection?

asyncua/server/binary_server_asyncio.py
```diff
@@ -65,15 +65,22 @@ def data_received(self, data):
...
+                    logger.error(f'Got malformed header {header}')
+                    self._buffer = b''
+                else:
+                    if len(buf) < header.body_size:
+                        logger.debug('We did not receive enough data from client. Need %s got %s', header.body_size,
+                                    len(buf))
+                        return
+                    # we have a complete message
+                    self.messages.put_nowait((header, buf))
+                    self._buffer = self._buffer[(header.header_size + header.body_size):]
```
oroulet, line 83: thinking loud here but maybe we should just close connection in all these cases

oroulet:
@schroeder- rebase?

asyncua/server/binary_server_asyncio.py
schroeder-: I think closing the connection, so the client can retry is a good idea.

oroulet:
looks good but need rebase

schroeder-:
Done

kohtala:
Hi. That was not a rebase, but merge. Looking at the nine commits, it would be best to squash the commits into one and make the commit based on the head of master branch. Just force push the one commit on your fix branch.

Please, we use this and we scan for vulnerabilities. It gives an alarm. A fix would be welcome. A clean pull request could speed up getting the fix in a release.

oroulet:
made a rebase here https://github.com/FreeOpcUa/opcua-asyncio/pull/1039

kohtala:
Thank you for this. Is there information when we could expect a release with the fix?

PR #1017, opened 2022-09-02
Always take NamespaceIndex into account when comparing NodeIds

cziebuhr:
This fix mostly applies to custom event or reference types, which coincidentally have the same Identifier than a predefined type with NamespaceIndex 0.

### DIFF
--- a/asyncua/common/events.py
+++ b/asyncua/common/events.py
@@ -216,7 +216,7 @@ async def select_event_attributes_from_type_node(node: "Node", attributeSelector
     curr_node = node
     while True:
         attributes.extend(await attributeSelector(curr_node))
-        if curr_node.nodeid.Identifier == ua.ObjectIds.BaseEventType:
+        if curr_node.nodeid == ua.NodeId(ua.ObjectIds.BaseEventType):
             break
         parents = await curr_node.get_referenced_nodes(
             refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse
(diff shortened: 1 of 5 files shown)


schroeder-:
Looks good from my point. Changing the tests is more a style thing, but its good to prevent copy paste errors and promoted a correct style.

Currently we have a problem with our checks. Will investigate if I have time.

asyncua/common/events.py
```diff
@@ -216,7 +216,7 @@ async def select_event_attributes_from_type_node(node: "Node", attributeSelector
     curr_node = node
     while True:
         attributes.extend(await attributeSelector(curr_node))
-        if curr_node.nodeid.Identifier == ua.ObjectIds.BaseEventType:
+        if curr_node.nodeid == ua.NodeId(ua.ObjectIds.BaseEventType):
```
oroulet, line 219: I remember I tried to avoid creates objects for comparison. Not sure what the cost of the NodeId class creation is...
we could at least avoid it by comparing manually namespace to 0

oroulet: All these NodeIds could be created and cached, but then it might be a memory issue. There are many standard nodes

cziebuhr: For the UaProcessor case, we could also just use a static lookup dict with NodeIds as keys.

cziebuhr: I haven't found it anymore, tried to reproduce, but couldn't. Nevertheless, I have some ideas in mind to speed up NodeId usage, maybe I will open another issue for it. But this will take some time.

PR #1018, opened 2022-09-05
Fix interpretation of IncludeSubtypes

### LINKED ISSUE #233: Browsing HasSubType references without "includeSubtypes"
Hi,

while browsing nodes in the example OPC UA server (examples/server-example.py) I noticed something weird (at least weird to me).

If I specify the HasSubType reference for the "referenceTypeId" parameter in the browse request, I will only get these references in the browse response if I also set the "includeSubtypes" parameter to true.
Other reference type IDs in the browse request wont show this behavior and I will get the requested references without setting includeSubTypes.

I only took a brief look at the code, but I would guess the behavior comes from 5f488ab2 (not verified, that this is the code path taken):

```python
    def _suitable_reftype(self, ref1, ref2, subtypes):
        """
        """
        if not subtypes and ref2.Identifier == ua.ObjectIds.HasSubtype:
            return False
```

I didn't see anything in the OPC UA specification that would account for special handling of the HasSubType reference type in this case. Other OPC UA stacks also return references with HasSubType directly (without setting "IncludeSubtypes").

Is there a reason why the OPC UA server behaves this way?

Best regards

cziebuhr:
Fix interpretation of IncludeSubtypes in Browse requests.

According to the spec, IncludeSubtypes indicates whether subtypes of the ReferenceType should be included in the browse. If TRUE, then instances of ReferenceType and its subtypes are returned.

The current implementation has a different interpretation: Get instances of ReferenceType and subtypes thereof. If IncludeSubtypes is FALSE, then exclude the type "HasSubtype".

The same applies to RelativePath requests.

Fixes #233.
There may be conflicts with #1017, I can rebase when it got merged.

**WARNING: This change has compatibility drawbacks, not sure how to deal with it in all cases.**

* Old client with new server (or other opcua servers):
If using add_object with type definition (or directly using instantiate), no children will get created because get_children_descriptions(includesubtypes=False) on the ObjectType node which calls get_references(refs=ua.ObjectIds.HierarchicalReferences, includesubtypes=False) won't return anything. This is a breaking change, but also fixes behaviour when used with other opcua servers.

* New client with old server:
Some places call get_referenced_nodes(refs=HasSubtype, includesubtypes=True). Includesubtypes could actually be set to False because HasSubtype probably won't get subclassed. It should be harmless to keep it set to True, keeping compatibility with old servers and just wasting some cpu cycles there. However, for code cleanliness I removed the includesubtypes=True parameter from the call. It's still active, as it defaults to True.

### DIFF
--- a/asyncua/common/events.py
+++ b/asyncua/common/events.py
@@ -219,7 +219,7 @@ async def select_event_attributes_from_type_node(node: "Node", attributeSelector
         if curr_node.nodeid.Identifier == ua.ObjectIds.BaseEventType:
             break
         parents = await curr_node.get_referenced_nodes(
-            refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse, includesubtypes=True
+            refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse
         )
         if len(parents) != 1:  # Something went wrong
             return None
@@ -278,9 +278,7 @@ async def get_event_obj_from_type_node(node):
                     # Add the sub-properties of the VariableType
                     for prop in await var.get_properties():
                         await self._add_new_property(prop, var)
-                parents = await curr_node.get_referenced_nodes(refs=ua.ObjectIds.HasSubtype,
-                                                               direction=ua.BrowseDirection.Inverse,
-                                                               includesubtypes=True)
+                parents = await curr_node.get_referenced_nodes(refs=ua.ObjectIds.HasSubtype, direction=ua.BrowseDirection.Inverse)
                 if len(parents) != 1:  # Something went wrong
                     raise UaError("Parent of event type could not be found")
                 curr_node = parents[0]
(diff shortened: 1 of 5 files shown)


schroeder-:
> Old client with new server (or other opcua servers):
> If using add_object with type definition (or directly using instantiate), no children will get created because get_children_descriptions(includesubtypes=False) on the ObjectType node which calls get_references(refs=ua.ObjectIds.HierarchicalReferences, includesubtypes=False) won't return anything. This is a breaking change, but also fixes behaviour when used with other opcua servers.
I would rly care about that, because in practices there aren't that many servers allowing to create nodes from the client.

cziebuhr:
> I would rly care about that, because in practices there aren't that many servers allowing to create nodes from the client.

Did you mean to NOT rly care?

schroeder-:
Sry, you are correct. Meant not really.

PR #1032, opened 2022-09-09
adds security policy Aes128-Sha256-RsaOaep

Kallelele:


schroeder-:
First there are no tests, so we aren't even sure if it works between our client and server.
Also the server part is missing. You need to expand the SecurityPolicyType.

I tired to connect to a server based on Softing SDK and got following error:
```
INFO:asyncua.client.client:connect
INFO:asyncua.client.ua_client.UaClient:opening connection
INFO:asyncua.client.ua_client.UASocketProtocol:open_secure_channel
WARNING:asyncua.uaprotocol:Received an error: ErrorMessage(Error=StatusCode(value=2149842944), Reason=None)
CRITICAL:asyncua.client.ua_client.UASocketProtocol:Received an error: ErrorMessage(Error=StatusCode(value=2149842944), Reason=None)
ERROR:asyncua.client.ua_client.UASocketProtocol:Exception raised while parsing message from server
  File "XXXX\\asyncua\client\ua_client.py", line 173, in _call_callback
    self._callbackmap[request_id].set_result(body)
KeyError: 0

The above exception was the direct cause of the following exception:

  File "XXXX\asyncua\client\ua_client.py", line 81, in _process_received_data
    self._process_received_message(msg)
  File "XXXX\asyncua\client\ua_client.py", line 106, in _process_received_message
    self._call_callback(0, ua.UaStatusCodeError(msg.Error.value))
  File "XXXX\asyncua\client\ua_client.py", line 175, in _call_callback
    raise ua.UaError(f"No request found for request id: {request_id}, pending are {self._callbackmap.keys()}, body was {body}") from ex
asyncua.ua.uaerrors._base.UaError: No request found for request id: 0, pending are dict_keys([1]), body was "The nonce does appear to be not a random value or it is not the correct length."(BadNonceInvalid)
```

schroeder-:
With the following changes a connection to the server was possible:
```
@@ -454,7 +454,7 @@ class SecurityPolicyAes128Sha256RsaOaep(SecurityPolicy):

     URI = "http://opcfoundation.org/UA/SecurityPolicy#Aes128_Sha256_RsaOaep"
     signature_key_size = 32
-    symmetric_key_size = 16
+    symmetric_key_size = 32
     AsymmetricEncryptionURI = "http://www.w3.org/2001/04/xmlenc#rsa-oaep"
     AsymmetricSignatureURI = "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"

@@ -489,7 +489,7 @@ class SecurityPolicyAes128Sha256RsaOaep(SecurityPolicy):
 
     def make_local_symmetric_key(self, secret, seed):
         # specs part 6, 6.7.5
-        key_sizes = (self.signature_key_size, self.symmetric_key_size, 16)
+        key_sizes = (self.signature_key_size, 16, 16)

         (sigkey, key, init_vec) = uacrypto.p_sha256(secret, seed, key_sizes)
         self.symmetric_cryptography.Signer = SignerHMac256(sigkey)
@@ -498,7 +498,7 @@ class SecurityPolicyAes128Sha256RsaOaep(SecurityPolicy):
     def make_remote_symmetric_key(self, secret, seed, lifetime):

         # specs part 6, 6.7.5
-        key_sizes = (self.signature_key_size, self.symmetric_key_size, 16)
+        key_sizes = (self.signature_key_size, 16, 16)

         (sigkey, key, init_vec) = uacrypto.p_sha256(secret, seed, key_sizes)
         if self.symmetric_cryptography.Verifier or self.symmetric_cryptography.Decryptor:
```

Kallelele:
> 

Hmm, the point of this change is to use aes128, for that the key needs to be 16 bytes

schroeder-:
> Hmm, the point of this change is to use aes128, for that the key needs to be 16 bytes

That is correct, but you need a 32 nounce otherwise the server will rejected it.  There for symmetric_key_size is used for the size.

Kallelele:
Ah

Kallelele:
> > Hmm, the point of this change is to use aes128, for that the key needs to be 16 bytes
> 
> That is correct, but you need a 32 nounce otherwise the server will rejected it. There for symmetric_key_size is used for the size.

16 is used for "SecurityPolicyBasic128Rsa15" which also states to use aes128. How does handle the issue you described?

schroeder-:
I think it was just a coincidence that symmetric_key_size was equal to the SecureChannelNonceLength.

Kallelele:
> I think it was just a coincidence that symmetric_key_size was equal to the SecureChannelNonceLength. Best way is to add a secure_channel_nonce_length to all security policys.

Ok, would that be here: https://github.com/FreeOpcUa/opcua-asyncio/blob/4f48eb2df8d998e8f576779fbae8c5c913fddab5/asyncua/common/connection.py#L184
and
https://github.com/FreeOpcUa/opcua-asyncio/blob/b978209820ee3852820d7862b0d91af9b3d15e22/asyncua/client/client.py#L336

schroeder-:
Good job. Works with Softing SDK server and i can access a asyncua example server with Aes128-Sha256-RsaOaep.

Kallelele:
> Good job. Works with Softing SDK server and i can access a asyncua example server with Aes128-Sha256-RsaOaep, with UAExpert.

Do I have to add reviewers to the pull request?

oroulet:
@schroeder- I do not know much the encryption code, so I propose to let you merge that one

Issue #1023, opened 2022-09-07
CVE-2022-25304


schroeder-:
For reference this is CVE-2022-25304 and GHSA-mfpj-3qhm-976m

GitHub already raised an alert in our sample server implementation.

_Originally posted by @GoetzGoerisch in https://github.com/FreeOpcUa/opcua-asyncio/issues/1013#issuecomment-1234160597_

kohtala:
I think this was fixed in #1039.

schroeder-:
No this is in correct. In this dos is because there are no limits on chunk size.
I am working on a fix.

kohtala:
So the CVE mentioned on #1013 is incorrect? If you know of a mistake on an issue, it would be polite to comment about it on that issue and not just create another issue.

schroeder-:
Fixed via #1040

SharonBrizinov:
Thanks @schroeder- ! it looks good

kohtala:
Thank you. Trivy no longer complains with the new version.

PR #1040, opened 2022-09-19
Check limits of messages (CVE-2022-25304)

schroeder-:
Check the limits of messages to prevent DOS from unlimited chunks. (CVE-2022-25304 #1023)

### DIFF
--- a/asyncua/common/connection.py
+++ b/asyncua/common/connection.py
@@ -19,12 +19,12 @@ logger = logging.getLogger('asyncua.uaprotocol')
 @dataclass
 class TransportLimits:
     '''
-        Limits of the server to prevent excessive resource usage
+        Limits of the tcp transport layer to prevent excessive resource usage
     '''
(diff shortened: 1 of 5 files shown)


asyncua/common/connection.py
```diff
@@ -15,6 +16,56 @@ class InvalidSignature(Exception):  # type: ignore
...
 
+@dataclass
+class TransportLimits:
+    '''
+        Limits of the server to prevent excessive resource usage
+    '''
+    max_recv_buffer: int = 65535
+    max_send_buffer: int = 65535
+    max_chunk_count: int = 0
+    max_message_size: int = 0
```
oroulet, line 27: @schroeder- Can't we come up with some default value here that are safer?  a message more than 100MB  should not make any sense. and maybe more 10000 chunks?
Those who want to transfer 1Giga can find that settings

oroulet:
waiting for that on, then making a new  release

PR #1043, opened 2022-09-21
DRAFT: add logging in case of limits error

oroulet:
last update breaks production here. Need to find out what is happening

### DIFF
--- a/asyncua/common/connection.py
+++ b/asyncua/common/connection.py
@@ -21,9 +21,11 @@ class TransportLimits:
     '''
         Limits of the tcp transport layer to prevent excessive resource usage
     '''
+    # Max size of a chunk we can receive
     max_recv_buffer: int = 65535
+    # Max size of a chunk we can send
     max_send_buffer: int = 65535
-    max_chunk_count: int = ((100 * 1024 * 1024) // 65535) + 1 #  max_message_size / max_recv_buffer 
+    max_chunk_count: int = ((100 * 1024 * 1024) // 65535) + 1  # max_message_size / max_recv_buffer
     max_message_size: int = 100 * 1024 * 1024  # 100mb
 
     @staticmethod
@@ -37,12 +39,18 @@ class TransportLimits:
     def check_max_msg_size(self, sz: int) -> bool:
         if self.max_message_size == 0:
             return True
-        return self.max_message_size <= sz
+        within_limit = self.max_message_size <= sz
+        if not within_limit:
+            logger.error("Message size: %s is > configured max message size: %s", sz, self.max_message_size)
+        return within_limit
 
     def check_max_chunk_count(self, sz: int) -> bool:
         if self.max_chunk_count == 0:
             return True
-        return self.max_chunk_count <= sz
+        within_limit = self.max_chunk_count <= sz
+        if not within_limit:
+            logger.error("Number of message chunks: %s is > configured max chunk count: %s", sz, self.max_chunk_count)
+        return within_limit
 
     def create_acknowledge_limits(self, msg: ua.Hello) -> ua.Acknowledge:
         ack = ua.Acknowledge()
@@ -64,6 +72,7 @@ class TransportLimits:
         self.max_recv_buffer = msg.ReceiveBufferSize
         self.max_send_buffer = msg.SendBufferSize
         self.max_message_size = msg.MaxMessageSize
+        logger.warning("updating limits from server: %s", self)
 
 
 class MessageChunk:


schroeder-:
Maybe add this to the test   test_message_limits:
async def test_message_limits(restore_transport_limits_server: Server): 
             ....
            # Test that chunks are working correct 
            async with client:
                 test_string = 'a' * (1024 * 5)
                 n = client.get_node(ua.NodeId())
                 await n.write_value(test_string, ua.VariantType.String)

oroulet:
@schroeder- looks like you had a quite bug bug here. Maybe also one in select_limit. Can you check that PR?
thanks

schroeder-:
My bad should have added that test in the first place. 
Maybe its time to create a project that test our client/server against some other sdk with some real world workloads to quicker gather interopt issues.

oroulet:
I think it is very hard to test all cases anyway

But now I get a ConnectionError instead of timeout in the tests. Any idea why? I cannot manage to find out

oroulet:
I added the test you adviced @schroeder- but it also fails on ConnectionError ?!?!?

schroeder-:
I thought it should be a ConnectionError, but I always got a Timeout error. So changing it should be good.

oroulet:
but I get also connectionErorr in the new test. I cannot find out why

tests/test_server.py
```diff
@@ -682,13 +682,26 @@ async def test_message_limits(restore_transport_limits_server: Server):
...
 
 
+async def test_message_limits_works(restore_transport_limits_server: Server):
+    server = restore_transport_limits_server
+    server.bserver.limits.max_recv_buffer = 1024
+    server.bserver.limits.max_chunk_count = 10
+    client = Client(server.endpoint.geturl())
+    # Test that chunks are working correct
+    async with client:
+        n = client.get_node(ua.NodeId())
```
schroeder-, line 699: You need to create a node or use a existing node with type string

asyncua/common/connection.py
```diff
@@ -64,6 +70,7 @@ def update_limits(self, msg: ua.Acknowledge) -> None:
         self.max_recv_buffer = msg.ReceiveBufferSize
         self.max_send_buffer = msg.SendBufferSize
         self.max_message_size = msg.MaxMessageSize
+        logger.warning("updating limits from server: %s", self)
```
schroeder-, line 73: Maybe make it a info and it is also trigger on the server so removing the server in the text, to prevent confusion.

tests/test_server.py
oroulet: thanks. I did that, but still failing.... I ca'nt rememeber how these tests work...

oroulet:
@schroeder- OK I added more logging and now it looks like it is our server that does not split in chunks the message as the client ask....

oroulet:
OK. I may have found out. Looks like both server and client end up with same send and recv buffer size, while they shoul dbe inverted

WARNING  asyncua.uaprotocol:connection.py:73 updating limits to: TransportLimits(max_recv_buffer=1024, max_send_buffer=65535, max_chunk_count=10, max_message_size=16777216)
WARNING  asyncua.uaprotocol:connection.py:73 updating limits to: TransportLimits(max_recv_buffer=1024, max_send_buffer=65535, max_chunk_count=10, max_message_size=16777216)

oroulet:
that was it, I made the tests pass by settin gthe same value to recv and send. But we still have an invertion issue somewhere

schroeder-:
in connection.py change this:
chunks = MessageChunk.message_to_chunks(
            self.security_policy,
            message,
            self._limits.max_send_buffer,
            message_type=message_type,
            channel_id=self.security_token.ChannelId,
            request_id=request_id,
            token_id=self.security_token.TokenId,
        )

self._limits.max_send_buffer -> min(self._limits.max_send_buffer, self._limits.max_recv_buffer)

because the we should always chose the smaller value to keep the server/client happy and don't violate there limits.

schroeder-:
<html>
<body>


ReceiveBufferSize | UInt32 | The largest MessageChunk that the sender can receive.This value shall not be larger than what the Client requested in the Hello Message.
-- | -- | --
SendBufferSize | UInt32 | The largest MessageChunk that the sender will send.This value shall not be larger than what the Client requested in the Hello Message.


</body>
</html>

oroulet:
everything I fix, shows a new bug !??! so not finished yet

oroulet:
@schroeder- now! it should be OK. Can you review.
There were many fails in that limit stuff, now I added tests and hopefully that works.

oroulet:
also to summarise:
the tests for checking limits had errors,
the negocations of limits were also wrong since we need to invert some values between server and client
assignement of limits had also error in handling 0 value
We were aborting connection as soon as we missed some data, we cannot do that, looks like sometimes we get the data later than expected.

schroeder-:
Looks good for me. Can be merged if tests pass.
Sry for the inconvenience I caused.

oroulet:
of course my changes impact the dos_server test. arrrg

oroulet:
@schroeder- As far as I understand the dos test was jus tsenidng a packet too small. Is that an issue in any situation? worst case we have many hanging connections. We should have a timeout somewhere to handle that, but we cannot simply disconnect everytime some packets comes a bit delayed

schroeder-:
No the dos sends an attack on our internal buffer and trigger out of memory. But you can disable the test in the first step and merge the change. I will then look into fixing the bug in a new pr and after that you can make a new release.

PR #1046, opened 2022-09-22
server: exit on malformed packet

schroeder-:
enable test for malformed header again and add missing return to prevent endless loop.


asyncua/server/binary_server_asyncio.py
```diff
@@ -74,6 +74,7 @@ def data_received(self, data):
                     # malformed header prevent invalid access of your buffer
```
oroulet, line 74: what is that test over?  it jsut means that body size i 0. Can't it happen?

schroeder-: Against a normal client no. This is only if an attack sends a special crafted packet. See #1013

PR #1048, opened 2022-09-23
review all big exception handlings and try to better follow best practices

oroulet:


tests/test_crypto_connect.py
```diff
@@ -300,7 +300,6 @@ async def test_certificate_handling_mismatched_creds(srv_crypto_one_cert):
             assert await clt.get_objects_node().get_children()
 
 
-@pytest.mark.skip("That test fails and hangs randomly on github so disabling for now")
 async def test_secure_channel_key_expiration(srv_crypto_one_cert, mocker):
     timeout = 1
```
schroeder-, line 304: Maybe we should make the timeout longer. Because I can reproduce the failing test on my environment if I reduce the timeout to 0.1 seconds. And the CI server is a lot more resource constraint and also this would explain why this error appeared after adding the new watchdog task.

oroulet: That test hangs completely in some situations. Looks like Task.cancel() does not  cancel your monitoring task in some situations. I am trying to understand that one

oroulet: Copy of a message i wrote at work: 

anyone interested in the dark side of python programming. I think I found the issue of  tests hanging sometimes
We are back to Task.cancel() not cancelling a task in some conditions. But I still do not exactly in what conditions...

It is possible to reproduce in asyncua with that test: https://github.com/FreeOpcUa/opcua-asyncio/pull/1048/files#diff-83dd4ff683bb84890ba88ddacd4b01524ea60dc44e2f8f411a5a8a03a1ea93baL304

using a command line like
 cpulimit -l 20 -f -- pytest -v -s --count 30 -k channel_key_expirat
We hang forever in some cases there: https://github.com/FreeOpcUa/opcua-asyncio/blob/hang/asyncua/client/client.py#L587  because the task is not cancelled as expected.
the task code is there: https://github.com/FreeOpcUa/opcua-asyncio/blob/hang/asyncua/client/client.py#L446



changing code like
 task.cancel()
 await task
to something like
 task.cancel()
 await asyncio.sleep(0)
 await task
reduces the probability of the issue a lot but not completely

I could not reproduce with a code like this
 task.cancel()
 await asyncio.sleep(0)
 task.cancel()
 await asyncio.sleep(0)
 await task
but that is too hacky

It might be related to: https://stackoverflow.com/questions/68759284/how-to-make-asyncio-cancel-to-actually-cancel-the-task

Task.cancel() docs say cancelling does not work if we catch it .I tripple check the code and I do not see any catch (CancelledErrror is aBaseException so catching Exception is fine)
It also says it needs some Python contect switch (not exactly sure what context switch is in python) but it looks to need more than one context switch to be 100% sure the cancelling happens.. https://docs.python.org/3/library/asyncio-task.html?highlight=cancel#asyncio.Task.cancel
Someone else has any idea?

schroeder-: I think task cancel is not the problem here. This is what the test wants:
T -> test task
S -> renew secure channel task

T: connect to server
T: wait for **timeout** seconds
S: renew secure_channel (after 0.7 * **timeout** seconds)
T: check if secure_channel is renewed

this is what happends on the CI because of limit cpu time:
T: connect to server
T: wait for **timeout** seconds
T: check if secure_channel is renewed, this is not the case so the test fails.
 
the renew secure channel task should be trigger after 70% of the **timeout** value.

schroeder-: #1050 should fix the test. Add a timestamp to prevent a too long delay, when scheduling the secure_channel task needs longer than expected.

oroulet: I do not know if there are other issues but i am 100% sure the task over does not always get into cancelled state after calling cancel()
There are many ways to solve it in a hacky way but i would really like to understand how it that happens

oroulet: I but yes i also noticed that sometimes the test fail.
What i am mainly concerned about is that it sometimes hangs

schroeder-: Ok the hang thing is a separate issue. I saw that too, but I am not a expert in all this asyncio things.

oroulet: Both thimgs seems to happen independently btw

oroulet: shit seems to be a python bug: https://stackoverflow.com/questions/70759235/confusing-asyncio-task-cancellation-behavior

we use wait_for in a task that can be cancelled and sometime it swallows it

PR #1050, opened 2022-09-24
fix renew channel test

schroeder-:


asyncua/client/client.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import timeit
```
oroulet, line 3: as far as I know timeit is for timing code. not to be used for timestamp. I usually use time.time() but I see now that it is not montonic so time.monotonic_ns() sounds like the correct choice

asyncua/client/client.py
```diff
@@ -459,22 +461,26 @@ async def _monitor_server_loop(self):
...
+        end = timeit.timeit()
+        ajust_time = end - connect_time
         try:
+
             # Part4 5.5.2.1:
             # Clients should request a new SecurityToken after 75 % of its lifetime has elapsed
             duration = self.secure_channel_timeout * 0.75 / 1000
             while True:
-                await asyncio.sleep(duration)
+                await asyncio.sleep(duration - ajust_time)
```
oroulet, line 478: so that code is only to fix the cases when we have crazy low secure_channel_timeout? so basically only for testing? If so it is not worth it, better modify the tests

schroeder-:
I close this as doubling the timeout seams enough.

PR #1057, opened 2022-09-28
Client: inform all subscription handler about connection lose

schroeder-:
If the watchdog task gets an error all subscription handler with **status_change_notification** method will be called with BadShutdown. 
Also only prints an info message if the connection is lost.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -454,8 +454,13 @@ class Client:
                 await asyncio.sleep(timeout)
                 # @FIXME handle state change
                 _ = await self.nodes.server_state.read_value()
+        except ConnectionError as e:
+            _logger.info("connection error in watchdog loop %s", e, exc_info=True)
+            await self.uaclient.inform_subscriptions(ua.StatusCodes.BadShutdown)
+            raise
         except Exception:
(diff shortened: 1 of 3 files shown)


oroulet:
I am wondering if we could add the same thing to clients subscribed to data changes. Ideally would the handler be called immediately with something raising an exception. But we should avoid breaking API if we can... or we break it and release 1.0. now that we have some safety in system it might be worth releasing 1.0 and it will make following correct versioning scheme easier.
We could add an error argument to datachange_notification which is None by default for example, although I am not that keen in that souiton

schroeder-:
I don't know because we already split between datachange and event. So I thought using status_change_notification can be used. 
Or as an alternative we could use a new method to the handler:
```
def on_connection_state(self, state):
     .....
```
Where state can be a statuscode or our own enum. Like Connected, Disconnected, SubscriptionCreated, SubscriptionDeleted.
This could be used later when we implement a automatic reconnect to inform the clients that we are reconnecting.

oroulet:
or simply add on_connect() and on_disconnect() or on_connection_lost() to the Handler?
Not sure what is best
if we have only one method then `on_connection_event(event)`is probably a better name

PR #1065, opened 2022-10-03
update README

oroulet:


oroulet:
@schroeder- tests failed. At least they did not hang ;-). Is it that timing issue again?

schroeder-:
Yes, I had to rerun a test another pr. Maybe make the timeout a bit longer.

PR #1066, opened 2022-10-04
Fix typo in in XmlImporte._check_if_namespace_meta_information_is_add ed which results in an exception

bitkeeper:
Fix typo in in `XmlImporte._check_if_namespace_meta_information_is_add` ed which results in an exception.

In case it was required after the import to still add an ns an exception occured. On normal flow the line `_check_if_namespace_meta_information_is_added`s  with the typo isn't executed and no test was available to test the behaviour.

* Added test for `_check_if_namespace_meta_information_is_added`
* Fix the typo:
```python
obj = await self.nodes.namespaces.add_object(idx, uri, ua.ObjectIds.NamespaceMetadataType, False)
```
should be:
```python
obj = await self.server.nodes.namespaces.add_object(idx, uri, ua.ObjectIds.NamespaceMetadataType, False)
```

### DIFF
--- a/tests/test_xml.py
+++ b/tests/test_xml.py
@@ -23,6 +23,7 @@ CUSTOM_NODES_NS_XML_PATH = BASE_DIR / "custom_nodesns.xml"
 CUSTOM_NODES_NS_XML_PATH1 = BASE_DIR / "custom_nodesns_2.xml"
 CUSTOM_NODES_NS_XML_PATH2 = BASE_DIR / "custom_nodesns_3.xml"
 CUSTOM_NODES_NS_XML_PATH3 = BASE_DIR / "custom_nodesns_4.xml"
+CUSTOM_NS_META_ADD_XML_PATH = BASE_DIR / "custom_ns_meta_add.xml"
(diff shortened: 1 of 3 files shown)


tests/test_xml.py
```diff
@@ -592,3 +593,9 @@ async def test_disable_xml_export_without_value(opc, tmpdir):
     assert dv.Value != v.Value
     assert v.Value.Value is None
     await opc.opc.delete_nodes([o2])
+
+async def test_xml_namespace_meta_add(opc):
+    with pytest.raises(ValueError):
+        await opc.opc.get_namespace_index("http://foobar.org/struct_optional/")
+    await opc.opc.import_xml(CUSTOM_NS_META_ADD_XML_PATH)
+    assert await opc.opc.get_namespace_index("http://foobar.org/struct_optional/") == 2
```
schroeder-, line 601: Change that to:
`assert await opc.opc.get_namespace_index("http://foobar.org/struct_optional/") > 0`
because if it doesn't raise a exception it should be ok.

bitkeeper: Done

PR #1083, opened 2022-10-14
improved exporting and importing extension objects to/from xml

smorokin:
I have improved the export and import of extension objects to xml. Currently only tested with a b&r plc. 

While testing I found that the the following entries are missing in 'ua.object_ids.py':
```python
class ObjectIds:
    # ...
    ProgramDiagnostic2DataType = 15396
    # ...
ObjectIdNames[15396] = 'ProgramDiagnostic2DataType'
```

I can't find how  `object_ids.py` is generated, so I could not fix it.

### DIFF
--- a/asyncua/common/utils.py
+++ b/asyncua/common/utils.py
@@ -120,7 +120,7 @@ def fields_with_resolved_types(
         try:
             field.type = resolved_fieldtypes[field.name]
         except KeyError:
-            _logger.info(f"could not resolve fieldtype for {field=} of {class_or_instance=}")
+            _logger.info(f"could not resolve fieldtype for field={field} of class_or_instance={class_or_instance}")
             pass
 
     return fields_
--- a/asyncua/common/utils.py
+++ b/asyncua/common/utils.py
@@ -5,6 +5,7 @@ Helper function and classes depending on ua object are in ua_utils.py
 
 import os
 import logging
+import sys
 from dataclasses import Field, fields
 from typing import get_type_hints, Dict, Tuple, Any, Optional
 from ..ua.uaerrors import UaError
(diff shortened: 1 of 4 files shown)


AndreasHeine:
> I have improved the export and import of extension objects to xml. Currently only tested with a b&r plc. 
> 
> While testing I found that the the following entries are missing in 'ua.object_ids.py':
> ```python
> class ObjectIds:
>     # ...
>     ProgramDiagnostic2DataType = 15396
>     # ...
> ObjectIdNames[15396] = 'ProgramDiagnostic2DataType'
> ```
> 
> I can't find how  `object_ids.py` is generated, so I could not fix it. 

https://github.com/FreeOpcUa/opcua-asyncio/tree/master/schemas

schroeder-:
I don't know if this patch is a good idea. Because it expands the xmlimport/export madness, because it just implements random bits from the xml data encoding. Instead we should implement the whole xml data encoding (https://reference.opcfoundation.org/v104/Core/docs/Part6/5.3.1/) and implement the same as uabinary.py.

Also this patch is missing any test, so we don't know if it works or will break in the future.

smorokin:
@AndreasHeine : Thank you, I had a better look and it seems I have this problem because my plc has a different version of the default nodeset than this library. I will handle this in my client code manually.

smorokin:
@schroeder- : Yes, it might be a better idea to create a `uaxml.py` and use that for the export / import, but in the meantime this might help some people (might also be a starting point for `uaxml.py`). I can add a test, but first I have to "downgrade" the changes to python 3.7.

smorokin:
@schroeder- : I have added 2 test cases. Do you want me to add something else here?

One thing I struggled with: I had problems because one of my tests had created a struct with the exact same name as other tests (`MyStruct`). Is this known (normally test cases should be independent).

schroeder-:
The test should be sufficient.
The tests aren't independent they all use the same server to reduce the time needed to execute the tests. Otherwise we would waste a lot of time init a new server every test.

smorokin:
Thank you for the quick reply. 
In hindsight I should have thought of this, but while writing them I assumed there was a weird bug with my test and python3.7. Nevermind, I know now.

oroulet:
should I merge @schroeder- ? I do not follow much that xml stuff. I do not use it

schroeder-:
I think its a good to merge.

smorokin:
Thank you!

PR #1087, opened 2022-10-17
add feat: subscription.py optional param for sampling_interval

AndreasHeine:


AndreasHeine:
@oroulet tested against a "Unified Automation Cpp Server"!

asyncua/common/subscription.py
```diff
@@ -171,12 +177,14 @@ async def _call_status(self, status: ua.StatusChangeNotification):
...
-                                    monitoring=ua.MonitoringMode.Reporting,
-                                   ) -> Union[int, List[Union[int, ua.StatusCode]]]:
+    async def subscribe_data_change(
+        self,
+        nodes: Union[Node, Iterable[Node]],
+        attr: ua.AttributeIds = ua.AttributeIds.Value,
+        queuesize: int = 0,
+        monitoring=ua.MonitoringMode.Reporting,
+        sampling_interval: ua.Duration = 0.0
+    ) -> Union[int, List[Union[int, ua.StatusCode]]]:
```
oroulet, line 187: ideally we should pass nodes AND the entire MonitoredItemCreateRequest so people can add whatever they want. but that breaks the API....

oroulet:
your PR is so hard to read because you have mixed up many formatting changes and typing with real code changes...

asyncua/common/subscription.py
oroulet: but create_monitored_items is hard to use, it does not any handler or anything, so you need to catch events by hand (I do not even remember how tot do that)
Anyway. let's do it the way you propose for now

Issue #1084, opened 2022-10-15
Datatypes marked isAbstract = True for datatypes that should not be abstract


tjchicoine:
**Describe the bug** <br />
For example : [BaseDataVariableType](https://reference.opcfoundation.org/Core/VariableTypes/v104/BaseDataVariableType/) should not be abstract, but is found to have isAbstract = True upon inspection with UAExpert. 

**To Reproduce**<br />
Run asyncio example server and observe the isAbstract attribute of BaseDataVariableType.

**Expected behavior**<br />
This node should not be marked abstract. If you explicitly set isAbstract = False in standard_address_space_services.py the problem is resolved. 

**Version**<br />
Python-Version: 3.9.1
opcua-asyncio Version 0.9.95

tjchicoine:
I fixed it - will make a pull request soon

PR #1089, opened 2022-10-18
isAbstract default to False

tjchicoine:
#1084


oroulet:
Read the first lines of that file ;-)

tjchicoine:
Is it a better idea to add explict declaration of abstract = false to the autogenerating code of standard_address_space_services?

schroeder-:
Yes changing adresspace.py is the goal. But we need to adapt schema/generate_standard_addresspace.py or you wait until if have time to fix it.

PR #1090, opened 2022-10-18
proposal-for-better-naming

AndreasHeine:


oroulet:
As good as it can be i think

AndreasHeine:
@oroulet this is just a proposal for a clean PR afterwards!

not sure exactly why it is like it is but for my understanding the naming "server" in the nodeclass-constructor is horribly misleading!

what i would like to change is that the "UaClient" and the "InternalSession" inherit from an ABC (AbstractSession) which has all the required method's for a OPC UA Session (Read/Write/Browse/...) in order to make the Typing more clear and cleaner

i first renamed "server" to "session" which is in my opinion the proper name.

from

```python
def __init__(self, session: Union[InternalSession, UaClient], nodeid: Union[ua.NodeId, str, int]):
```

to

```python
def __init__(self, session: AbstractSession, nodeid: Union[ua.NodeId, str, int]):
```

feedback welcome!

oroulet:
Are they so many methods that are the same between client and server?if so an abstract class is fine

schroeder-:
Like the renaming, make sense to me.

AndreasHeine:
ABC is just for keeping it sync and visible for both cases

schroeder-:
Maybe use typing protocols instead of abc: https://mypy.readthedocs.io/en/stable/protocols.html . But its only available in python >= 3.8 or we introduce typing-extensions, which  would allow us to remove some typing hacks for Python 3.7.

schroeder-:
No but library typing-extensions backport it to 3.7 so we could use  it.

oroulet:
If there is no code what is the advantage of having that abstract class? Interfaces are fines when you want to be able to replace one by another but in that case nobody will replace a server by a client. They just have a similar api for convenience

AndreasHeine:
need to look into viewservice and noderegister if it makes sense to implement... for now just an error!
but the ABC exposes that node-class on serverside is compromised with missing methods!!!

AndreasHeine:
@oroulet we are currently not supporting "BrowseNextRequest_Encoding_DefaultBinary" in uaprocessor!?
image

AndreasHeine:
node_register is not implemented as well so i can just straight return the provided nodeids ^^

```python
elif typeid == ua.NodeId(ua.ObjectIds.RegisterNodesRequest_Encoding_DefaultBinary):
    _logger.info("register nodes request (%s)", user)
    params = struct_from_binary(ua.RegisterNodesParameters, body)
    _logger.info("Node registration not implemented")
    response = ua.RegisterNodesResponse()
    response.Parameters.RegisteredNodeIds = params.NodesToRegister
    # _logger.info("sending register nodes response")
    self.send_response(requesthdr.RequestHandle, seqhdr, response)
```

asyncua/common/session_interface.py
```diff
@@ -0,0 +1,226 @@
...
+    # View Service Set: https://reference.opcfoundation.org/Core/Part4/v104/5.8.1/
+
+    @abstractmethod
+    async def browse(self, parameters: ua.BrowseParameters) -> List[ua.BrowseResult]:
+        '''
+        https://reference.opcfoundation.org/Core/Part4/v104/5.8.2/
+
+        This Service is used to discover the References of a specified Node. 
+        The browse can be further limited by the use of a View. 
+        This Browse Service also supports a primitive filtering capability.
```
oroulet, line 23: What was the point again of these interface things?
I am really not keen in that kind of empty interfaces. but if you all love them....

AndreasHeine: Node-Class can be used from client- and server-side but there was no interface for there commonalities, so the Node-Class has issues from server-side because it lacks methods so Node-Class will have exceptions if you call them from server-side... the ABC is just the method-gathering which both have to implement, otherwise it will crash (exception) on class construction (what is great!) and not if you accidentally use them!

GoetzGoerisch:
@AndreasHeine  just for information, as I happend to me also in the past.

> stricter separation of Connection and Session so the connection can be reestablished without interfering the session and in the ClientConnection could have a queue for session requests in case of reconnecting... as well as future sessionless services (https://reference.opcfoundation.org/Core/Part4/v104/6.3.1/) which woulld take place in UaClient aswell... another thing is that a client could have multiple sessions with different identities via one connection but that's a different topic...

Always look on the latest spec version:
[https://reference.opcfoundation.org/Core/Part4/**v105**/6.3.1/](https://reference.opcfoundation.org/Core/Part4/v105/6.3.1/)

schroeder-:
@AndreasHeine go ahead, I noticed the missing bounderies between session, secure channel and connection. Which need to be cleaned up anyway sometime, when we are going for session recovery, which is a long road ahead.

AndreasHeine:
> @AndreasHeine just for information, as I happend to me also in the past.
> 
> > stricter separation of Connection and Session so the connection can be reestablished without interfering the session and in the ClientConnection could have a queue for session requests in case of reconnecting... as well as future sessionless services (https://reference.opcfoundation.org/Core/Part4/v104/6.3.1/) which woulld take place in UaClient aswell... another thing is that a client could have multiple sessions with different identities via one connection but that's a different topic...
> 
> Always look on the latest spec version: [https://reference.opcfoundation.org/Core/Part4/**v105**/6.3.1/](https://reference.opcfoundation.org/Core/Part4/v105/6.3.1/)

luckily the have not changed something important ^^

oroulet:
> > @AndreasHeine just for information, as I happend to me also in the past.
> > > stricter separation of Connection and Session so the connection can be reestablished without interfering the session and in the ClientConnection could have a queue for session requests in case of reconnecting... as well as future sessionless services (https://reference.opcfoundation.org/Core/Part4/v104/6.3.1/) which woulld take place in UaClient aswell... another thing is that a client could have multiple sessions with different identities via one connection but that's a different topic...
> > 
> > 
> > Always look on the latest spec version: [https://reference.opcfoundation.org/Core/Part4/**v105**/6.3.1/](https://reference.opcfoundation.org/Core/Part4/v105/6.3.1/)
> 
> luckily the have not changed something important ^^

donæt they have a link to latest spec?

asyncua/client/ua_client.py
```diff
@@ -775,3 +778,9 @@ async def set_publishing_mode(self, params) -> ua.uatypes.StatusCode:
         self.logger.debug(response)
         response.ResponseHeader.ServiceResult.check()
         return response.Parameters.Results
+
+    async def transfer_subscriptions(self, params: ua.TransferSubscriptionsParameters) -> List[ua.TransferResult]:
+        # Subscriptions aren't bound to a Session and can be transfered!
+        # https://reference.opcfoundation.org/Core/Part4/v104/5.13.7/
+        self.logger.info("TransferSubscriptions has not been implemented yet!")
```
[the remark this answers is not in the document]
HarryZ10: @oroulet Is there a current push to get this part implemented?

PR #1096, opened 2022-10-24
proposal-for-Roles-and-Permissions

AndreasHeine:
feedback or volunteers welcome ;)

### DIFF
--- a/asyncua/server/users.py
+++ b/asyncua/server/users.py
@@ -7,6 +7,36 @@ from dataclasses import dataclass
 from typing import Optional
 
 
+'''
+
+TODO for Roles and Permissions:
+
+https://reference.opcfoundation.org/v104/Core/docs/Part3/4.8.2/
+
+Anonymous 	The Role has very limited access for use when a Session has anonymous credentials.
+AuthenticatedUser 	The Role has limited access for use when a Session has valid non-anonymous credentials but has not been explicitly granted access to a Role.
+Observer 	The Role is allowed to browse, read live data, read historical data/events or subscribe to data/events.
+Operator 	The Role is allowed to browse, read live data, read historical data/events or subscribe to data/events.
+In addition, the Session is allowed to write some live data and call some Methods.
+Engineer 	The Role is allowed to browse, read/write configuration data, read historical data/events, call Methods or subscribe to data/events.
+Supervisor 	The Role is allowed to browse, read live data, read historical data/events, call Methods or subscribe to data/events.
+ConfigureAdmin 	The Role is allowed to change the non-security related configuration settings.
+SecurityAdmin 	The Role is allowed to change security related settings.
+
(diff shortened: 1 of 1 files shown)


schroeder-:
I don't think we should limit everything to well known roles. Because this limits the advantages of the role model, as it allows to combine multiple roles for one sessions. This allows very fine grain access to nodes. 
I use an commercial C++ SDK, everything is handled like this:
1. You create multiple RoleType objects (NodeId is later used as reference) like for WriteTestNodes, AdminFileAccess, ..usw.
2. In the step of Authentication you add 1 or more Roles to the current session. So every session carries a list of roles which can be modified in the running sessions. Every service checks against the list of roles for access.

oroulet:
The main issue is well to implement the lower level permission like AdminFileAccess.. That will require a lot of if then else everywhere . I do not even want to think about it

schroeder-:
No the roles like AdminFileAccess are user defined, for their own purpose. The user adds to the nodes the corresponding rolepermissions and if no rolespermissions are definied the parent/default permissions are used.

AndreasHeine:
if i gor you right:
the User will get a list of roles
and you create RolePermissions (Role + Permissions) e.g. AdminFileAccess
which will be used in the NodeAttribute RolePermission/UserRolePermissions of lets say all FileTypeNodes!?

schroeder-:
I did an example for an api:
```Python
class RoleAuth:
	def on_init_roles(self, server):
		self.file_access = server.add_role(NodeId(4, 'RoleFileAccess'), 'FileAccess')
		self.write_access = server.add_role(NodeId(4, 'RoleWriteAccess'), 'WriteAccess')
		self.read_access = server.add_role(NodeId(4, 'RoleReadAccess'), 'ReadAccess')

	def on_auth(self, user: str, server: Server) -> List[ua.Node]:
		# Called when user is authenticated
		if user == "User1":
			return [file_access, read_access]
		if user == "User2":
			return [write_access]
		if user == "Admin":
			return [file_access, read_access, write_access, server.roles.admin]
		return []

auth = RoleAuth()
server = Server(..., role_auth=auth)
auth.on_init_roles(server)
...


var = server.add_variable(....)			
var.add_rolepermission(auth.read_access, PermissionType.Browse | PermissionType.Read | PermissionType.ReceiveEvents)
var.add_rolepermission(auth.write_access, PermissionType.Browse | PermissionType.Read | PermissionType.Write | PermissionType.ReceiveEvents)
file = server.add_object(....)
var.add_rolepermission(auth.file_access, PermissionType.Browse | PermissionType.Read | PermissionType.Call)
```

AndreasHeine:
ok got yah!

i have used it with another open source stack but only for Well Knowns Roles:
image

what is see in your code example:
is it a good idea to link a username to a set/list of RolePermissions?
i would link the Role of that session to the RolePermissions!?

just to seperate usermanagement (usernames and passwords) from internal stuff.

PR #1097, opened 2022-10-25
Change autogeneration of status codes to use enums

glomium:
I've changed the autogeneration script for the status codes to use enums. For good measurement, I've also included the additional property  `is_good` and `doc` according to `ua.uatypes` in the Enum. The mask for `is_good` is precompiled.

To allow comparison with integers I've used `IntEnum` to break nothing. `ua.status_codes` is auto-generated and additionally `ua.uatypes` needed to be modified.

I'm not sure why `doc = doc.replace("'", '"')` was included in the old generation of the documentation strings. I've removed the line and also the generation of additional quotation marks.


oroulet:
looks good. But what was the motivation? just code cleanup?

asyncua/ua/uatypes.py
```diff
@@ -318,7 +318,9 @@ class StatusCode:
 
     def __post_init__(self):
         if isinstance(self.value, str):
-            object.__setattr__(self, "value", getattr(status_codes.StatusCodes, self.value))
+            object.__setattr__(self, "value", getattr(status_codes.StatusCodes, self.value).value)
+        elif isinstance(self.value, status_codes.StatusCodes):
+            object.__setattr__(self, "value", self.value.value)
```
oroulet, line 323: Can we change type of Value to be `status_codes.StatusCodes`?
Since it is an int enum it should not need any other change

oroulet:
I agree that having both Statuscode and StatusCodes is confusing. I am not sure why it is like it now.
But merging StatusCode and StatusCodes might not be that hard.

you can autogenerate StatusCode like you do now with StatusCodes.
then make StatusCodes an alias to SatusCode
Then override constructor to make sure constructions like StatusCode(StatusCodes.BadXXX) do not fail

And is_good() method  and doc() property

asyncua/ua/uatypes.py
glomium: There seems to be at least one comparison, where the value attribute is compared with lt or gt, which is not supported by enums.

glomium:
I've pushed another commit :)

* `schemas/generate_statuscode.py` creates a file `_generated.py` that only contains the staus-codes and their names as a list as well as a dictionary with the documentation.
* The generated objects from `_generated.py` are used by `status_codes.py` to create the enum. Thus, a child of IntEnum is created with the discussed properties and attributes. In addition, it has a new `default` function and a `_missing_`, which is used on creation if a status code is not part of the enum (a function the currently used implementation uses).
* I've added a `DeprecationWarning` to `get_name_and_doc` and updated its only left usage in `uaerrors/_base.py`
* `uatypes.py` was updated to use the functions of the newly added enum. I've also included better `__str__` and `__repr__` functions.

You could use `from asyncua.ua.status_codes import StatusCodes as StatusCode` as a replacement, however, you can't call `StatusCode()` without an argument - the enum forbids this (and it's not documented how this behavior can be changed). Thus the `default` function was added. It can be used in the `default_factory` of dataclass-objects and where `StatusCodes` is called without an argument.

If these changes are fine with you, I will prepare another commit that emits a `DeprecationWarning`  if `StatusCode` is called without an argument and apply the previously proposed refactoring.

oroulet:
I am a bit confused here. What is the goal of that last commit? you create a StatusCodes enum that does everything that StatusCode does but keep both for compatibiity?

oroulet:
if so we should at least change internal code to not use StatusCode class

oroulet:
@glomium ? also you need to rebase

oroulet:
@glomium ?

Issue #1098, opened 2022-10-26
Server ignores DataChangeFilter trigger value


JohanAR:
**Describe the bug** <br />
Our client creates a DataChangeFilter with Trigger set to OpcUa_DataChangeTrigger_StatusValueTimestamp, but it seems like opcua-asyncio server ignores this and still only sends datachange notifications when status or value changes.

**To Reproduce**<br />

```
        /* Sample code for creation of data change filter */
        OpcUa_DataChangeFilter* pDataChangeFilter = NULL;
        OpcUa_EncodeableObject_CreateExtension(
            &OpcUa_DataChangeFilter_EncodeableType,
            &createRequest[i].RequestedParameters.Filter,
            (OpcUa_Void**)&pDataChangeFilter);
        if (pDataChangeFilter)
        {
            pDataChangeFilter->DeadbandType = OpcUa_DeadbandType_None;
            pDataChangeFilter->DeadbandValue = 0.0;
            pDataChangeFilter->Trigger = OpcUa_DataChangeTrigger_StatusValueTimestamp;
        }    
```    

**Expected behavior**<br />
DataChange notification when write_attribute_value is called with a new timestamp, even if value and status are unchanged.

**Screenshots**<br />
I tried to follow the logic in the server, and it looks like server.write_attribute_value eventually reaches the following lines in address_space.py:
```
        # only send call callback when a value or status code change has happened
        if (old.Value != value.Value) or (old.StatusCode != value.StatusCode):
            cbs = list(attval.datachange_callbacks.items())
```
Just guessing, but maybe it's possible to remove this hardcoded check, and instead do a test in monitored_item_service.py based on the Trigger value?

**Version**<br />
Python-Version:<br />
opcua-asyncio Version (e.g. master branch, 0.9):

JohanAR:
I've implemented this and it looks like it's working with our client. I'll submit a PR if the existing tests are passing but I don't have time to write any new tests now.

PR #1099, opened 2022-10-27
Respect DataChangeTrigger in server

JohanAR:
Moves DataChange filtering from address_space to monitored_item_service, and uses the configured DataChangeTrigger instead of always detecting changes in status/value.

Solves #1098

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -779,14 +779,9 @@ class AddressSpace:
             # Only check datatype if no bad StatusCode is set
             return ua.StatusCode(ua.StatusCodes.BadTypeMismatch)
 
-        old = attval.value
         attval.value = value
-        cbs = []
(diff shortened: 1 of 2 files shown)


oroulet:
looks good and does not seem to break anything. jsut waiting a bit in case someone finds something

JohanAR:
I wish I had time to write proper tests for it, but we have a deadline coming up and team lead won't let me spend any more time on it. I have tested setting the DataChangeTrigger to both StatusValue and StatusValueTimestamp, and leaving it unset, using our client and test code. Everything looks like it's working, but I'm far from an OpcUa expert so I'm a little worried about breaking something that I don't know about. I hope your existing tests have good coverage :)

I might be able to return to this and write those tests in a couple of weeks, so it doesn't break in the future, I hope.

PR #1103, opened 2022-10-28
support python 3.11

oroulet:


oroulet:
@schroeder- any ideas what is happening here?
I fixed an obvious issue that shoul dave failed before 3.11 but now that error is a bit strange. I have no ideas how that code could have worked before. Maybe they added some check at dataclass definition. it would require to add reordering of dependencies when importting and I really do not feel like doing that...

oroulet:
I think I will jsut disable that test for now

asyncua/common/structures104.py
```diff
@@ -157,7 +157,7 @@ def get_default_value(uatype, enums=None):
         # We have an enum, try to initilize it correctly
         val = list(getattr(ua, uatype).__members__)[0]
         return f"ua.{uatype}.{val}"
-    return f"ua.{uatype}()"
+    return f"field(default_factory=ua.{uatype})"
```
schroeder-, line 160: This changes breaks the old struct code. 
The problem is that the old code registers the variables to ua namespace after executing the codes in another step. 
This would require some reworking of the old struct code.

schroeder-:
You can merge and I try the change the old struct code after that.

asyncua/common/structures104.py
oroulet: I do not understand that one. That change should only change the way the dataclass is initiliazed. How does it change the registratoin to the ua namespace?

oroulet:
> You can merge and I try the change the old struct code after that.

but currently I have 0 idea why the old code is broken

asyncua/common/structures104.py
schroeder-: The old code first creates all classes and after that they are registered to the ua namespace. So if they are used in the default_factory that will fail.

schroeder-: If you change the line back the test will pass.

oroulet: arrgg. OK now I remember I found out that yesterday, but then you cannot simply register because you also need to reorder dependencies and that is not done in that code... Not sure what to do. either we do not support 3.11 or we break old structure support. Or we need someone to implement that reodering because I really do not have time

oroulet:
OK. I have one absolutely horrible bypass:
 `    return f"field(default_factory=lambda :ua.{uatype})"
`
that makes everything work. probably by bypassing some new python 3.11 checks

oroulet:
let's try that one

schroeder-:
I think that hack is actually the best way without reordering the whole old struct code.

PR #1104, opened 2022-10-28
Only set value to Null if status code is bad

### LINKED ISSUE #1102: server treats uncertain as error, and sends null instead of value
**Describe the bug** <br />
Values with status code uncertain are replaced with null values

**To Reproduce**<br />
call write_attribute_value with a value and status code for example UncertainSensorCalibration

**Expected behavior**<br />
I expect to get a data change notification containing the value and status code that I wrote.

**Version**<br />
Python-Version: 3.9<br />
opcua-asyncio Version: 1.0

The code refers to https://reference.opcfoundation.org/v104/Core/docs/Part4/7.7.1/ which says:
> The data value. If the StatusCode indicates an error then the value is to be ignored and the Server shall set it to null.

But https://reference.opcfoundation.org/v104/Core/docs/Part4/7.34.1/ says:
> Good Success   00   Indicates that the operation was successful and the associated results may be used.
   Uncertain Warning   01   Indicates that the operation was partially successful and that associated results might not be suitable for some purposes.
   Bad Failure   10   Indicates that the operation failed and any associated results cannot be used.

Which I interpret as that only "bad failure" statuses should be considered errors.

I propose changing "not value.StatusCode.is_good()" to "value.StatusCode.is_bad()", unless someone who knows more about OpcUa than me can clarify the situation.

JohanAR:
Previously it also did so for uncertain status code.

Fixes #1102

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -771,7 +771,7 @@ class AddressSpace:
         attval = node.attributes.get(attr, None)
         if attval is None:
             return ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid)
-        if value.StatusCode is not None and not value.StatusCode.is_good():
+        if value.StatusCode is not None and value.StatusCode.is_bad():
             # https://reference.opcfoundation.org/v104/Core/docs/Part4/7.7.1/
             # If the StatusCode indicates an error then the value is to be ignored and the Server shall set it to null.
             value = dataclasses.replace(value, Value=ua.Variant(ua.Null(), ua.VariantType.Null))
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -329,12 +329,33 @@ class StatusCode:
 
     def is_good(self):
         """
-        return True if status is Good.
+        return True if status is Good (00).
         """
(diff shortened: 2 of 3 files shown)


JohanAR:
>  Reserved   11   Reserved for future use. All Clients should treat a StatusCode with this severity as “Bad”.

Maybe is_bad() should return true for this too?

oroulet:
yes I think so

JohanAR:
Updated the code.

It would be more efficient to only check bit 31 but it felt slightly more readable, and more consistent with is_good and is_uncertain, to mask out bits 30-31 and then compare them with both bad and reserved.

AndreasHeine:
@oroulet @JohanAR i am confused by the title a variant can be empty at any time -> https://reference.opcfoundation.org/Core/Part6/v104/5.1.6/

the statuscode comes only into play in context of a datavalue!

JohanAR:
Tried to clarify the commit message

JohanAR:
No idea why the github checks failed now, I only updated the commit message. Is it possible to re-trigger them somehow?

oroulet:
That test fails sometimes. I thought we had fixed it ....

PR #1111, opened 2022-11-03
refractored add references functions to greatly improve xml import speed.

jcbastosportela:
With this changes I was able to import a set of XMLs in about 35s in my machine. Without this changes the same set of inf models won't finish importing even after 5 minutes!

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -33,6 +33,13 @@ from .users import User, UserRole
 _logger = logging.getLogger(__name__)
 
 
+# usefull const NodeIds: avoid constructing multiple times for comparsion purpose
+NODE_ID_HAS_PROPERTY = ua.NodeId(ua.ObjectIds.HasProperty)
+NODE_ID_NULL = ua.NodeId(ua.ObjectIds.Null)
+NODE_ID_HASSUBTYPE = ua.NodeId(ua.ObjectIds.HasSubtype)
+NODE_ID_DEFAULT = ua.NodeId()
+
+
 class AttributeValue(object):
     """
     The class holds the value(s) of an attribute and callbacks.
@@ -55,7 +62,7 @@ class NodeData:
     def __init__(self, nodeid: ua.NodeId):
         self.nodeid = nodeid
         self.attributes: Dict[ua.AttributeIds, AttributeValue] = {}
-        self.references: List[ua.ReferenceDescription] = []
+        self.references: set[ua.ReferenceDescription] = set()   # use a set because it guaranties uniqueness
         self.call = None
 
     def __str__(self) -> str:
@@ -148,7 +155,7 @@ class ViewService(object):
         return True
 
     def _suitable_reftype(self, ref1: ua.NodeId, ref2: ua.NodeId, subtypes: bool) -> bool:
-        if ref1 == ua.NodeId(ua.ObjectIds.Null):
+        if ref1 == NODE_ID_NULL:
             # If ReferenceTypeId is not specified in the BrowseDescription,
             # all References are returned and includeSubtypes is ignored.
             return True
@@ -162,7 +169,8 @@ class ViewService(object):
         nodedata = self._aspace.get(ref)
         if nodedata is not None:
             for ref_desc in nodedata.references:
-                if ref_desc.ReferenceTypeId == ua.NodeId(ua.ObjectIds.HasSubtype) and ref_desc.IsForward:
+                # first compare the bool, as it is faster, if false we save time
+                if ref_desc.IsForward and ref_desc.ReferenceTypeId == NODE_ID_HASSUBTYPE:
                     yield ref_desc.NodeId
                     yield from self._get_sub_ref(ref_desc.NodeId)
 
@@ -298,7 +306,7 @@ class NodeManagementService:
 
             # check properties
             for ref in self._aspace[item.ParentNodeId].references:
-                if ref.ReferenceTypeId == ua.NodeId(ua.ObjectIds.HasProperty):
+                if ref.ReferenceTypeId == NODE_ID_HAS_PROPERTY:
                     if item.BrowseName.Name == ref.BrowseName.Name:
                         self.logger.warning(
                             f"AddNodesItem: Requested Browsename {item.BrowseName.Name}"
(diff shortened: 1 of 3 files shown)


asyncua/server/address_space.py
```diff
@@ -33,6 +33,13 @@
 _logger = logging.getLogger(__name__)
 
 
+# usefull const NodeIds: avoid constructing multiple times for comparsion purpose
+NODE_ID_HAS_PROPERTY = ua.NodeId(ua.ObjectIds.HasProperty)
+NODE_ID_NULL = ua.NodeId(ua.ObjectIds.Null)
+NODE_ID_HASSUBTYPE = ua.NodeId(ua.ObjectIds.HasSubtype)
+NODE_ID_DEFAULT = ua.NodeId()
```
oroulet, line 40: Does that help?
I have been considering doing that in main library but i thought someone concluded it does not help much

asyncua/ua/uaprotocol_auto.py
```diff
@@ -5906,6 +5906,8 @@ def NodeClass(self):
     def NodeClass(self, val):
         self.NodeClass_ = val
 
+    def __hash__(self):
+        return hash((self.ReferenceTypeId, self.NodeId, self.IsForward))
```
oroulet, line 5910: You cannot modify an autogenerated file,
Modify the generating file!

asyncua/ua/uatypes.py
```diff
@@ -390,16 +390,6 @@ class NodeId:
...
-
-
-    :ivar Identifier:
-    :vartype Identifier: NodeId
-    :ivar NamespaceIndex:
-    :vartype NamespaceIndex: Int
-    :ivar NamespaceUri:
-    :vartype NamespaceUri: String
-    :ivar ServerIndex:
-    :vartype ServerIndex: Int
```
oroulet, line 402: Why do you remove documentation?

asyncua/ua/uatypes.py
```diff
@@ -380,7 +380,7 @@ class NodeIdType(IntEnum):
 _NodeIdType = NodeIdType  # ugly hack
 
 
-@dataclass(frozen=True, eq=False, order=False)
+@dataclass(slots=True, frozen=True, eq=False, order=False)
```
oroulet, line 383: I remember we considered using slot instead of our custon frozen thing. We had some reasons for not using it...not remembering what...

oroulet (changes requested):

asyncua/server/address_space.py
jcbastosportela: Yes, from my tests this helps.

asyncua/ua/uaprotocol_auto.py
jcbastosportela: Didn't notice this was generated, sorry, will look into the generator

asyncua/ua/uatypes.py
jcbastosportela: Slots help, but it was not the biggest improvement. Will revert this

asyncua/ua/uatypes.py
jcbastosportela: This doc was redundant with the Args: above, also it does not match the dataclass argument list

oroulet: This is a special format to generate documentation automatically. better to update it than remove it

jcbastosportela: Will do.
Thought the format above was supported (i think that is google style, also the one I use)

jcbastosportela:
I tried to run the update_ua_nodeset.py and I keep getting errors in multiple places.
Some of the issues I faced:

- update_ua_nodeset.py : 62 fails because of the permissions of .git folder in windows (worked around by manually deleting and skipping those steps)
- update_ua_nodeset.py : 64 fails because for current master branch of opc ua that file no longer exists; targeted branch (workaround by skipping that rm)
- generate_event_object.py : 85 fails because seems that master opc ua has some NodeIds not implemented by the generator. Error:
```
...
  File "C:\Users\jportela\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "c:\Sandbox\opcua-asyncio\schemas\generate_event_objects.py", line 158, in <module>
    ecg.generate_events_code(model)
  File "c:\Sandbox\opcua-asyncio\schemas\generate_event_objects.py", line 138, in generate_events_code
    self.generate_event_class(event)
  File "c:\Sandbox\opcua-asyncio\schemas\generate_event_objects.py", line 116, in generate_event_class
    self.add_properties_and_variables(event)
  File "c:\Sandbox\opcua-asyncio\schemas\generate_event_objects.py", line 62, in add_properties_and_variables
    ref.refBrowseName, self.get_value(ref),
  File "c:\Sandbox\opcua-asyncio\schemas\generate_event_objects.py", line 85, in get_value
    str(ob_ids.ObjectIdNames[int(str(reference.refId).split("=")[1])]).split("_")[0])
KeyError: 31771
``` 
- when giving "-b v1.04" fails with:
```
['git', 'clone', '--depth=1', 'https://github.com/OPCFoundation/UA-Nodeset.git', 'UA-Nodeset-master', 'v1.04']
fatal: Too many arguments.
```

Other notes: the generator script assumes that the script is running with cwd on the same dir of the update_ua_nodeset.py, maybe this should be fixed as I noticed that later.

In the end I quit trying to run it :(
Because the most important changes of this PR needs to be done on the generator and I cannot run it I am moving this PR to draft until I get more time do dedicate to this.

schroeder-:
> update_ua_nodeset.py : 62 fails because of the permissions of .git folder in windows (worked around by manually deleting and skipping those steps)
This is caused by visual studio code. There is no workaround only closing visual studio code.

> update_ua_nodeset.py : 64 fails because for current master branch of opc ua that file no longer exists; targeted branch (workaround by skipping that rm)
Last time i tried it work. Will try soon.

> generate_event_object.py : 85 fails because seems that master opc ua has some NodeIds not implemented by the generator. Error:
Noticed it last time too. But just reverted to the current used nodeset UA1.05.01. 

The easiest way currently is to check out the nodeset your self via git. Then execute all the scripts: 
```
py -3 generate_address_space.py
py -3 generate_event_objects.py
py -3 generate_ids.py
py -3 generate_protocol_python.py
py -3 generate_model_event.py
py -3 generate_event_objects.py
py -3 generate_statuscode.py
py -3 generate_uaerrors.py
```
Also be carefull if you make errors generating schemas you need to revert the changes via git. Because some generating scripts use generated code.

PR #1126, opened 2022-11-18
UaFile: Comfort method to write file

martigr:
Currently it is possible to write a file with the low-level file services, but the comfort function in `asyncua/client/ua_file.py` is missing.

I have now added the `write()` method to `asyncua/client/ua_file.py`. This allows to write a file comfortably without using the low-level functions.

```python
async with Client(url=url) as client:
    file_node = client.get_node("ns=2;s=NameOfNode")
    async with UaFile(file_node, 'w') as ua_file:
        with open("filename.txt", 'rb') as fd:
            data = fd.read()
            await ua_file.write(data) # write file
```

### DIFF
--- a/asyncua/client/ua_file.py
+++ b/asyncua/client/ua_file.py
@@ -8,6 +8,8 @@ class UaFile:
         self._handle = None
         if open_mode == 'r':
             self._init_open = ua.OpenFileMode.Read.value
+        elif open_mode == 'w':
+            self._init_open = ua.OpenFileMode.Write.value
         else:
             raise ValueError("file mode is not supported")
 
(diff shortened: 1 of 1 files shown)


AndreasHeine:
what if the file is larger then the avalible memmory? 


using a iterator of a fileobject:
```python
for line in file:
    print(line)
```

schroeder-:
> what if the file is larger then the avalible memmory?

This is not a problem because most opc ua servers doesn't allow more than 16mb of data in one service call.

martigr:
> what if the file is larger then the avalible memmory?
> 
> using a iterator of a fileobject:
> 
> ```python
> for line in file:
>     print(line)
> ```

The code snippet in my comment is just an example of the usage.
I agree that some memory issues could occur. In a real world application writing in chunks would be better.

AndreasHeine:
> > what if the file is larger then the avalible memmory?
> 
> This is not a problem because most opc ua servers doesn't allow more than 16mb of data in one service call.

the thing i am thinking of is do i really need the whole file in ram (maybe not the only file ^^)
maybe i am overconcerned...

PR #1131, opened 2022-11-25
Fix off-by-one error in continuation point calculation.

alexander-travov:
I believe there is an off-by-one error in calculation of continuation point timestamp.


alexander-travov:
For example I have a var on server with history of values [0, 1, 2, 3, 4, 5].
When I read history of that variable in portions of maxlen=3 with the use of continuation point, I get: [0, 1, 2], [4, 5]. The value 3 is never returned.

With the fix I get: [0, 1, 2], [3, 4, 5] as expected.

alexander-travov:
Same error in https://github.com/FreeOpcUa/python-opcua/pull/1494

PR #1135, opened 2022-12-01
Set ServerTimestamp when attributes/values are written

cziebuhr:
This is a follow up on #461.

According to [7.11.4 ServerTimestamp](https://reference.opcfoundation.org/Core/Part4/7.11.4/): The serverTimestamp is used to reflect the time that the Server received a Variable value or knew it to be accurate.

I think it shouldn't be set by a client, so it gets overwritten if specified. After working with some historian systems, I came to the conclusion, that it's essential to always use the same clock for each kind of timestamp.

SourceTimestamp is only set for Value Attributes, the spec doesn't specify this for ServerTimestamp, so I assume it's ok to be set for all attributes.

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -100,8 +100,12 @@ class AttributeService:
                 ):
                     res.append(ua.StatusCode(ua.StatusCodes.BadUserAccessDenied))
                     continue
+            if writevalue.AttributeId == ua.AttributeIds.Value and self._aspace.force_server_timestamp:
+                dv = dataclasses.replace(writevalue.Value, ServerTimestamp=datetime.utcnow(), ServerPicoseconds=None)
+            else:
+                dv = writevalue.Value
             res.append(
-                await self._aspace.write_attribute_value(writevalue.NodeId, writevalue.AttributeId, writevalue.Value)
+                await self._aspace.write_attribute_value(writevalue.NodeId, writevalue.AttributeId, dv)
             )
         return res
 
@@ -506,6 +510,7 @@ class NodeManagementService:
             dv = ua.DataValue(
                 ua.Variant(getattr(attributes, name), vtype, is_array=is_array),
                 SourceTimestamp=datetime.utcnow() if add_timestamps else None,
+                ServerTimestamp=datetime.utcnow() if add_timestamps and self._aspace.force_server_timestamp else None,
             )
             nodedata.attributes[getattr(ua.AttributeIds, name)] = AttributeValue(dv)
 
@@ -605,6 +610,7 @@ class AddressSpace:
 
     def __init__(self):
         self.logger = logging.getLogger(__name__)
+        self.force_server_timestamp: bool = True
         self._nodes: Dict[ua.NodeId, NodeData] = {}
         self._datachange_callback_counter = 200
         self._handle_to_attribute_map: Dict[int, Tuple[ua.NodeId, ua.AttributeIds]] = {}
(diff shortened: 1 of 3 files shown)


cziebuhr:
There's at least one drawback with the non-conditional approach on server-side: It's hard to write application-tests (e.g. for history server implementations) with fixed timestamps. An option would be to only allow specifying ServerTimestamp for users with UserRole.Admin.
What do you think?

schroeder-:
Admin is a good idea. Also we could add a Flag to change the behavior of the timestamps.

cziebuhr:
Added a flag to disable setting ServerTimestamp (especially for testing).
Limited the change to value attributes to be more consistent in the library itself.

bitkeeper:
Some feedback:
After updating our OPC UA Server with this change, our already running scada server (Atvise) breaks. The subscriptions will not work anymore. 

Luckily with the flag we good override the behaviour and continue as usual:
```python
  server.set_force_server_timestamp(False)
```
Because of this it wonders me, if the default behaviour should be to don't add the server time stamp, else you would risk braking existing environments.

cziebuhr:
@bitkeeper can you provide some more details? Which timestamps do you set when writing values? What happens on the scada server? Are you also using historical data access?

bitkeeper:
- The initial value has a sourcetimestamp, coming from the `node.add_variable`.
- On update `write_value` with `ua.DataValue` is used,  no source timestamps is set.
- No historical data access in this case.
- Both scada server and OPC UA server are both time synchronized.
- When with an existing already running scada project, the OPC UA server is replaced with server timestamp patch, the values are no longer updated in scada.

I can reproduce the simular behaviour with UaExpert:
- Setup UaExpert:
  - Add a subscription to the the node
  - Change the monitored item setting data change trigger to "Status/Value/Timestamp"
- The initial value has a source timestamp + server timestamp
image

- The first update works
image

- Any updates after it fails

cziebuhr:
What do you mean with "any updates after it fails"? Does the `write_values` call fail?
The screenshots are showing Attributes View, not Data Access View, so monitored items settings should be irrelevant here.
Why are you using "Status/Value/Timestamp" at all? If not setting SourceTimestamp, it should provide the same results as "Status/Value" (except for inital value set by `add_variable`).

bitkeeper:
Fails in this context means dat the monitored item isn't updated anymore in the Data Access View, not a code fail.

As stated in previous message I can reproduce the behaviour of the scada server with UaExpert (easier substitute then using the scada server), if the monitored item trigger setting of the item in UaExpert is set to "Status/Value/Timestamp".

And yes you can change the trigger of UaExpert to "Status/Value" but that isn't the point. The point is that before this MR it worked fine with "Status/Value/Timestamp" and now it doens't anymore. 
Of course you can change the behaviour back by using `server.set_force_server_timestamp(False)`, but this will be easy to overlook when someone just 'upgrade' the opcua-asyncio package to a newer version.

cziebuhr:
Can't reproduce it. The only thing I could imagine when writing values in the server: without set_force_timestamp, the ref to the DataValue object is saved directly, but copied when set_force_timestamp is set. Both should be fine, because DataValue is frozen and can't be mutated afterwards. Could there be a logic issue in the application when potentially trying to read the value back and comparing with what was written?

bitkeeper:
@cziebuhr the reproduced behaviour with UaExpert was though a different bug with monitored items (see #1253). 

Still no clue why the thirdparty client breaks with the forced server timestamps. For our server application we can fix it with `server.set_force_server_timestamp(False)` so for now we are fine.

Issue #1155, opened 2022-12-21
DateTime encoding


alexander-travov:
I have a question regarding DateTime encoding. It arose when I tried to use `datetime.min` as a StartTime param in HistoryRead request. The OPC server to which I send that request could not handle encoded negative value correctly.

According to the [spec](https://reference.opcfoundation.org/Core/Part6/v105/docs/5.2.2.5):

> A date/time value is encoded as 0 if either
> The value is equal to or earlier than 1601-01-01 12:00AM UTC.
> The value is the earliest date that can be represented with the [DevelopmentPlatform](https://reference.opcfoundation.org/search/206?t=DevelopmentPlatform)’s encoding.

> A date/time is encoded as the maximum value for an [Int64](https://reference.opcfoundation.org/search/206?t=Int64) if either
> The value is equal to or greater than 9999-12-31 11:59:59PM UTC,
> The value is the latest date that can be represented with the [DevelopmentPlatform](https://reference.opcfoundation.org/search/206?t=DevelopmentPlatform)’s encoding.

If I understand correctly, spec says that valid Int64 values for encoded DateTime are non-negative:
[0, 1, ... 2650467743989999999] and MaxInt64=2^63-1.

2650467743989999999 stands for datetime(9999, 12, 31, 23, 59, 58, 999999900ns) - the last represantable OPC DateTime before 9999-12-31 11:59:59PM UTC.

The [datetime_to_win_epoch](https://github.com/FreeOpcUa/opcua-asyncio/blob/bb03fed7eaef4eedb43bc078b72e740c7982e653/asyncua/ua/uatypes.py#L167) function does not seem to perform that clamping:
``` python
# correct, all zeros
>>> asyncua.ua.ua_binary.Primitives.DateTime.pack(datetime(1601, 1, 1))
b'\x00\x00\x00\x00\x00\x00\x00\x00'

# incorrect, should be all zeros b'\x00\x00\x00\x00\x00\x00\x00\x00'. We get negative number instead.
>>> asyncua.ua.ua_binary.Primitives.DateTime.pack(datetime.min)  # datetime.min is 0001-01-01 which is before 1601-01-01
b'\x00\x00\x89\xdd\xe81\xfe\xf8'

# incorrect, should be b'\xff\xff\xff\xff\xff\xff\xff\x7f' as struct.pack('<q', 2**63-1)
[...]

PR #1157, opened 2022-12-22
Truncate datetime

schroeder-:
#1155
If a datetime is equal or before 1601-01-01 12:00AM UTC or equal or later  9999-12-31 11:59:59PM UTC return the correct values as of https://reference.opcfoundation.org/Core/Part6/v105/docs/5.2.2.5#_Ref400565610 .
Also removed the warning in win_epoch_to_datetime. Because the behavior is defined in the spec.

### DIFF
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -194,7 +194,7 @@ def win_epoch_to_datetime(epch):
         # FILETIMEs after 31 Dec 9999 are truncated to max value
         return MAX_FILETIME_EPOCH_DATETIME
     if epch < 0:
-        return 0
+        return FILETIME_EPOCH_AS_DATETIME
     return FILETIME_EPOCH_AS_DATETIME + timedelta(microseconds=epch // 10)
 
 
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -45,6 +45,8 @@ FILETIME_EPOCH_AS_DATETIME = datetime(1601, 1, 1)
 FILETIME_EPOCH_AS_UTC_DATETIME = FILETIME_EPOCH_AS_DATETIME.replace(tzinfo=timezone.utc)
 MAX_FILETIME_EPOCH_DATETIME = datetime(9999, 12, 31, 23, 59, 59)
 MAX_FILETIME_EPOCH_AS_UTC_DATETIME = MAX_FILETIME_EPOCH_DATETIME.replace(tzinfo=timezone.utc)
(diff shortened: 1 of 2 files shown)


asyncua/ua/uatypes.py
```diff
@@ -165,7 +167,21 @@ class Guid(uuid.UUID):
 
 
 def datetime_to_win_epoch(dt: datetime):
+    if dt.tzinfo is None:
+        ref = FILETIME_EPOCH_AS_DATETIME
+        max_ep = MAX_FILETIME_EPOCH_DATETIME
+    else:
+        ref = FILETIME_EPOCH_AS_UTC_DATETIME
+        max_ep = MAX_FILETIME_EPOCH_AS_UTC_DATETIME
     ref = FILETIME_EPOCH_AS_DATETIME if dt.tzinfo is None else FILETIME_EPOCH_AS_UTC_DATETIME
```
alexander-travov, line 176: This line should be removed as it repeats the work of lines above.

asyncua/ua/uatypes.py
```diff
@@ -174,12 +189,12 @@ def get_win_epoch():
 
 
 def win_epoch_to_datetime(epch):
-    try:
-        return FILETIME_EPOCH_AS_DATETIME + timedelta(microseconds=epch // 10)
-    except OverflowError:
-        # FILETIMEs after 31 Dec 9999 can't be converted to datetime
-        logger.warning("datetime overflow: %s", epch)
-        return datetime(MAXYEAR, 12, 31, 23, 59, 59, 999999)
+    if epch >= 2650467743989999999:
```
alexander-travov, line 192: I believe it should be `>` instead of `>=` here.

2650467743989999999 encodes the last respresentable DateTime before 9999-12-31-59-59.

Or `>=` can stay but constant should be changed to `N+1`:
``` python
>>> int((datetime(9999, 12, 31, 23, 59, 59) - datetime(1601, 1, 1)).total_seconds()) * 10_000_000
2650467743990000000
```

alexander-travov: I also suggest creating named constants at the beginning of the file for these values to show that there is no magic about them. Something like:
``` python
# instead of 2650467743990000000
MAX_OPC_FILETIME = int((MAX_FILETIME_EPOCH_DATETIME - FILETIME_EPOCH_AS_DATETIME).total_seconds()) * HUNDREDS_OF_NANOSECONDS
# instead of 9223372036854775807
MAX_INT64 = 2 ** 63 - 1
```

alexander-travov: `MAX_INT64` constant can be reused in test too.

PR #1161, opened 2023-01-01
Fix a few typos in documentation and code comments

kbrix2000:
While reading the documentation, I noticed a few typos, which I fixed in this pull request. Moreover, there are also a few typos fixed in code comments.


oroulet:
Thanks

kbrix2000:
@oroulet Thanks, too! Can we also initiate an update of the online documentation at [ReadTheDocs](https://opcua-asyncio.readthedocs.io)?

PR #1173, opened 2023-01-11
Interpret Null UserIdentityToken as Anonymous during ActivateSession

almetge:
OPC-UA specification Part 4, 5.6.3 specifies that a Null or empty user token shall always be interpreted as anonymous. Add a test for this case and a fix to properly handle it.

I have a client that uses a Null token when connecting that is refused by the server. I don't know if the empty case would be different.

I'm not sure if the fix should remain in `internal_session.py` or move somewhere in the parser or accepted tokens, feel free to point me toward a more suitable place if needed.


oroulet:
could not find a better place.

Issue #1160, opened 2022-12-28
Tests are broken?


ratara:
Hi,
I think it is most likely a failure on my site but because the minimal server example works fine, it is a little bit strange.
When I run ./run_tests.sh most tests fail due to missing attributes.

~/Dokumente/opcua-asyncio$ ./run-tests.sh 
============================================================= test session starts =============================================================
platform linux -- Python 3.10.6, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/ralf/Dokumente/opcua-asyncio, configfile: pytest.ini, testpaths: tests
plugins: mock-3.10.0, asyncio-0.20.3, cov-4.0.0
asyncio: mode=strict
collected 581 items                                                                                                                           

tests/test_callback_service.py .                                                                                                        [  0%]
tests/test_client.py FFFF.FFFF                                                                                                          [  1%]
tests/test_common.py FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF [ 21%]
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF.                                                                             [ 31%]
tests/test_connections.py FF                                                                                                            [ 31%]
tests/test_crypto_connect.py FFFFFFFFssFFFFFFFFFFFFFF.                                                                                  [ 36%]
tests/test_custom_structures.py ..FFFsFFFF..FFFFFFFFFsFF                                                                                [ 40%]
tests/test_ha_client.py FFFFFFFFFFFFFFFF                                                                                                [ 43%]
tests/test_history.py FFFFFFFFFFFFFFFFFFFFFFFF                                                                                          [ 47%]
tests/test_history_events.py FFFFFFFFFFFFFFFFFF                                                                                         [ 50%]
tests/test_history_limits.py FFFFFF                                                                                                     [ 51%]
tests/test_permissions.py FFF                                                                                                           [ 51%]
tests/test_server.py FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF.FFFF                                                                        [ 59%]
tests/test_standard_address_space.py s                                                                                                  [ 59%]
[...]

PR #1179, opened 2023-01-15
Replace os path with pathlib

ratara:
Hi,
I took the task to replace os.path with pathlib. I got trouble to run the unit tests, so you should run them to check for errors.

I also changed some type hinting and added a FIXME comment about a critical security vulnerability according to sonarlint.


ratara:
Well, as mentioned in the other issue (https://github.com/FreeOpcUa/opcua-asyncio/issues/1160), I got the unit test running now.
I will run them now, after I solved the conflicts.

ratara:
There are some unit test fails because of a weird re-use of variables with completely different types. Looks like a technical dept to my. I will take a look..

schroeder-:
These changes are breaking the api. You should always accept both, str and path. Because in my opinion, this change is not worth breaking every application that is using this library.

ratara:
> These changes are breaking the api. You should always accept both, str and path. Because in my opinion, this change is not worth breaking every application that is using this library.

I already added str and path support because the change broke tests. I will push a suggestion as soon as all tests pass.

oroulet:
you need to rebase your branch.,
Also do we break compatibiliyt or not now?

ratara:
image
@oroulet 
@schroeder-  is talking about the section in the screenshot. At first I changed that to Path because it was the obvious move but as @schroeder- is pointing out, this would break the API because at least the security string is used with this method directly. Someone else could use this method too with str. So I added str support again. But as you can see, this leads to a more complex is/else structure due to polymorphic use of that function.

It should work now. But maybe it would be nice if @schroeder- acknowledge that this was his point and that it is fine for him now.

schroeder-:
Looks good 👍

oroulet:
huum gitlabs says that branch is still outdated . Try `git pull --rebase  origin main`

ratara:
> huum gitlabs says that branch is still outdated . Try `git pull --rebase origin main`

No, you just approved my other branch and we got a conflict here. Solved!

oroulet:
I wrote my last comment before merging your other branch, and now I still cannot merge. gitlab tell med there is a conflict: 
image

ratara:
> I wrote my last comment before merging your other branch, and now I still cannot merge. gitlab tell med there is a conflict: 
> image
> 

Ok, I try the rebase tomorrow.

ratara:
:*-( 
I tried the rebase and got merge conflicts that result in a bad state where I got unit test fails. Strange thing is, I am getting errors in files I did not touch although the commits itself look like the ones of sql injection. I need more time(and focus) to recover from this.

PR #1181, opened 2023-01-16
Use info log level for client and server limit update message

### LINKED ISSUE #1180: WARNING asyncua.uaprotocol: updating server limits to...
**Describe the bug** <br />
After an upgrade to `0.9.98` I see the following warnings:
```
2023-01-16 14:33:01,068 WARNING  asyncua.uaprotocol: updating server limits to: TransportLimits(max_recv_buffer=65535, max_send_buffer=65535, max_chunk_count=1601, max_message_size=104857600)
2023-01-16 14:33:01,077 WARNING  asyncua.uaprotocol: updating server limits to: TransportLimits(max_recv_buffer=65535, max_send_buffer=65535, max_chunk_count=1601, max_message_size=104857600)
...
```

**To Reproduce**<br />
 * Start a server (`python examples/server-minimal.py`)
 * Connect with a client to the server (`python examples/client-minimal.py`)
 

**Expected behavior**<br />

I don't expect to see warning for the regular usage.

**Version**
I tested with two sets of versions
Python-Version: 3.9.15
opcua-asyncio Version: 0.9.98

Python-Version: 3.10.9
opcua-asyncio Version: 2c78916fccef7cc72e38ad5c6308400222e0e2aa (current master)

rnestler:
Fixes #1180

### DIFF
--- a/asyncua/common/connection.py
+++ b/asyncua/common/connection.py
@@ -60,7 +60,7 @@ class TransportLimits:
         self.max_recv_buffer = ack.SendBufferSize
         self.max_send_buffer = ack.ReceiveBufferSize
         self.max_message_size = ack.MaxMessageSize
-        logger.warning("updating server limits to: %s", self)
+        logger.info("updating server limits to: %s", self)
         return ack
 
     def create_hello_limits(self, msg: ua.Hello) -> ua.Hello:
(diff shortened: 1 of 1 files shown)


schroeder-:
Maybe we should check if the values are different and only then issue a warning?

rnestler:
> Maybe we should check if the values are different and only then issue a warning?

Is it really a bad thing if this happens that it needs a warning?

schroeder-:
Normally its not a bad thing. But I theory this could be a source of errors, if the server lowers the limit to low.

schroeder- (approved):

PR #1183, opened 2023-01-18
Only log when TransportLimits actually changed

rnestler:
Based on #1181


oroulet:
need rebase too

asyncua/common/connection.py
```diff
@@ -51,17 +52,27 @@ def is_chunk_count_within_limit(self, sz: int) -> bool:
             _logger.error("Number of message chunks: %s is > configured max chunk count: %s", sz, self.max_chunk_count)
         return within_limit
 
+    def _set_new_limits_from_ack(self, ack: ua.Acknowledge, role: Literal["client", "server"]) -> None:
+        new_limits = TransportLimits(
+            max_chunk_count=ack.MaxChunkCount,
+            max_recv_buffer=ack.SendBufferSize,
+            max_send_buffer=ack.ReceiveBufferSize,
```
rnestler, line 59: I just realized that these need to be swapped for `"client"`.

rnestler: Fixed.

asyncua/common/connection.py
```diff
@@ -51,17 +52,30 @@ def is_chunk_count_within_limit(self, sz: int) -> bool:
...
+    def _set_new_limits_from_ack(self, ack: ua.Acknowledge, role: Literal["client", "server"]) -> None:
+        max_recv_buffer = ack.ReceiveBufferSize if role == "client" else ack.SendBufferSize
+        max_send_buffer = ack.SendBufferSize if role == "client" else ack.ReceiveBufferSize
+
+        new_limits = TransportLimits(
+            max_chunk_count=ack.MaxChunkCount,
+            max_recv_buffer=max_recv_buffer,
+            max_send_buffer=max_send_buffer,
+            max_message_size=ack.MaxMessageSize,
+        )
```
rnestler, line 64: I'm now not sure if it makes sense to create the `new_limits` here or if we should keep this in the other methods, since it differs between client and server.

asyncua/common/connection.py
```diff
@@ -51,17 +52,30 @@ def is_chunk_count_within_limit(self, sz: int) -> bool:
             _logger.error("Number of message chunks: %s is > configured max chunk count: %s", sz, self.max_chunk_count)
         return within_limit
 
+    def _set_new_limits_from_ack(self, ack: ua.Acknowledge, role: Literal["client", "server"]) -> None:
```
oroulet, line 55: if you really need that Role, it should be an enum.
But in that case I would jsut create two methods, one for server and one for client. That is simpler

asyncua/common/connection.py
oroulet: Sorry That MR disapeard into the backlog.. but yes inline would be better I think

PR #1185, opened 2023-01-19
Minor logging issues

lbckmnn:
I noticed two issues with logging:
- the root logger was used in some modules
- some modules use logger instead of _logger

This PR should fix both issues.


oroulet:
merged manually

Issue #276, opened 2020-08-28
security vulnerability: history_sql.py isn't injection safe!


swamper123:
Like @AndreasHeine mentioned in [python-opcua](https://github.com/FreeOpcUa/python-opcua/issues/1118) there are possible SQL Injections in history_sql.py.

The lines I found are:

https://github.com/FreeOpcUa/opcua-asyncio/blob/e6146e8666a7c1ad43ccdf74cd5fad9d5605a2ba/asyncua/server/history_sql.py#L58

https://github.com/FreeOpcUa/opcua-asyncio/blob/e6146e8666a7c1ad43ccdf74cd5fad9d5605a2ba/asyncua/server/history_sql.py#L67

https://github.com/FreeOpcUa/opcua-asyncio/blob/e6146e8666a7c1ad43ccdf74cd5fad9d5605a2ba/asyncua/server/history_sql.py#L98

https://github.com/FreeOpcUa/opcua-asyncio/blob/e6146e8666a7c1ad43ccdf74cd5fad9d5605a2ba/asyncua/server/history_sql.py#L142

https://github.com/FreeOpcUa/opcua-asyncio/blob/e6146e8666a7c1ad43ccdf74cd5fad9d5605a2ba/asyncua/server/history_sql.py#L155

https://github.com/FreeOpcUa/opcua-asyncio/blob/e6146e8666a7c1ad43ccdf74cd5fad9d5605a2ba/asyncua/server/history_sql.py#L170

Because I neither worked with SQL before, nor understanding the better way Andreas mentioned, I may not provide a PR for that at the moment (I have to check on that on Monday again, I got to many other things in my mind, but maybe I will put that on my agenda.).

JoeyFaulkner:
this imo is a good first issue if anyone's interested, there is an explainer here (https://docs.python.org/3.4/library/sqlite3.html) how to do this with sqlite.

AndreasHeine:
@JoeyFaulkner its not only the query-params, in asyncua most are parametrized but the table-name is always f-string insertion and even if its possible or not to inject a evil-node name with identifier-type string it should be avoided at all costs! if someone builds a server with dynamic eventgeneration it could be potentialy possible to break into the db. the eventdata could be sensitiv or process critical and should be protected. we need to find a consisten solution for both python-opcua and asyncio!

JoeyFaulkner:
Ah dang! Good point. Thanks for the explanation

swamper123:
[...]

PR #1186, opened 2023-01-22
276 table names are now validated to prevent sql injection

ratara:
Well, I googled a lot because I thought this is no new problem. I found a library called psycopg2 that solves the problem for postgresql. I think this library would work here but there was to much overhead to integrate this library.
I decided to solve the problem with a self-written method to validate table names.

Please take a look because it felt too easy. Maybe I forgot major problems here.

See issue: https://github.com/FreeOpcUa/opcua-asyncio/issues/276


AndreasHeine:
maybe we should replace some caracters rather then dont accept them! 

; can also be injected as %3B... so they should be disallowed as well!

ratara:
> maybe we should replace some caracters rather then dont accept them!
> 
> ; can also be injected as %3B... so they should be disallowed as well!

I can not recommend to replace characters because this leads to table names that are wrong and unexpected.
For the URL encoding topic: I am not sure, if there is anybody in the call chain that re-interprets the string in that way.

oroulet:
I am really not en expert on that but OK..

PR #1187, opened 2023-01-25
Don't use ua.Null

cziebuhr:
Prefer `None` over `ua.Null()`, otherwise a check in server code like `value.Value.Value is None` doesn't work


asyncua/server/address_space.py
```diff
@@ -780,7 +780,7 @@ async def write_attribute_value(self, nodeid: ua.NodeId, attr: ua.AttributeIds,
         if value.StatusCode is not None and value.StatusCode.is_bad():
             # https://reference.opcfoundation.org/v104/Core/docs/Part4/7.7.1/
             # If the StatusCode indicates an error then the value is to be ignored and the Server shall set it to null.
-            value = dataclasses.replace(value, Value=ua.Variant(ua.Null(), ua.VariantType.Null))
```
oroulet, line 783: maybe we should be make Null() return None?

Issue #1149, opened 2022-12-19
Support for OPC UA Programs


CraigPersonal100:
OPC UA Programs can be thought of like OPC UA Methods, used when it would not make sense to block control flow after invoking a long running task. For example, something like "TurnLightOn" could be a method while something like "MoveJointToHomePosition" could be a program. Like methods, they can be invoked by a client, but unlike methods they can also be managed by a client, e.g, to suspend it or monitor progress.

Unified Automation defines them here: https://documentation.unified-automation.com/uasdknet/2.5.7/html/L2UaPrograms.html

Their behaviour is defined by a Program Finite State Machine (PFSM).  This is not yet implemented in asyncua (see below).

https://github.com/FreeOpcUa/opcua-asyncio/blob/bb03fed7eaef4eedb43bc078b72e740c7982e653/asyncua/common/statemachine.py#L9

Is there any work being done to implement PFSM in asyncua? And would it be possible to add an api to the Nodes to easily create programs, in a similar way to how we create methods?

Maybe something like (using rclpy ROS2 Actions as an API inspiration)

```python
#server.py
async def move_joint_to_home_position(ctx: ServerProgramHandle, limit_speed: bool) -> bool:
    while true:
         #interact with hardware to invoke behaviour
         #read back from hardware to see if done or fault
         #check ctx to see if client has requested suspend or hault
         #send feedback back to client using ctx
         #break from the loop if done
    return true

server = Server()
#init and setup server and namespace
...
objects = server.nodes.objects
await objects.add_folder(idx, "myEmptyFolder")
myobj = await objects.add_object(idx, "MyObject")
await myobj.add_program(idx, "move_home", move_joint_to_home_position, [ua.VariantType.Bool], [ua.VariantType.Bool])
```

This will work in a similar way to creating a method with an async function, but the server owns a state machine that represents the methods state. 

The client could work in a similar way to invoking methods, but with some extra functionality

```python
#client.py
#Setup client
...
handle: ClientProgramHandle= await client.nodes.objects.start_program(f"{nsidx}:myEmptyFolder:MyObject:move_home", true)
while not handle.is_in_terminal_state():
    print(handle.read_feedback())
if handle.has_result()
    res: bool = handle.result()
```

AndreasHeine:
hey @CraigPersonal100,

currently i stopped at finitestatemachine!

the correct reference would by the official one -> https://reference.opcfoundation.org/v104/Core/docs/Part10/

[...]

PR #1190, opened 2023-02-02
Fix finitestatemachne

AndreasHeine:


AndreasHeine:
this PR is foundational work for Part10: Programs

there is still stuff to do according to the instantiation of optionals... currently we only do all or nothing but for Part 10 we probably need a better way to say what optionals we want to instantiate!? Otherwise we can only support a fullblown Program and a minimal one...


Referenced issues/discussions

https://github.com/FreeOpcUa/opcua-asyncio/issues/1149
https://github.com/FreeOpcUa/opcua-asyncio/discussions/1189

PhilWillecke:
I have seen you did some updates on the statemachine.
What is the current status of the TODOs?

Thanks  for the great work and thanks in advance

PR #1206, opened 2023-02-15
Loglevel

cziebuhr:
Silence some warnings


AndreasHeine:
anyhow log message in debug or warn would be helpful!

cziebuhr:
See my commit message: It get's already logged as info (CancelledError) or exception (Exception) in _monitor_server_loop / _renew_channel_loop. 

In my case, loglevel is set to error and the connection loss has already been handled. But I still need to call disconnect() for proper cleanup. The message will then be unconditionally logged again as exception.

Issue #1067, opened 2022-10-05
[ua.uatypes.String] used baseclass


bitkeeper:
Most types in `ua.uatypes`, for example like `CharArray` and `Float`, are derived from a matching Python datatype.
But this isn't the case for `ua.uatypes.String`. In #609 the typedef is changed from `class String(str):` to `class String:`.

Not sure why this is done and if the original reason is still the case (@oroulet ?). If I change it back and run pytest no new errors are introduced. It even fix an existing issue in the test:
```
./tests/test_common.py::test_alias[client] Failed: [undefined]AttributeError: 'String' object has no attribute 'encode'
```

By having no str as baseclass makes use of the python typing system hard:
```python
my_string : ua.String = "" # due typing invalid
my_string : ua.String = ua.String()  # Quite unusable, return empty ua.String object
```

What about changing it back to class `class String(str):`?

schroeder-:
I think the problem is, that in opc ua a string is not a python string. Because a ua.String can be Null(None), empty('') or any str content. So if we are transparent then a ua.String is like Optional[str]. The problem is that a server/client interpret None and empty different, so we can't just convert a Null to empty without consequence.
I don't know how we can describe that in python in a class.

bitkeeper:
@schroeder- are you aware of an existing test where this behaviour is demonstrated/tested?

The current definition `class String:` also can't hold this values? And it places where a ua.String is currently str are used. For example the structs dataclasses. So the current implementation also doesn't seems entirely strict.

schroeder-:
There is a test for it , but it doesn't use a struct so no no linter will find this.
https://github.com/FreeOpcUa/opcua-asyncio/blob/602249d9a02dadfa6e1e82bd900a2064854d50c1/tests/test_common.py#L481-L491

Here is decoder/encoder:
https://github.com/FreeOpcUa/opcua-asyncio/blob/602249d9a02dadfa6e1e82bd900a2064854d50c1/asyncua/ua/ua_binary.py#L64-L76

bitkeeper:
Created the a test for None/''/'value' behaviour below. @schroeder- can you review if the test below is correct?
A side from correct python typing, the results for `class String:` and `class String(str):` is the same. So it looks that using `class String(str):` will not breaking anything.




The nodeset:
```xml
  <UADataType NodeId="ns=1;i=6101" BrowseName="4:MyCrazyStruct">
    <DisplayName>MyCrazyStruct</DisplayName>
[...]

PR #1224, opened 2023-02-22
[ua.uatypes.String] used str as base class instead off nothing

bitkeeper:
Most types in `ua.uatypes`, for example like `CharArray` and `Float`, are derived from a matching Python datatype.
But this isn't the case for `ua.uatypes.String`. In #609 the typedef is changed from `class String(str):` to `class String:`.

Not sure why this is done and if the original reason is still the case . If I change it back and run pytest no new errors are introduced. It even fix an existing issue in the test:
```
./tests/test_common.py::test_alias[client] Failed: [undefined]AttributeError: 'String' object has no attribute 'encode'
```

By having no str as baseclass makes use of the python typing system hard:
```python
my_string : ua.String = "" # due typing invalid
my_string : ua.String = ua.String()  # Quite unusable, return empty ua.String object
```
Fro more background informatio see discussion #1067

### DIFF
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -149,7 +149,7 @@ class Null:  # Null(NoneType) is not supported in Python
     pass
 
 
-class String:  # Passing None as arg will result in unexpected behaviour so disabling
+class String(str):
     pass
 
 


oroulet:
I am rather sure there was a reason for that. I removed the inheritance due to something... but looks like we did not write why or tests and now nobody knows...
So let's try again...

PR #1227, opened 2023-02-22
tweak default values used fields of ExtensionObject dataclasses

bitkeeper:
Changes:
* Python typing friendly for numbers and strings
* Make optional basic type fields default None

### DIFF
--- a/asyncua/common/structures104.py
+++ b/asyncua/common/structures104.py
@@ -132,7 +132,10 @@ def clean_name(name):
     return newname
 
 
-def get_default_value(uatype, enums=None, hack=False):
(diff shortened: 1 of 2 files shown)


cziebuhr:
Be careful with using `ua.String`, it doesn't inherit from `str`:
- `bool(ua.String())` will return `True`, potentially leading to logic errors
- I assume it can't be encoded: `ua.String()` is not `None`, so encoding will issue an `.encode()` on it, which fails.

I recently had a similar issue with `ua.Null`, see #1187.

cziebuhr:
Ah, didn't see #1224. Should be ok then.

PR #1238, opened 2023-03-05
[Datatypes] allow optional arrays

schroeder-:
See end of #839. Allow Datatypes with something like this:

```Python
class X: 
      x: Optional[list[int]]
```

### DIFF
--- a/asyncua/common/structures104.py
+++ b/asyncua/common/structures104.py
@@ -229,8 +229,9 @@ class {struct_name}{base_class}:
             uatype = 'String'
         elif sfield.ValueRank >= 1 or sfield.ArrayDimensions:
             uatype = f"List[{uatype}]"
-        elif sfield.IsOptional:
+        if sfield.IsOptional:
             uatype = f"Optional[{uatype}]"
+            default_value = 'None'
         fields.append((fname, uatype, default_value))
     if is_union:
         # Generate getter and setter to mimic opc ua union access
@@ -256,7 +257,6 @@ class {struct_name}{base_class}:
     else:
         for fname, uatype, default_value in fields:
             code += f"    {fname}: {uatype} = {default_value}\n"
-
     return code
 
 
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -5,7 +5,7 @@ Binary protocol specific functions and constants
 import functools
 import struct
 import logging
-from typing import Any, Callable
+from typing import Any, Callable, Union
 import typing
 import uuid
 from enum import Enum, IntFlag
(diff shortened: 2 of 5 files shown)


asyncua/common/structures104.py
```diff
@@ -229,8 +229,9 @@ class {struct_name}{base_class}:
             uatype = 'String'
         elif sfield.ValueRank >= 1 or sfield.ArrayDimensions:
             uatype = f"List[{uatype}]"
-        elif sfield.IsOptional:
+        if sfield.IsOptional:
             uatype = f"Optional[{uatype}]"
+            default_value = 'None'
```
oroulet, line 234: that change has probably a lot of impact, maybe even can break compatibilty with some server/client...
It is correct though...

oroulet: but you should really test with several server/client and confirm it workss....

schroeder-: Shouldn't break anything. Optional values are just set to None, which every client/server should support. Also we already do this some lines above anyway. It's just added again, if a array is optional.

bitkeeper: When in the past an optional array is used (which was just an empty array, but not `None`), the values could be iterated like:

```python
for entry in mystruct.myarray:
 ...
```
Because `mystruct.myarray` now can be `None` such code will throw exceptions.
And because the code is generated on runtime, the Python typing will not be helpfull to detect such problems in your code.

That said, I like it more with the Optional[ArrayType], is more in line with the already Optionals of the other variables.

schroeder-: Before this pr optional arrays where not possible. Also in theory they are useless in general, because they only safe about 4byte in a struct. But there are some specs out there that use them for what ever reason, instead of normal arrays. 
But I can change the default value to a empty list. It doesn't realy matter.

bitkeeper: I don't mind the change.

Anyway to prevent typos due optional etc, instead of using the dynamic generated code, we import generated code. In this way we keep the Python typing system + linter hints.

With [uagendt.py](https://github.com/bitkeeper/opcua-asyncio/commit/a090fa7c686f7021300fc876d47ab8908733ebe6). It uses the code generate structures104 and some registration stuff + namespace corrections.

schroeder-: My problem is that it interferes with our binary decoder. If the default value is not None. It will be a empty array, if the value is not set and if it is an empty array. So no a application can't know if the field is left out. Also it's a confusing testcase, where we have to handle the client side test (value is encoded/decoded) and the server side test (no encoding/decoding) different.

PR #1258, opened 2023-03-24
Adds missing method: status_change_notification

mahadi:
See for details: https://github.com/FreeOpcUa/opcua-asyncio/discussions/1257

### DIFF
--- a/asyncua/common/subscription.py
+++ b/asyncua/common/subscription.py
@@ -174,9 +174,9 @@ class Subscription:
             return
         try:
             if asyncio.iscoroutinefunction(self._handler.status_change_notification):
-                await self._handler.status_change_notification(status.Status)
+                await self._handler.status_change_notification(status)
             else:
-                self._handler.status_change_notification(status.Status)
+                self._handler.status_change_notification(status)
         except Exception:
             self.logger.exception("Exception calling status change handler")
 
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -151,7 +151,9 @@ class _SubHandler:
 
     def event_notification(self, event):
         self.sync_handler.event_notification(event)
-
+        
+    def status_change_notification(self, status: ua.StatusChangeNotification):
+        self.sync_handler.status_change_notification(status)
 
 class Client:
     def __init__(self, url: str, timeout: int = 4, tloop=None):
--- a/tests/test_connections.py
+++ b/tests/test_connections.py
@@ -54,8 +54,8 @@ async def test_client_connection_lost():
         def __init__(self) -> None:
             self.status = ua.StatusCodes.Good
 
-        def status_change_notification(self, status):
-            self.status = status
+        def status_change_notification(self, status: ua.StatusChangeNotification):
+            self.status = status.Status
     # Test the disconnect behavoir
     port = find_free_port()
     srv = Server()


schroeder-:
You are right, I am not sure if we should change the parameter to StatusChangeNotification or change the type hint to StatusCode? Because StatusChangeNotification only contains a StatusCode and DignosticInfo.

mahadi:
In the code base we have different usages of the `status_change_notification` method:

- examples/client_reconnect.py in class SubHandler: `def status_change_notification(self, status: ua.StatusChangeNotification)`
- asyncua/common/subscription.py in class SubHandler: `def status_change_notification(self, status: ua.StatusChangeNotification)`
- asyncua/client/ua_client.py in class UaClient: `async def inform_subscriptions(self, status: ua.StatusCode)`
- test/test_connections.py: in class LostSubHandler: `def status_change_notification(self, status)`

So it seems to be mixed somehow. Basically i would propose to change it to `ua.StatusChangeNotification` for all occurences and let the user decide how to deal with the additional information (DiagnosticInfo) which can be useful from my point of view.

At the moment i am using 

``` python
    from asyncua.ua.status_codes import get_name_and_doc

    def status_change_notification(self, status_code: StatusCode):
        details = get_name_and_doc(status_code)
        logger.info(f"Received a StatusChangeNotification: {status_code}: {details}")
```
to gain additional information.

But i guess this change to `StatusChangeNotification` could break user code.

What do you think?

schroeder-:
It's a newly added feature, from the end of last year. So a breaking change could be justified.

oroulet:
Should I merge?

mahadi:
Not yet from my point of view: If we decide to pass a StatusChangeNotification from now then i need to add the adaptation of the remaining code. Then you could merge.

oroulet:
Flag it as draft then

dirzu:
# subscription
 async def _call_status(self, status: ua.StatusChangeNotification):
        try:
            if asyncio.iscoroutinefunction(self._handler.status_change_notification):
                await self._handler.status_change_notification(status.Status)
            else:
                self._handler.status_change_notification(status.Status)
        except Exception:
            self.logger.exception("Exception calling status change handler")
#ua client
  async def inform_subscriptions(self, status: ua.StatusCode):
        """
            Inform all current subscriptions with a status code. This calls the handlers status_change_notification
        """
        status_message = ua.StatusChangeNotification(Status=status)
        notification_message = ua.NotificationMessage(NotificationData=[status_message])
        for subid, callback in self._subscription_callbacks.items():
            try:
                parameters = ua.PublishResult(
                    subid,
                    NotificationMessage_=notification_message
                )
                if asyncio.iscoroutinefunction(callback):
                    await callback(parameters)
                else:
                    callback(parameters)
            except Exception:  # we call user code, catch everything!
                self.logger.exception("Exception while calling user callback: %s")

I had an error recently with status datatype  in subscription status is expected to be int or StatusChangeNotification?

mahadi:
Can you please format the code so it is easier to read?

If i check the _call_status method it should pass a StatusChangeNotification: https://github.com/FreeOpcUa/opcua-asyncio/blob/ab6d30a9b628553ab461b47ba61b93fe4d6f7871/asyncua/common/subscription.py#L225

Which version are you using? Where did you copy the above code from?

dirzu:
Hi the code is from asyncua/common/subscription.py i had installed an older version.
It was an older version, i've upgraded since, now is ok

> Can you please format the code so it is easier to read?
> 
> If i check the _call_status method it should pass a StatusChangeNotification:
> 
> https://github.com/FreeOpcUa/opcua-asyncio/blob/ab6d30a9b628553ab461b47ba61b93fe4d6f7871/asyncua/common/subscription.py#L225
> 
> Which version are you using? Where did you copy the above code from?

Hi, the code is from asyncua/common/subscription.py. It was an older version, i've upgraded since, now is ok.

Issue #1256, opened 2023-03-23
Error in xmlimporter.py? Can not import the xml file of the Scales Companion Specification


MM-MII:
Hi, 
I am currently facing an issue with my opc ua server. I am trying to import the xml file of several Companion Specifications, as my information model relies on DataTypes that are specified within these specifications. 
Loading the xml File of DI, Machinery, PackML and IA seems to work fine. Loading the Scales nodeset does not work - my server script and the error message follows. 
All Nodesets come from the OPCF Github: https://github.com/OPCFoundation/UA-Nodeset

Server:
```
import asyncio
from asyncua import ua, Server
import os

# Define the path to the XML file containing the information model
endpoint_url = "opc.tcp://localhost:4840/"
DI_path = os.path.abspath("C:/Users/X/Documents/Demonstrator_Software/Server/Opc.Ua.Di.NodeSet2.xml")
PackML_path = os.path.abspath("C:/Users/X/Documents/Demonstrator_Software/Server/Opc.Ua.PackML.NodeSet2.xml")
IA_path = os.path.abspath("C:/Users/X/Documents/Demonstrator_Software/Server/Opc.Ua.IA.NodeSet2.xml")
machinery_path = os.path.abspath("C:/Users/X/Documents/Demonstrator_Software/Server/Opc.Ua.Machinery.NodeSet2.xml")
scales_path = os.path.abspath("C:/Users/X/Documents/Demonstrator_Software/Server/Opc.Ua.Scales.NodeSet2.xml")
#nodeset_path = os.path.abspath("C:/Users/X/Documents/Demonstrator_Software/Server/opcuacs_demonstrator.xml")

async def main():
    # Initialize the OPC UA server
    server = Server()
    await server.init()
    server.set_endpoint(endpoint_url)
    server.set_server_name("Test Server")

    server.set_security_policy(
        [
            ua.SecurityPolicyType.NoSecurity,
            #ua.SecurityPolicyType.Basic256Sha256_SignAndEncrypt,
            #ua.SecurityPolicyType.Basic256Sha256_Sign,
        ]
    )

    # Load the information model from the XML file
      
    await server.import_xml(DI_path)
    print("DI loaded")
    await server.import_xml(PackML_path)
    print("PackML loaded")
    await server.import_xml(IA_path)
    print("IA loaded")
    await server.import_xml(machinery_path)
    print("Machinery loaded")
    await server.import_xml(scales_path)
    print("Scales loaded")


[...]

PR #1259, opened 2023-03-26
Harding xml import

schroeder-:
Resolves error when import a VariableType with values #1256.
Encoding error when imported via client.
Respect datatypes in the sorting of import nodes.
Loading structs and enums directly on import, so these types can be used in the nodeset see #1248. 
Added a test for following companion specifications to verify the importer: DI, IA, , Scales, PackML, MachineVision.
Prevent endlessloop in sort nodes


oroulet:
does that work? Should we restart tests?

schroeder-:
Is still not complete. I keep working on it. Not to easy to keep all test working 😅.

schroeder-:
Also not sure if we should realy add a 6mb nodeset for our test, maybe I look for a smaller companion specification with complex content.

GoetzGoerisch:
> Also not sure if we should realy add a 6mb nodeset for our test, maybe I look for a smaller companion specification with complex content.

Which TypeDefinition are you specifically looking at in the CS named above?

We could add the OPC UA NodeSet repo as a git submodule like open62541 does.
- Benefit, always the latest released versions available

schroeder-:
@GoetzGoerisch I just added Nodesets that i found in our issue tracker.

Removes TMC because we it has a Datatype MessageType, which collides with our internal datatype. 

```
@dataclass
class MessageType:
    Invalid: bytes = b'INV'  # FIXME: check value
    Hello: bytes = b'HEL'
    Acknowledge: bytes = b'ACK'
    Error: bytes = b'ERR'
    SecureOpen: bytes = b'OPN'
    SecureClose: bytes = b'CLO'
    SecureMessage: bytes = b'MSG'
```
Possibilities are to change rename or move the module of MessageType.
The extended possibility is to create a datatype with a added namespace like "MessageType_6", but this involves a lot of changes in our codebase.

schroeder-:
Added the nodeset via github submodule. My only concern is that it will break our tests and block changes, if a nodeset changes. 
I am not a expert in git, is there a way use specific tag or revision for the submodule?

GoetzGoerisch:
The gut commit which you added is fixed. Updates can be made manually or with automation via dependabot or renovate, which raises PR when ever there is an update upstream.

If this is desired I would contribute configuration for dependabot or renovate.

schroeder-:
@GoetzGoerisch thanks for the clarification, create a dependabot configuration.

oroulet:
please start name of MR with WIP: or Draft: if it is not finished. I am a bit confused here

schroeder-:
It's finished now. There is only one nodeset not working because of a internal class we use. But this can make it into another pr.

PR #1265, opened 2023-03-27
Fix publishing in high load situations

cziebuhr:
PublishResults get created each PublishingInterval and queued in UaProcessor, even if no PublishRequest has been received yet.

In high load / latency situations, this leads to two issues:
a) Consuming more and more memory, because more results are getting queued than PublishRequests will arrive
a) QueueSize for monitored items doesn't really work, because each PublishingInterval, changes are already converted to PublishResults. New changes will end up in new queues, instead of old values getting discarded.

This fix now waits till a PublishRequest is available, before actually creating PublishResults.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -187,8 +187,8 @@ class InternalSession(AbstractSession):
         """COROUTINE"""
         return await self.iserver.method_service.call(params)
 
-    async def create_subscription(self, params, callback=None):
-        result = await self.subscription_service.create_subscription(params, callback, external=self.external)
+    async def create_subscription(self, params, callback, request_callback=None):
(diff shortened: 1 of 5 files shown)


oroulet:
@AndreasHeine since you seems to know the protocol. Is that change correct?
That change make some changes/PublishResuts disapear... but maybe it is fine..

AndreasHeine:
@oroulet sounds reasonable if the fix exactly does that!

@cziebuhr can you test your fix against different other UA-Clients / UA-Stacks e.g. UAExpert (Unified Automation), Prosys-UAClient (Java-Implementation), opcua-commander (node-opcua), ...
so we make sure the rest is still working if we merge and release!?
can you add some unit-tests for that?
and thanks for your efforts ;)

cziebuhr:
Added a unit-test and tested with UAExpert, OSIsoft PI and KEPServerEX.

PR #1270, opened 2023-03-31
Add complete dependabot config

schroeder-:
Somehow the config got lost


oroulet:
What is that supposed to do?

schroeder-:
Create pull request when the opc foundation nodesets (git submodule) get updated. So our tests can run against the current nodesets.

PR #1276, opened 2023-04-11
Add changelog

rth:
Adds a changelog since v0.9.94

Closes https://github.com/FreeOpcUa/opcua-asyncio/discussions/1010

For each release the changelog is constructed by searching PRs between previous and current release dates. For instance,
```
is:pr is:closed sort:updated-desc merged:2022-11-07..2023-04-05 base:master 
```
It's not perfect, but in any case users can discover more context by clicking on the corresponding PR links.

The rendered markdown version can be seen [here](https://github.com/rth/opcua-asyncio/blob/add-changelog/CHANGELOG.md)

Ideally changelog entries should be added in new PRs (unless someone want's to do this before each future release).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -217,35 +217,27 @@ results in an exception
 ### Added
 - Add support for pre 1.04 optionsets
   [#900](https://github.com/FreeOpcUa/opcua-asyncio/pull/900)
-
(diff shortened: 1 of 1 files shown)


oroulet:
There are tools to use commits, but often it is too detailled, so using PR as you did was a good idea.

PR #1282, opened 2023-04-12
Correct certificate checks for application_uri and hostname

eilers:
application_uri is a string, but san_uri is an array of strings. The comparison will fail therefore.


schroeder-:
Good find, want to check it with the up comming cert generation feature, but got delayed as always.
@oroulet  good to merge.

PR #1289, opened 2023-04-20
Update handling of empty server uri params

Kallelele:
When sending FindServerRequest with empty serverUris parameter ServerUris can't be looped over.

When length of the list for serverUris is -1, _create_uatype_array_deserializer will return None.
This can't be looped over.

I suggest to set params_server_uris to empty list if this happens.

### DIFF
--- a/asyncua/crypto/uacrypto.py
+++ b/asyncua/crypto/uacrypto.py
@@ -17,6 +17,10 @@ import logging
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


Kallelele:
Added missing exception in uacrypto used in security_policies.py
https://github.com/FreeOpcUa/opcua-asyncio/blob/c6b58b6fa7b754dfaaa48b79c11f28c6be14acc4/asyncua/crypto/security_policies.py#L427

PR #1309, opened 2023-05-18
Typing nodeid for better type hints.

compsci-simon:
When working with a nodeid, it should never be None, as a node will raise an error on initialisation if the nodeid cannot be set. Thus, it is appropriate to set the type of the nodeid to ua.NodeId which can help provide better type hints.


compsci-simon:
I did not create this pull request off the master branch.

PR #1311, opened 2023-05-18
Add `read_attributes` method to `Client`

petebachant:
`read_values` is a nice and fast way to get a bunch of values in one request, but I wanted access to the `SourceTimestamp` in addition to the `Value`, which wasn't possible with this method, so I added it. If there's another way to read the values and timestamps from hundreds of nodes in one request, this isn't necessary.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -865,14 +865,12 @@ class Client:
             node.nodeid = node.basenodeid
             node.basenodeid = None
 
-    async def read_attributes(self, nodes, attr=ua.AttributeIds.Value):
+    async def read_attributes(self, nodes: List[Node], attr=ua.AttributeIds.Value):
         """
         Read the attributes of multiple nodes.
         """
         nodeids = [node.nodeid for node in nodes]
-        return await self.uaclient.read_attributes(
-            nodeids, ua.AttributeIds.Value
-        )
+        return await self.uaclient.read_attributes(nodeids, attr)
 
     async def read_values(self, nodes):
         """
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -865,13 +865,24 @@ class Client:
             node.nodeid = node.basenodeid
             node.basenodeid = None
 
-    async def read_values(self, nodes):
+    async def read_values(self, nodes, depth=2):
         """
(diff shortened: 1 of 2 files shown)


AndreasHeine:
the timestamp and statuscode will be present if you just use read attribute method which return is a datavalue. read_values is a highlevel-abstraction of read_attributes!

petebachant:
> the timestamp and statuscode will be present if you just use read attribute method which return is a datavalue. read_values is a highlevel-abstraction of read_attributes!

Are those accessible from the `Client` and `SyncClient` classes? I didn't see that as an instance method.

AndreasHeine:
> > the timestamp and statuscode will be present if you just use read attribute method which return is a datavalue. read_values is a highlevel-abstraction of read_attributes!
> 
> Are those accessible from the `Client` and `sync.Client` classes? I didn't see that as an instance method.

i dont use the sync stuff that often, i need to look deeper into it!

AndreasHeine:
yes you are right its missing in sync... but it should ve an easy addon similar to read_values 😉

asyncua/client/client.py
```diff
@@ -865,13 +865,21 @@ async def unregister_nodes(self, nodes):
             node.nodeid = node.basenodeid
             node.basenodeid = None
 
+    async def read_attributes(self, nodes, attr=ua.AttributeIds.Value):
```
oroulet, line 868: OK. But could you add types?

petebachant: Type hints added!

asyncua/client/client.py
```diff
@@ -865,13 +865,19 @@ async def unregister_nodes(self, nodes):
             node.nodeid = node.basenodeid
             node.basenodeid = None
 
+    async def read_attributes(self, nodes: List[Node], attr: ua.AttributeIds = ua.AttributeIds.Value):
```
oroulet, line 868: I added a comment because I wondered what is the return type of that method. Now I am still wondering ;-) since you did not add the type of the reutrn value  List[ua.DataValue] maybe?

oroulet: merging now anyway, we are missing types at many places..

PR #1343, opened 2023-06-20
Sample on instantiating objects from imported nodesets into a server

herculesp17:
This sample server documented how to import several nodesets in a server and how to instantiate objects with the mandatory nodes predefiend in a  nodeset. It further shows how to write and update nodes.


oroulet:
looks good!

Issue #1323, opened 2023-06-01
Limitation for historical data read does not work


KallistoD:
FreeOpcUa / opcua-asyncio / asyncua / common / node.py / read_raw_history() has description:
> Read raw history of a node
> result code from server is checked and an exception is raised in case of error
> If numvalues is > 0 and number of events in period is > numvalues
> then result will be truncated

Expected behavior:
If non-zero value provided for numvalues then read_raw_history() should request from server and then return no more than numvalues historical values for a Node (OPC UA variable).
This option is useful to manage memory usage on client side then reading huge amount of historical data in chunks. It is no always possible to predict memory usage by setting starttime and endtime values. So, numvalues is the solution.

Actual behavior:
read_raw_history() reads and returns all the historical data regardless of numvalues option. It does not provide result truncation as declared.
Non-zero numvalues causes a series of history_read() requests instead of single request. But all of them performed within single read_raw_history() call, not allowing to control memory usage on client side with numvalues option.

Please, consider to fix this behavior. Possible solution is to break from 'while True' loop inside read_raw_history() if numvalues is non-zero. Single call of history_read() should be enough if numvalues is non-zero and passed into details.NumValuesPerNode field. Something like adding at the end of 'while True' loop the following lines:
`if numvalues > 0:`
`    break`

The above applicable to opcua-asyncio code in master branch on 1st of June, 2023.
Documentation reference: https://reference.opcfoundation.org/v104/Core/docs/Part11/6.4.3/#6.4.3.1

schroeder-:
Actually i wouldn't do that. Because it will not free the continuations points on the server. 
The problem is that the server doesn't respect NumValuesPerNode, so i would ask the server developer to fix that behaviour.
Otherwise you can use the low level client and use our custom read_raw_history()
```
def my_read_history_raw(client: Client,  starttime=None, endtime=None, numvalues=0, return_bounds=True):
        details = ua.ReadRawModifiedDetails()
        details.IsReadModified = False
        if starttime:
            details.StartTime = starttime
        else:
            details.StartTime = ua.get_win_epoch()
        if endtime:
            details.EndTime = endtime
        else:
            details.EndTime = ua.get_win_epoch()
        details.NumValuesPerNode = numvalues
        details.ReturnBounds = return_bounds
        history = []
        continuation_point = None
        while len(history) < numvalues:
            valueid = ua.HistoryReadValueId()
            valueid.NodeId = self.nodeid
[...]

PR #1351, opened 2023-06-23
Fix for numvalues parameter in Node.read_raw_history()

KallistoD:
The description of the function FreeOpcUa / opcua-asyncio / asyncua / common / node.py / read_raw_history() declares that non-zero `numvalues` parameter should truncate the amount of historical data to receive from a server:
> Read raw history of a node
> result code from server is checked and an exception is raised in case of error
> If numvalues is > 0 and number of events in period is > numvalues
> then result will be truncated

Actually, `numvalues` parameter is practically useless and has no effect on the function behavior. The function always reads and returns full set of historical information as defined by the other parameters regardless of `numvalues` value.
This pull request eliminates the inconsistence between the function declaration and actual behavior. If the pull request applied, the function asks for the limited amount of historical data from a server (as requested by non-zero `numvalues` parameter). It makes it possible to effectively control RAM usage of client side without guessing right period of time for historical data to request.
Please, note:
- This pull request may cause ContinuationPoint to leak on server side. But server stability should not rely on client behavior anyway. The client may crash at anytime, and ContinuationPoint may leak as well. So, possible ContinuationPoint leaks must be managed on server side (possible, through expiration) anyway.
- This pull request does not violate the [standard](https://reference.opcfoundation.org/v104/Core/docs/Part11/6.4.3/#6.4.3.1):
> When a continuationPoint is returned, a Client wanting the next numValuesPerNode values shall call ReadRaw again with the continuationPoint set.

So, the client is free to decide if it wants to request the rest of historical data from ContinuationPoint or not.
The pull request was created as a continuation of this [issue](https://github.com/FreeOpcUa/opcua-asyncio/issues/1323). Tested against 3rd party OPC UA server.

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -543,7 +543,7 @@ class Node:
         details.ReturnBounds = return_bounds
         history = []
         continuation_point = None
-        while True:
+        while len(history) < numvalues:
             result = await self.history_read(details, continuation_point)
             result.StatusCode.check()
             continuation_point = result.ContinuationPoint
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -543,7 +543,7 @@ class Node:
         details.ReturnBounds = return_bounds
         history = []
         continuation_point = None
-        while len(history) < numvalues:
+        while True:
             result = await self.history_read(details, continuation_point)
             result.StatusCode.check()
             continuation_point = result.ContinuationPoint
@@ -551,7 +551,9 @@ class Node:
             # No more data available
             if continuation_point is None:
                 break
-
+            # No more data needed
+            if numvalues > 0:
+                break
         return history
 
     async def history_read(self, details, continuation_point=None):


schroeder-:
I don't like this change, because this is a bug on the third party server. Also if you surpply num = 0 then it is unlimited values. So this change breaks this change as well.

schroeder-:
Maybe we should make this change optional, via a boolean parameter?

KallistoD:
Thank you for your attention. My bad. I applied the wrong solution from the [issue](https://github.com/FreeOpcUa/opcua-asyncio/issues/1323). Now, I fixed it.
There is absolutely no bug in 3rd party OPC UA server. It respects `numvalues` value according to the [standard](https://reference.opcfoundation.org/v104/Core/docs/Part11/6.4.3/#6.4.3.1). Behavior checked with WireShark. If the pull request applied:
- `read_raw_history()` call with `numvalues=0` executes `self.history_read()` once. It returns whole set of historical data as requested by `starttime` and `endtime` parameters
- `read_raw_history()` call with `numvalues>0` executes `self.history_read()` once as well. But it returns the set of historical data limited by `numvalues` historical values

I feel no need in additional boolean parameter. If you need to limit the returned array size, you set `numvalues>0`. If you don't need it, you set `numvalues=0`. This should work with any OPC UA server, which respects [standard](https://reference.opcfoundation.org/v104/Core/docs/Part11/6.4.3/#6.4.3.1).

schroeder-:
You are right and with this change this should work, as expected.

PR #1360, opened 2023-07-10
Fix missing SecurityLevel of the EndpointDescription

bitkeeper:
According to Part 4 - 7.10 this should be an a number that indicates how secure the EndPoint is. The current implementation used a level of 0 for every EndPoint.
The number is Server implementation specific, how higher the number how safer the security is.
See https://reference.opcfoundation.org/Core/Part4/v104/docs/7.10

Case:
Hit this issue when implementing a server which supports a GDS push application/certificate with the Ua GDS Server. Ua GDS server ignores endpoints with a security level of 0.

### DIFF
--- a/asyncua/server/server.py
+++ b/asyncua/server/server.py
@@ -402,6 +402,63 @@ class Server:
                                              ua.MessageSecurityMode.Sign, self.certificate, self.iserver.private_key,
                                              permission_ruleset=self._permission_ruleset))
 
+
+    @staticmethod
+    def lookup_security_level_for_policy_type( security_policy_type: ua.SecurityPolicyType ) -> ua.Byte:
+        """Returns the security level for an ua.SecurityPolicyType.
+
+        This is endpoint & server implementation specific!
+
+        Returns:
+            ua.Byte: the found security level
+        """
+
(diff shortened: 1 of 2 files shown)


asyncua/server/server.py
```diff
@@ -402,6 +402,63 @@ async def _setup_server_nodes(self):
                                              ua.MessageSecurityMode.Sign, self.certificate, self.iserver.private_key,
                                              permission_ruleset=self._permission_ruleset))
 
+
+    @staticmethod
+    def lookup_security_level_for_policy_type( security_policy_type: ua.SecurityPolicyType ) -> ua.Byte:
```
bitkeeper, line 407: Not sure if `server.py` is the right location for the two helper methods. The helpers are only relevant for the Server and server tests.
Maybe the there is better location in an other file with more helpers?

Considered the following locations:
* `uatypes`: `ua.SecurityPolicyType` is implemented in there, but only contains type related stuff
* `uacrypto`: looking at the other members is doesn't belong here also

oroulet: not sure either. we just leave it here..

PR #1366, opened 2023-07-18
Fix typo of the StatusCode name

okapies:


oroulet:
@okapies as written in fil header, this file is autogenerate, you cannot simply modify it. Either the error is in the spec from the foundation or in the script generating it. a PR to fix generating script is welcome, but unfortunately I cannot merge that one

PR #1368, opened 2023-07-18
Add several helpers for creating certificates

bitkeeper:
To encourage the use of certifactes, it should be easy to create certificates from within your opc-asyncio applications.

The PR adds several helpers for creating certificates, that follow the guidelines from OPC 10000-4 6.1 / OPC 10000-6 6.2.2 for ApplicationInstanceCertificate, for the following tasks:
* create private keys
* create self signed certficates
* create certifcate signing requests
* create ca for signing csr

The helpers are:
* generate_private_key
* generate_app_certificate_signing_request
* generate_self_signed_app_certificate
* sign_certificate_request

Test and example are included.

### DIFF
--- a/asyncua/crypto/cert_gen.py
+++ b/asyncua/crypto/cert_gen.py
@@ -1,0 +1,246 @@
+"""
+    crypothelper contains helper functions to isolate the lower level cryto stuff from the GDS client.
+"""
+
+import datetime
+
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat._oid import _OID_NAMES as OID_NAMES
+from cryptography.x509.oid import NameOID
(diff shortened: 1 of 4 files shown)


schroeder-:
Good job! Had prepared the same functions, but didn't find the time to finishe the task. Also a good idea would be to https://github.com/FreeOpcUa/opcua-asyncio/blob/master/examples/generate_certificate.sh change to a python script using these functions? Maybe add the code to same exmaples, where you check if the certificates exists and if not create a new certificate?

bitkeeper:
@schroeder- yes replacing the script would be nice.

This PR is part of a set of security related PRs:
* #1360
* #1368 
* And upcoming PR for Truslist + CRL support

Those where all required for creating a proto of a GDS push/pull client.

oroulet:
should some of these methods be exposed/installed as command line?

bitkeeper:
With these functions we could indeed create a uacert command-line tool or so. But in general, I think, it is a better practice,  to let your client/server generate it, if needed (no cert present or invalid self signed) at startup.

PR #1369, opened 2023-07-19
Add raise_on_partial_error parameter to write_values()

okapies:
`write_values()` on `Client` always throws `UaStatusCodeError` even when the [service partially succeeded](https://reference.opcfoundation.org/Core/Part4/v105/docs/5.3) and hides the rest of the error codes other than the first one. This patch adds a `raise_on_partial_error` parameter to make it returns all status codes for each write operation.


okapies:
We can use `UaStatusCodeErrors` suggested in #1370.

oroulet:
I think what you did is the best we can do with standard python., This is also what asyncio does with all there gather/test methods

PR #1370, opened 2023-07-20
Add get_children_by_path()

okapies:
Add a batch version of `get_child()` method in `Node` and `UaStatusCodeErrors` for batch operations.

(Actually it seems this is a good example of the use case of `Result` type implemented by [dry-python/returns](https://github.com/dry-python/returns) and [rustedpy/result](https://github.com/rustedpy/result) instead of using exceptions)


oroulet:
rust stuff is nice, but probably not something we want to add in a library. We have to use the standard python things

PR #1371, opened 2023-07-21
Add sync_wrapper functions

okapies:
Add `sync_uaclient_method` and `sync_async_client_method` to use the underlying asynchronous methods while leveraging `sync.Client`.

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -2,6 +2,7 @@
 sync API of asyncua
 """
 import asyncio
+import functools
 from pathlib import Path
 from threading import Thread, Condition
 import logging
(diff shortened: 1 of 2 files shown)


oroulet:
What issue does that code fix? I am a bit in a hurry but I could not understand.

okapies:
There are some functionalities not implemented in the `sync` client (both async `Client` and `UaClient`). I want to use them directly combined with the thread loop run in the sync `Client`.

oroulet:
we can have that method in addition, but why not adding the missing methods in the sync client?

okapies:
Thanks for the merge. Because I finally found that I need to use `UaClient` directly for my use cases (e.g. handling detailed statuses for each item from the OPC server) in addition to #1369 and #1370.

PR #1376, opened 2023-07-25
Alarms & Conditions

cziebuhr:
This is a continuation on https://github.com/FreeOpcUa/opcua-asyncio/pull/365.

Respective event classes and support for variables in event classes have already been added to master.

### ConditionRefresh
Events with Retain field are now stored in SubscriptionService (global) instead of MonitoredItemService (per Subscription). trigger_event methods have been extended to limit notifications to specific subscriptions or monitored items. RefreshStartEvent / RefreshEndEvent is also only sent on the subscription where ConditionRefresh got called.

### ConditionId / NodeId property
For easier handling, NodeId gets automatically added in SelectClauses, when the EventType is a ConditionType. This also works well with the history subsystem, which only calls subscribe_events and not subscribe_alarms_and_conditions.

### Defaults for properties of type ua.VariantType.NodeId
Change default to None instead of using the respective EventType, which is almost always not what is wanted.

Missing:
- Support for notifiers other than server object (HasNotifier references)
- Examples

### DIFF
--- a/asyncua/common/event_objects.py
+++ b/asyncua/common/event_objects.py
@@ -6,7 +6,7 @@ Model Uri:http://opcfoundation.org/UA/"
 Version:1.05.01"
 Publication date:2022-02-24T00:00:00Z"
 
-File creation Date:2022-09-22 16:18:40.843698"
+File creation Date:2023-07-25 09:08:45.084676"
 """
 from asyncua import ua
 from .events import Event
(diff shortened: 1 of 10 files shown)


oroulet:
I have no ideas if that code is correct but at least code looks ok and you did not break tests and added some so I merge

PR #1380, opened 2023-07-31
Fix/trailing bracket missing

AiyionPrime:
When a string start with an opening squared bracket, it is blindly assumed to end with a closing one as well.
Both characters get (also blindly) stripped from it and the remains are interpreted as an array of strings.

I propose to check for the last character as well, in order to cheaply avoid false positives, where something other than a closing bracket gets removed and a regular string gets interpreted as array of strings wrongly.

### DIFF
--- a/asyncua/common/ua_utils.py
+++ b/asyncua/common/ua_utils.py
@@ -83,7 +83,7 @@ def string_to_val(string, vtype):
     Note: no error checking is done here, supplying null strings could raise exceptions (datetime and guid)
     """
     string = string.strip()
-    if string.startswith("["):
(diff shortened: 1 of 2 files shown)


AiyionPrime:
@oroulet it appears you committed lastly in https://github.com/FreeOpcUa/opcua-asyncio/commit/5a5e423ea2b6563b4e27f90d8e1343a35b50a07b. I've got troubles finding the matching PR.

Maybe you could take a look at this.

oroulet:
I suppose this is fine..

AiyionPrime:
@oroulet Thanks, that was way faster than expected!

PR #1382, opened 2023-08-02
Add from_string() and to_string() to RelativePath

okapies:
Add `from_string()` and `to_string()` to `RelativePath` based on the [reference implementation](https://github.com/OPCFoundation/UA-.NETStandard/blob/3a378e864453e58662b47e2e3c52a61966962933/Stack/Opc.Ua.Core/Types/Utils/RelativePath.cs). Its text format conforms to [A.2 BNF of RelativePath](https://reference.opcfoundation.org/Core/Part4/v105/docs/A.2) in OPC-UA spec Part 4.

Note that this implementation still does not support non-standard `ReferenceType` (i.e., `ns!=0`) because it requires type information taken from the server to convert the passed browse name into node ID, but I think it is helpful in most use cases.

### DIFF
--- a/asyncua/ua/relative_path.py
+++ b/asyncua/ua/relative_path.py
@@ -1,0 +1,243 @@
+from enum import IntEnum
+import re
+from typing import List, Optional, Tuple
+
+from asyncua.ua.uatypes import NodeId, NodeIdType, RelativePath, RelativePathElement, QualifiedName
+from asyncua.ua.object_ids import ObjectIds, ObjectIdNames
+
+
+_NS_IDX_PATTERN = re.compile(r"([0-9]*):")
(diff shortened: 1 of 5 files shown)


schroeder-:
RelativePathElement and RelativePath are auto generated you have to add them to https://github.com/FreeOpcUa/opcua-asyncio/blob/d14c44e6e03b930b294e4adc8db2baeed8cb662b/schemas/generate_protocol_python.py#L7

okapies:
Thanks for the catch. Added.

okapies:
Maybe we should also include `#` in `_INVALID_NAME_CHARS`. OPCFoundation/UA-.NETStandard#2250

PR #1392, opened 2023-08-04
Update dev_requirements.txt

54ny7:
I think types_aiofiles was a typo. I can only install types-aiofiles module.


54ny7:
Aren't these missing?
cryptography
IPython
aiosqlite
dateutil (I cannot install it)

oroulet:
Not sure. Easy to check

PR #1399, opened 2023-08-10
Adds a truststore for certficates

bitkeeper:
This PR introduces a new class ` asyncua.crypto.truststore.TrustStore`.
With this `TrustStore` certificates can be validated against trusted certificates, trusted ca certificates and certificate revocation lists.
The TrustStore can handle multiple locations to collect trusted certificates and CRLs.

Test is included.

A subsequent PR is required to integrate `TrustStore` with `Client` and `Server` code.


oroulet:
I do not understand much of that, but code seems clean and documented and has tests so I merge

PR #1415, opened 2023-08-25
Adds a certificate validation hook to the client and server

bitkeeper:
With the hook you can register callback method with the following type `CertificateValidatorMethod`:
```
Callable[[x509.Certificate, ApplicationDescription], Awaitable[None]]
```

The method must raise an `ServiceError` exception when the certificate is invalid. The makes it possible to provide your own logic for validation.

An default implementation of such certificate validation hook is also provided by `ucrypto.validator.CertificateValidator`.

`CertificateValidator` supports checking:
- Uri
- Timerange
- Key Usage
- Extended Key Usage
- Trust
- Revoked

For the last two checks a instance of a `TrustStore` needs to be provided to `CertificateValidator`.

Which checks must be performed can be configured with `CertificateValidatorOptions`. Those are provide to the constructor of `CertificateValidator` and can be changed later with `CertificateValidator.set_validate_options`.

Examples of client and server with encryption are changed to demonstrate the use of the validator.

Basic testing of the validator is present.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -1,14 +1,18 @@
 import logging
 from enum import Enum
-from typing import Iterable, Optional, List
+from typing import Iterable, Optional, List, Tuple, TYPE_CHECKING
 
 from asyncua import ua
 from asyncua.common.session_interface import AbstractSession
 from ..common.callback import CallbackType, ServerItemCallback
 from ..common.utils import create_nonce, ServiceError
+from ..crypto.uacrypto import x509
 from .address_space import AddressSpace
 from .users import User, UserRole
 from .subscription_service import SubscriptionService
+
+if TYPE_CHECKING:
+    from .internal_server import InternalServer
 
 
 class SessionState(Enum):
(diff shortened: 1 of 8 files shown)


bitkeeper:
@oroulet & @schroeder- could you perform some review and provide feedback about it?

- Is this the correct to create the hook or is there a more prefered way?
- Should the code in `validater.py` be added to to an existing file instead?

oroulet:
still draft?

asyncua/server/internal_session.py
```diff
@@ -26,10 +30,10 @@ class InternalSession(AbstractSession):
     _counter = 10
     _auth_counter = 1000
 
-    def __init__(self, internal_server, aspace: AddressSpace, submgr: SubscriptionService, name,
+    def __init__(self, internal_server: "InternalServer", aspace: AddressSpace, submgr: SubscriptionService, name,
```
oroulet, line 33: you can probably add `if typing.TYPE_CHECKING: import InternalServer` instead of using a strring. did not test though...

oroulet:
looks ok for what I understand

schroeder-:
Looks good for me.

bitkeeper:
@oroulet no draft anymore.
It is good to go.

schroeder- (approved):

PR #1423, opened 2023-09-04
Add type hints to Client and its missing sync methods

okapies:
- Add type hints to `Client`, `Server` and `Shortcuts`
- Add type hints to `sync.Client` and `sync.Shortcuts`
- Add the following missing methods to `sync.Client`:
  - `connect_sessionless()`
  - `disconnect_sessionless()`
  - `connect_socket()`
  - `disconnect_socket()`
  - `set_locale()`
  - `get_subscription_revised_params()`
  - `get_objects_node()`
  - `get_server_node()`
  - `connect_and_find_servers()`
  - `connect_and_find_servers_on_network()`
  - `send_hello()`
  - `open_secure_channel()`
  - `close_secure_channel()`
  - `get_endpoints()`
  - `register_server()`
  - `unregister_server()`
  - `find_servers()`
  - `find_servers_on_network()`
  - `create_session()`
  - `check_connection()`
  - `server_policy_id()`
  - `server_policy_uri()`
  - `activate_session()`
  - `close_session()`
  - `get_keepalive_count()`
  - `delete_nodes()`
  - `import_xml()`
  - `export_xml()`
  - `register_namespace()`
  - `register_nodes()`
  - `unregister_nodes()`

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -3,7 +3,7 @@ import logging
 import socket
 from cryptography import x509
 from cryptography.hazmat.primitives.asymmetric.types import PrivateKeyTypes
-from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple, Type, Union, cast
 from urllib.parse import urlparse, unquote
 from pathlib import Path
 
@@ -100,7 +100,7 @@ class Client:
     __repr__ = __str__
 
     @staticmethod
-    def find_endpoint(endpoints: List[ua.EndpointDescription], security_mode: ua.MessageSecurityMode, policy_uri: str) -> ua.EndpointDescription:
+    def find_endpoint(endpoints: Iterable[ua.EndpointDescription], security_mode: ua.MessageSecurityMode, policy_uri: str) -> ua.EndpointDescription:
         """
         Find endpoint with required security mode and policy URI
         """
@@ -126,7 +126,7 @@ class Client:
             raise TypeError(f"Password must be a string, got {pwd} of type {type(pwd)}")
         self._password = pwd
 
-    def set_locale(self, locale: List[str]) -> None:
+    def set_locale(self, locale: Sequence[str]) -> None:
         """
         Sets the preferred locales of the client, the server chooses which locale he can provide.
         Normally the first matching locale in the list will be chosen, by the server.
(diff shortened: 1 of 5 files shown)


oroulet:
thank you for your work. type hints is definitely something that has been missing for a long time

asyncua/client/client.py
```diff
@@ -424,7 +427,7 @@ async def unregister_server(self, server, discovery_configuration=None):
             return await self.uaclient.unregister_server2(params)
         return await self.uaclient.unregister_server(serv)
 
-    async def find_servers(self, uris=None):
+    async def find_servers(self, uris: Optional[Iterable[str]] = None) -> Sequence[ua.ApplicationDescription]:
```
rth, line 430: @okapies My earlier comment was only about the input type to be Iterable/Sequence (since it can be anything list-like user provided). The output type is usually still a List since that is constructed in the code and so we know exactly what type it is. As it is for instance the case in that example https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#standard-duck-types

okapies: Ok. I’ll create another PR later.

okapies: @rth I later realized that the return types should have been `Sequence` instead of `List` if we don't have an intention of providing the value as a *mutable* sequence no matter what the value actually is (i.e. users have the ability to manipulate the returned list directly). What do you think?

rth: Right but that type annotation should match `type(obj)` and if that type is `list` IMO type annotations should reflect that.

For instance, in this case, it's valid to get the output of `find_servers` and delete the first element in place, no? If we make it a sequence then mypy would produce errors in downstream code, which would be rather confusing.

rth: So overall I would say if you want the output to be non-mutable, it should be some other type, rather than enforced through type annotations. Unless you saw documentation suggestions that have the opposite opinion?

okapies: You are right if you are worried about a backward compatibility issue. Someone will get mypy warnings when they already use in-place assignments or deletions for the returned list in their code.

But, IMHO, minimizing the capabilities provided to the users is generally better because it gives us a free hand in the implementation (e.g., cache the return value). Also, type annotations are just documentation that is enforceable by the static code analyzer. In that sense, they are not necessarily the same as the actual type. Using *hidden* methods on the returned value is non-compliant with the spec declared as the type hint, but users can use them anyway by suppressing warnings.

That's why I recommend you choose `Sequence` as a return type, but choosing `List` is also not a big issue, of course.

rth: No strong opinion, I can see your point. 

In any case, once you are happy with the state of types, don't forget to add a [`py.typed`](https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/) file so that downstream projects could use these.

Issue #1281, opened 2023-04-11
open_secure_channel fails with TimeoutError with stripped credentials


rth:
In v0.9.96  https://github.com/FreeOpcUa/opcua-asyncio/pull/1005 introduced,

> Strip credentials from server_url afterwards, preventing them to be sent unencrypted in e.g. send_hello.

corresponding to [this line](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/asyncua/client/client.py#L61). The problem with this change creating a connection to a [Ignition's OPCUA server](https://docs.inductiveautomation.com/display/DOC81/Ignition%27s+OPC+UA+Server) in `open_secure_channel` fails with a TimeoutError.

<details>

```py
Traceback (most recent call last):
  File "example.py", line 65, in main
    loop.run_until_complete(task(loop))
  File "example.py", line 649, in run_until_complete
    return future.result()
  File "example.py", line 39, in task
    async with client:
  File "/app/opcua-asyncio/asyncua/client/client.py", line 88, in __aenter__
    await self.connect()
  File "/app/opcua-asyncio/asyncua/client/client.py", line 287, in connect
    await self.open_secure_channel()
  File "/app/opcua-asyncio/asyncua/client/client.py", line 374, in open_secure_channel
    result = await self.uaclient.open_secure_channel(params)
  File "/app/opcua-asyncio/asyncua/client/ua_client.py", line 308, in open_secure_channel
    return await self.protocol.open_secure_channel(params)
  File "/app/opcua-asyncio/asyncua/client/ua_client.py", line 225, in open_secure_channel
    await asyncio.wait_for(self._send_request(request, message_type=ua.MessageType.SecureOpen), self.timeout)
  File "/usr/local/lib/python3.10/asyncio/tasks.py", line 458, in wait_for
    raise exceptions.TimeoutError() from exc
asyncio.exceptions.TimeoutError
```
</details>

If I remove that line, the connections works with `master` branch. With that line, I get a TimeoutError for any version later than v0.9.96 which introduced this change (including master)

If it helps debugging the `params` sent to open_secure_channel look as follows,
```
OpenSecureChannelParameters(ClientProtocolVersion=0, RequestType=<SecurityTokenRequestType.Issue: 0>, SecurityMode=<MessageSecurityMode.SignAndEncrypt: 3>, ClientNonce=b'<some-binary-string>', RequestedLifetime=3600000)
```

[...]

PR #1429, opened 2023-09-07
Make possible to not strip credentials in Client.server_url

rth:
Closes https://github.com/FreeOpcUa/opcua-asyncio/issues/1281

This adds `strip_url_credentials` parameter to `Client`, with a default to True.
Setting it to False is significantly less secure but may be necessary with some servers (in particular the Ignition server in my case)

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -108,7 +108,7 @@ class Client:
     @property
     def server_url(self):
         """Return the server URL with stripped credentials
-        
+
         if self.strip_url_credentials is True
         """
         url = self._server_url
@@ -367,7 +367,7 @@ class Client:
         """
         connect to socket defined in url
         """
-        
+
         await self.uaclient.connect_socket(self.server_url.hostname, self.server_url.port)
 
     def disconnect_socket(self):
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -34,14 +34,9 @@ class Client:
 
     _username = None
     _password = None
+    strip_url_credentials = True
 
-    def __init__(
-        self,
-        url: str,
-        timeout: float = 4,
-        watchdog_intervall: float = 1.0,
-        strip_url_credentials: bool = True
-    ):
+    def __init__(self, url: str, timeout: float = 4, watchdog_intervall: float = 1.0):
         """
         :param url: url of the server.
             if you are unsure of url, write at least hostname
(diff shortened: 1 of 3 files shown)


asyncua/client/client.py
```diff
@@ -35,7 +35,13 @@ class Client:
     _username = None
     _password = None
 
-    def __init__(self, url: str, timeout: float = 4, watchdog_intervall: float = 1.0):
+    def __init__(
+        self,
+        url: str,
+        timeout: float = 4,
+        watchdog_intervall: float = 1.0,
+        strip_url_credentials: bool = True
```
oroulet, line 43: probably useful feature but I really do not want to start adding even more parameters to the constructor for every small features. let it be a class member that user can override. You are welcome to add a line in the doc for it though..

tests/test_client.py
```diff
@@ -182,3 +182,13 @@ async def test_translate_browsepaths(server, client: Client):
...
+
+
+async def test_strip_credentials_in_url():
+    """Check that the credentials are correctly stripped in the server url"""
+
+    client = Client('opc.tcp://user:password@dummy_address:10000')
+    assert client.server_url.netloc == 'dummy_address:10000'
+
+    client = Client('opc.tcp://user:password@dummy_address:10000', strip_url_credentials=False)
+    assert client.server_url.netloc == 'user:password@dummy_address:10000'
```
oroulet, line 194:
```
  client = Client('opc.tcp://user:password@dummy_address:10000')
  client.strip_url_credentials=False
  ```

Comment (rth) [reply tests/test_client.py:194]: @oroulet thanks for the feedback! This wouldn't work though, because the stripping of the server_url happens inside __init__ and so setting this parameter after the `__init__`wouldn't help. Unless we do something like,
```
Client.strip_url_credentials=False
client = Client('opc.tcp://user:password@dummy_address:10000')
```
but that's not very pythonic. Or otherwise we would need to move this url stripping to some other method, and I'm not sure which.

oroulet: make self.server_url a property and you have solved your problem.
Might be other/better ways to do that too. like doing the stripping right before using that string is used. It cannot be used many places. probably just once in connect()

oroulet: negligible or not negligible? should only be at connect time, isnt'it?

rth: Negligible. Yeah, as far as I can tell, it only happens at connection time a few times, and 0.4µs per call is negligible with respect to the connection latency.

asyncua/client/client.py
oroulet: still I would like to remove that parameter. users can set it on client object after creation

rth: sorry I forgot; done now.

PR #1431, opened 2023-09-09
Fix to use List as a return type of the methods

okapies:
As pointed out in https://github.com/FreeOpcUa/opcua-asyncio/pull/1423#discussion_r1319602061. cc: @rth


oroulet:
correct

Issue #234, opened 2020-07-14
[Topic] Support of DataTypeDefinition


driftregion:
Hi,
I'm using `python-opcua` to interact with a couple of Siemens PLCs: an S7-1511 and an S7-1212.  I'm finding that their implementations of user-defined types are different.

## Ways to access user-defined type information
| S7-1511 | S7-1212 |
| - | - |
| `TypeDictionary`, `DataTypeDefinition` | `DataTypeDefinition` |

Apparently, the `TypeDictionary` approach has been removed in OPC-UA 1.04.  This post has more information:

https://forum.prosysopc.com/forum/opc-ua-java-sdk/get-datatype-from-siemens-s7-1500-server-to-add-to-typedictionary/#p4112 

Revision 1.04 Highlights (copied from OPC-10000-3: Address Space Model, Release 1.04)
| Mantis ID | Summary | Resolution |
| - | - | - |
| 3018 | Handling of DataType Encoding Information | Added Property to DataType NodeClass in 5.8.3 containing information about data type definition. Added DataTypes for handling the information in 8.48, 8.49, 8.50, 8.51, and 8.52. Removed the old approach having DataTypeDictionaries. This effects 5.6.2, where Properties have been removed, 5.8, where the old approach was defined in detail, and 5.5.1 as well as 7, where the ReferenceType HasDefinition and its usage was removed. The old approach is moved to an annex of Part 5 and can still be applied by OPC UA Applications. |

I'm able to read the `DataTypeDefinition` as follows, but haven't yet figured out how to parse it.
```python
type_node = self.client.get_node(variable_node.get_data_type())
dv = type_node.get_attribute(opcua.ua.attribute_ids.AttributeIds.DataTypeDefinition)
```

oroulet:
OK looks like we have some work to do. 
Can you attach content of that attribute? is it xml? or maybe base64 encoded data?

oroulet:
that sounds anyway like a better solution thatn the TypeDictonnary stuff...

driftregion:
It's either a `StructureDefinition` or an `EnumDefinition` defined in `OPC 10000-3: Address Space Model`.
 
![image](https://user-images.githubusercontent.com/20824939/87509234-b82ac100-c6a3-11ea-8a4a-09ecc2639f4a.png)

[...]

PR #1432, opened 2023-09-11
Add type hints to the Node methods

okapies:
- Add type hints to the `Node` methods
- Fix the return type of `history_read()` in `Session`
- Fix mypy errors and warnings
- Fix to use `isinstance(...)` instead of `type(...) in ...`

This is a squashed version of #1427.

### DIFF
--- a/asyncua/common/structures104.py
+++ b/asyncua/common/structures104.py
@@ -370,6 +370,7 @@ async def _get_parent_types(node: Node):
 
 async def load_custom_struct(node: Node) -> Any:
     sdef = await node.read_data_type_definition()
+    assert isinstance(sdef, ua.StructureDefinition), f"Expected StructureDefinition, got: {type(sdef)}"
     name = (await node.read_browse_name()).Name
     for parent in await _get_parent_types(node):
         parent_sdef = await parent.read_data_type_definition()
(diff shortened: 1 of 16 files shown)


asyncua/common/xmlexporter.py
```diff
@@ -35,7 +39,7 @@ class XmlExporter:
         ]
     }
 
-    def __init__(self, server, export_values: bool = False):
+    def __init__(self, server: asyncua.Server, export_values: bool = False):
```
oroulet, line 42: I have not checked but as far as I remember this is not necessarily a Server object but can also be a Client object..
so Union[asyncua.Server, asyncua.client]

asyncua/common/structures104.py
```diff
@@ -370,6 +370,7 @@ async def _get_parent_types(node: Node):
 
 async def load_custom_struct(node: Node) -> Any:
     sdef = await node.read_data_type_definition()
```
oroulet, line 372: can that method return something else that StructureDefinition?

okapies: It seems it can be `EnumDefinition` as described in https://github.com/FreeOpcUa/opcua-asyncio/issues/234#issuecomment-658570430.

oroulet:
still many missing types but that is a good work in the right direction. Thanks

okapies:
Thanks for your review!

PR #1438, opened 2023-09-12
Add type hints to SyncNode

okapies:
- Add type hints to `SyncNode`
- Fix a type hint of `Node.add_reference()`
- Add the missing methods to `SyncNode`:
    - add_reference_type()
    - delete_reference()
    - get_access_level()
    - get_description_refs()
    - get_encoding_refs()
    - get_methods()
    - get_referenced_nodes()
    - read_data_type_definition()
    - read_event_notifier()
    - register()
    - set_attr_bit()
    - set_event_notifier()
    - unregister()
    - unset_attr_bit()
    - write_array_dimensions()
    - write_data_type_definition()
    - write_value_rank()

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -1,19 +1,30 @@
 """
 sync API of asyncua
 """
+from __future__ import annotations
+
 import asyncio
+from datetime import datetime
 import functools
+import sys
 from cryptography import x509
 from pathlib import Path
 from threading import Thread, Condition
 import logging
-from typing import Any, Dict, Iterable, List, Sequence, Tuple, Type, Union, Optional
+from typing import Any, Dict, Iterable, List, Sequence, Set, Tuple, Type, Union, Optional, overload
+
+
+if sys.version_info >= (3, 8):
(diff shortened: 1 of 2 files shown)


asyncua/sync.py
```diff
@@ -285,7 +296,7 @@ def load_type_definitions(self, nodes=None):
         pass
 
     @syncmethod
-    def load_data_type_definitions(self, node: Optional["SyncNode"] = None, overwrite_existing: bool = False) -> Dict[str, Type]:  # type: ignore[empty-body]
+    def load_data_type_definitions(self, node: Optional[SyncNode] = None, overwrite_existing: bool = False) -> Dict[str, Type]:  # type: ignore[empty-body]
```
oroulet, line 299: you should also anotate the `__init__`method with reutrn value None so it gets checked by mypy. I have seen that you always skip that one into your MRs, but otherwise it looks good

okapies: Thanks. I'll put them later.

PR #1440, opened 2023-09-13
Add type hints to Subscription

okapies:
- Add type hints to `Subscription`
- Add `SubscriptionHandler` protocol class for type hinting

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -14,7 +14,7 @@ from ..common.xmlimporter import XmlImporter
 from ..common.xmlexporter import XmlExporter
 from ..common.node import Node
 from ..common.manage_nodes import delete_nodes
-from ..common.subscription import SubHandler, Subscription
+from ..common.subscription import Subscription, SubscriptionHandler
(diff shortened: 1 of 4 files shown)


oroulet:
to solve that error. use `if typing.TYPE_CHECKING: import whatever for typing`

okapies:
@oroulet Some tests failed because they pass `Node` to `subscribe_events` directly instead of `NodeId`:

https://github.com/FreeOpcUa/opcua-asyncio/blob/39e0710822ff178487b03f40fa22d8a6b8208289/tests/test_subscriptions.py#L733-L740

against the pydoc:

```python
    async def subscribe_events(
        ...
    ) -> int:
        """
        ...
        :param evtypes: ua.ObjectIds or ua.NodeId
```

Do you think we still need to support the `Nodes`, or fix the tests to pass `NodeId`?

oroulet:
We have been supporting both for ages, so we should continue with it

okapies:
Ok, I'll fix both the implementation and the type hint.

okapies:
Thank you!

PR #1442, opened 2023-09-15
Use timeout in sync code ThreadLoop post

sigmunau:


schroeder-:
I don't no much about the sync code. But there is a timeout in the async layer, this should be adjusted instead.

oroulet:
good point that there is a timeout in async layout too. it is already adjusted since we pass timeout to the async client constructor.
That MR make sure that we do not wait forever even if the sync/async engine(thread + loop) stop for some reason. But we might set that timeout to some margin higher than the async client timeout.... we want the normal timeout to trigger before that one ...

oroulet:
merged in another MR

PR #1443, opened 2023-09-15
remove potential infinite hang in sync wrapper

oroulet:


oroulet:
@sigmunau ok you think?

PR #1444, opened 2023-09-17
Draft: Update address space to 1.05.02

oroulet:
seems to break some stuff that need to be fixed somehow...


oroulet:
@ZuZuD client HR client failing with that update

@schroeder- @AndreasHeine or anyone knowing events. What is that failing test supposed to test? I have no idea what is correct. I did not change anything apart from updating address space

schroeder-:
If i find free time i can look into the event stuff. 
Ma guess would be they changed some refs types which our server or event generator doesn't support yet.

AndreasHeine:
@oroulet did they add/remove eventfields in 1.05?

oroulet:
> @oroulet did they add/remove eventfields in 1.05?

Zero ideas but it looks like something like that. It is probably just a matter of adapting the tests..

asyncua/common/event_objects.py
```diff
@@ -30,6 +31,10 @@ def __init__(self, sourcenode=None, message=None, severity=1):
         self.add_property('LocalTime', ua.uaprotocol_auto.TimeZoneDataType(), ua.VariantType.ExtensionObject)
         self.add_property('Message', ua.LocalizedText(message), ua.VariantType.LocalizedText)
         self.add_property('Severity', severity, ua.VariantType.UInt16)
+        self.add_property('ConditionClassId', None, ua.VariantType.NodeId)
+        self.add_property('ConditionClassName', None, ua.VariantType.LocalizedText)
+        self.add_property('ConditionSubClassId', None, ua.VariantType.NodeId)
+        self.add_property('ConditionSubClassName', None, ua.VariantType.LocalizedText)
```
schroeder-, line 37: These optional Fields got added to every event.

SnoopAir:
Hello, I'm started using asyncio/asyncua for the first time in a project. The given custom NodeSet depends on 1.05.02 and I wonder what is preventing this merge request from being done. 
Are there open issues/questions preventing the merge? Is there a guesstimate when it will be done? Or anything that can be helped with?

Gregor841:
My programming skills are probably to little to actively help, but i would also highly appreciate if 1.05.02 would be supported as it is blocking our development as the 05.02 is a requirement for the modules we would like to load to our OPC UA Server. 
The Python Server is just amazing for people "only" having an automation background and compared to the .NET SDK it takes 10x less time to get results.

schroeder-:
For the failing test, you need to add these event fields, that got added in 1.05.02 for every event:

     
'ConditionClassId', None, ua.VariantType.NodeId
'ConditionClassName', None, ua.VariantType.LocalizedText
'ConditionSubClassId', None, ua.VariantType.NodeId
'ConditionSubClassName', None, ua.VariantType.LocalizedText

oroulet:
> 'ConditionClassId', None, ua.VariantType.NodeId
> 'ConditionClassName', None, ua.VariantType.LocalizedText
> 'ConditionSubClassId', None, ua.VariantType.NodeId
> 'ConditionSubClassName', None, ua.VariantType.LocalizedText

Thanks, but to add where? I could not find out where. everything in event code seems to be read from  the address space. Where are the hard coded things?

schroeder-:
Maybe because they are optional?

schroeder-:
The properties get added mutliple times:

https://github.com/FreeOpcUa/opcua-asyncio/blob/8cd54934c2dbf8110c091fb5f00047cd3b623faf/asyncua/common/event_objects.py#L34-L37

and 

https://github.com/FreeOpcUa/opcua-asyncio/blob/8cd54934c2dbf8110c091fb5f00047cd3b623faf/asyncua/common/event_objects.py#L393-L396

this could be the error.

oroulet:
removing these, did not help, Did not change anything in fact. but I still do not understand anything of that event code ;-(

schroeder-:
Ok its not the lines. 
But if these lines are added, they also added multiple times in the address space. This could trigger the issue. Maybe start a simple server and look if BaseEventType and ConditionId contain both these 4 values.
Because from the logging of the test it looks like in 1 case these values are dupplicated and in another case they are not dupplicated in the list.

schroeder-:
From the spec:
The BrowseNames of the ConditionClassId, ConditionClassName, ConditionSubClassId, and ConditionSubClassName Properties originated from OPC 10000-9. They have now been generalized to also apply to Events, however to preserve backward compatibility the names have been preserved.

Maybe just use convert the list to sets and compare those?

GoetzGoerisch:
How could one help to move this forward?

PR #1445, opened 2023-09-19
Add set_attribute_value_callback() to Server

okapies:
- Add `set_attribute_value_callback()` to `Server`
- Make `value_callback()` trigger with `NodeId` and `AttributeIds` as parameters

It is a simplified version of #144 without async callback support.

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -9,7 +9,7 @@ import shelve
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
 from functools import partial
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 if TYPE_CHECKING:
     from typing import Callable, Dict, List, Union, Tuple, Generator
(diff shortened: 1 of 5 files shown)


okapies:
@oroulet I'm very sorry to bother you, but could you please give some feedback on this? I need this feature to build continuous testing for my project.

oroulet:
Sorry, that one was a bit hard to review and I am really busy currently.

oroulet:
dont' we already have some value callback? But that only works for AttibuteValue and not for any attribute?
why do you need that feature? I think I have never heard of anyone using that valuecallback mechanism..

okapies:
We have an emulator of a set of sensor data for testing our system backed by OPC Classic. The emulator gives us both value itself *and* function that provides value dynamically for each tag (equivalent to `Node` in Classic). I'm working on migrate it to OPC-UA and `value_callback` mechanism is really helpful.

> But that only works for AttibuteValue and not for any attribute?
I'm sorry but I don’t think I understand. Is there any type of attribute holds other than `AttributeValue`?

okapies:
Do you happen to have any updates on this?

okapies:
@oroulet Hi, any updates? Please let me know if you have any other concerns.

oroulet:
sorry, I still need to have a look at that code. I am far too busy at work currently....

okapies:
Ok, thanks for your reply.

asyncua/server/address_space.py
```diff
@@ -769,9 +769,10 @@ def read_attribute_value(self, nodeid: ua.NodeId, attr: ua.AttributeIds) -> ua.D
             dv = ua.DataValue(StatusCode_=ua.StatusCode(ua.StatusCodes.BadAttributeIdInvalid))
             return dv
         attval = node.attributes[attr]
+        # TODO: async support by using asyncio.iscoroutinefunction()
         if attval.value_callback:
-            return attval.value_callback()
-        return attval.value
+            return attval.value_callback(nodeid, attr)
+        return attval.value  # type: ignore[return-value] # .value must be filled
```
oroulet, line 775: why do you need to disable typing? Now way to type it?

oroulet:
OK, so I finally manage to look at that code, as far as I understand you simply added a method to use a functionality that has been there from the start. I suppose it does not hurt and we can merge

okapies:
Thanks for the merge!

PR #1454, opened 2023-09-29
Allow List,Union,Optional named fields

### LINKED ISSUE #1441: load_data_type_definitions crashes on S7-1500 UDT
I tried to use your example but it crashes at my S7-1500 UDT

```
    client = Client(url=config['opc']['server'])
    client.secure_channel_timeout = config['opc']['timeout']
    client.session_timeout = config['opc']['timeout']


    await client.connect()
    root = client.get_root_node()

    print("Objects node is: ", root)

    # Node objects have methods to read and write node attributes as well as browse or populate address space
    print(f"Children of root are:")
    for c in await root.get_children():
        print(f'\t{c} {await c.read_display_name()} {await c.read_browse_name()}')

    namespaces = await client.get_namespace_array()
    for ns in namespaces:
        print( f'Index of {ns}: {await client.get_namespace_index(ns)}' )

    print("Register namespace")
    idx = await client.register_namespace(namespaces[3])
    print("idx", idx)

    path = f'ns=3;s="S360 Main Data DB"."Offline Nest"[0]."PDR"."LPC"'
    print(f'Create node for {path}')
    node = client.get_node(path)
    dt_id = await node.read_data_type()

    dt = client.get_node(dt_id)
    dtd = await dt.read_data_type_definition()

    value = await node.read_value()
    print("*"*50)
    print(f"Node:{node}")
    print(f'Value:{value}')
    print(f'Datatype:{dt}')
    print(f'Datatypedefinition:{dtd}')
    print()
    print()

    print("*"*50)
    print("Load datatype defintions")
    await client.load_data_type_definitions()
```

Log: [opc.log](https://github.com/FreeOpcUa/opcua-asyncio/files/12611113/opc.log)

That is what Prosys is showing:

image
image
image

I am not quiet good with OPC-UA yet, so I don't know really what I could do to help you to recreate it but let me know and I will try to give you all data you need asap

Also: Can I only create definitions for a subset?

I checked #234 and tried this:

```
    node = client.get_node(path)
    dt_id = await node.read_data_type()

    dt = client.get_node(dt_id)
    dtd = await dt.read_data_type_definition()

    value = await node.read_value() 
```
But I did not understand how to access the data in a convenient pythonic way

schroeder-:
Fixes #1441


oroulet:
Are your renaming List to TypeList and Union to _TypeUnion???
I would rather modify the attribute of the model instead of modifying the name of python builtins

schroeder-:
Its only used in the exec context.  No other code is changed.

oroulet:
still if someone wants to try to read that code , and it will happen at least for debugging, then the code is going to be almost unreadable if we start changing name of builtin types to adapt to any crazy ua datamodel. We already do renaming of many of these things at different places by adding a underscore at their front and I think that is what you should do here. If a field is called List then rename it _List

schroeder-:
I just importet the typing module. So there are no collisions and the code is still readable.

oroulet:
that is OK

PR #1478, opened 2023-10-27
feat: Add typehints for methods.to_variant()

AiyionPrime:
I hope I did not miss anything.

Locally the tests run through; ~except for two (likely unrelated), which demanded two resource files~.


oroulet:
not 100% sure what the correct arg type is... but ok for now

PR #1484, opened 2023-11-06
Recursive parse parallel crawl

dpmilian:
Instead of performing requests sequentially (awaiting asyncio coroutines in a loop) parallelize them with asyncio.gather functionality.


dpmilian:
Should decrease time to run client `load_data_type_definitions()`

Issue #1364, opened 2023-07-18
sync.Clinet is not reusable


okapies:
**Describe the bug**
Call `connect()` on `sync.Client` after `disconnect()` will raise `ThreadLoopNotRunning` error.

**To Reproduce**<br />
```python
>>> client = Client('opc.tcp://localhost')
>>> client.connect()
>>> client.disconnect()
>>> client.connect()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/topaz/venv/lib/python3.10/site-packages/asyncua/sync.py", line 95, in wrapper
    result = self.tloop.post(aio_func(*args, **kwargs))
  File "/home/user/topaz/venv/lib/python3.10/site-packages/asyncua/sync.py", line 51, in post
    raise ThreadLoopNotRunning(f"could not post {coro}")
asyncua.sync.ThreadLoopNotRunning: could not post <coroutine object Client.connect at 0x7faf0b51ace0>
```

**Expected behavior**
(Re-)initialize `tloop` every time we call `connect()`.

**Version**
- Python-Version: 3.10
- opcua-asyncio Version (e.g. master branch, 0.9): 1.0.2

oroulet:
This is really common. You cannot restart a Thread object. But the error message is is really hard to understand...we need to improve that..

okapies:
Yes, but the async `Client` is reusable instead. This is problematic especially when want to combine both versions. Can we fix to start the thread when `connect` instead of `__init__`?

miki5799:
Same is happening to me and I agree with your proposed expected behaviour.

aclementptw:
I'd like to bump this, as we are currently migrating from opcua to opcua-asyncio using the wrapper functions.
We do reuse `Client`s and this poses a problem for us.

miki5799:
> Yes, but the async `Client` is reusable instead. This is problematic especially when want to combine both versions. Can we fix to start the thread when `connect` instead of `__init__`?

Hello everyone, 

I have submitted a PR, in which a new thread is started analogous to `__init__` iff, the ThreadLoop has stopped, so that the updated `connect` method looks like this:
```python
def connect(self) -> None:
    if not self.tloop.is_alive():
        self.tloop = ThreadLoop()
        self.tloop.start()
        self.close_tloop = True
    self.tloop.post(self.aio_obj.connect())
```
Analogous for the Server's `start` method.

PR #1503, opened 2023-11-23
Make Sync Client and Server reusable

miki5799:
As the issue https://github.com/FreeOpcUa/opcua-asyncio/issues/1364 suggests, OpcUa Client (and Server) are not reusable, since disconnecting the client or stopping the server, stops the Sync Wrapper's ThreadLoop respectively and therefore lead to ThreadLoopNotRunnning exceptions on any call that depends on the ThreadLoop.

As in the discussion of https://github.com/FreeOpcUa/opcua-asyncio/issues/1364 the Sync Wrapper should start the ThreadLoop on connection (or server.start) if ThreadLoop is not alive anymore.
This proposed behaviour has been implemented in this PR.


asyncua/sync.py
```diff
@@ -249,9 +249,12 @@ def application_uri(self):
...
         self.aio_obj.application_uri = value
 
-    @syncmethod
     def connect(self) -> None:
-        pass
+        if not self.tloop.is_alive():
+            self.tloop = ThreadLoop()
+            self.tloop.start()
+            self.close_tloop = True
+        self.tloop.post(self.aio_obj.connect())
```
oroulet, line 257: we cannot have the same code several places. At least put that code in a method. 
also if you want to delay threadloop initializatin, you might need to do it at several places not only here... there is a connect_sessionless method for example and maybe others...

asyncua/sync.py
```diff
@@ -592,9 +595,12 @@ def register_namespace(self, url):
...
         pass
 
-    @syncmethod
     def start(self):
-        pass
+        if not self.tloop.is_alive():
+            self.tloop = ThreadLoop()
+            self.tloop.start()
+            self.close_tloop = True
+        self.tloop.post(self.aio_obj.start())
```
oroulet, line 603: arrg same code again... Making that class re-usable might not be a good idea.  there is probably a reason, threading.Thread is not reusable...

miki5799: Thanks for your time and effort, but could you clarify further, since I'm not quite sure in what way initializing a new Thread or ThreadLoop could pose a problem in this context.

oroulet: I was simply refering to the fact the threading.Thread in the python standard library does not allow re-use.  We copied their API when creating that class. We can make that class reusable, but I see your code over is not really clean, so maybe it is hard to solve that issue in a safe way...

miki5799:
Superseded by #2010. The new approach keeps a single ThreadLoop alive instead of replacing it and separates connection teardown from final client cleanup.

PR #1528, opened 2023-12-16
CI stuff

oroulet:


AiyionPrime:
I reproduced the same behaviour locally on a python 3.12 without performance optimizations: Same issue.

I did a binary search on the tests and found (at least) two which are 'helping' the failing test not to run into a timeout.

I think its the subscription line in e.g. `test_set_monitoring_mode`:
    await opc.opc.create_subscription(100, None)
    
Removing all other except the failing test and this snippet:
```python
@pytest.mark.parametrize("opc", ["client"], indirect=True)
async def test_set_monitoring_mode(opc):
    await opc.opc.create_subscription(100, None)
```

leads to your found tested issue.

I'm not sure yet, what change in python 3.12 introduced the problem, but I thought sharing this early might help others seeing the problem faster than me.

oroulet:
unfortunately I think  the test got removed at some point since it worked so random
. Current tests pass on 3.14 at least

Issue #1519, opened 2023-12-08
get_children() memory leak or runaway behaviour in asyncua 1.0.5 


magnetohack:
**Describe the bug** <br />
get_children() appears to have a memory leak or runaway behaviour for asyncua version 1.0.5 that was not present in version 1.0.4.

**To Reproduce**<br />
Steps to reproduce the behavior incl code:

Example code for traversing down to the relevant data nodes. Last call to get_children() fails.
<pre>
# Connect and traverse tree to locate top node for sensor data
client = Client(url)
await client.connect()

# Get a list of all 'sensor data' nodes
root = client.get_root_node()
children = await root.get_children()
c85 = await children[0].get_children()
c851 = await c85[1].get_children()
allsensors = await c851[2].get_children()
</pre>

The last get_children() call should return about 83000 items. Using asyncua version 1.0.4 this works fine, while in version 1.0.5 it consumes all available memory before failing.

**Expected behavior**<br />
Expect the get_children() call to return list of child nodes without crashing or consuming all memory.

**Version**<br />
Python-Version: Tested on 3.9 and 3.11. Problem in both cases<br />
opcua-asyncio Version 1.0.5 (current default with pip install)

oroulet:
rather strange, I do not see any other changes than typing in node.py and client code since 1.0.4...

oroulet:
@magnetohack I went again through all commits but I really cannot find any code that can be related to that behaviour. Do you know how to use git bisect? it would be great if you could run that between 1.0.4 and and master to find out where that behaviour has changed

joshbode:
I am also seeing this issue for 1.0.5, i.e. getting children and rapidly consuming all available memory - I will try to see what might be causing it.
1.0.4 does not appear to have this issue.

joshbode:
it's something in daa74c2.
0cd54642 doesn't have the issue.

joshbode:
looks like the line:
```
        while head.ContinuationPoint:
```
is possibly the problem - `head` doesn't get updated in the loop

schroeder-:
Your change is correct. I would add the None check. Would be nice if you could make a pr with this changes.

oroulet:
we got it by the classical "Hide code change into a typing or formatting commit". I should really be more hard on these things

oroulet:
[...]

PR #1531, opened 2023-12-18
Fix infinite loop causing memory leak in GH-1519

joshbode:


joshbode:
I don't know the codebase/protocol well enough to know whether we can always expect there to be at least one value in the iterator, or whether we need to have a fallback and a check, e.g.

```python
head = next(iter(results), None)
if head is None:
    break
```

asyncua/common/node.py
```diff
@@ -450,6 +450,7 @@ async def _browse_next(self, results: Iterable[ua.BrowseResult]) -> List[ua.Refe
             params.ContinuationPoints = [head.ContinuationPoint]
             params.ReleaseContinuationPoints = False
             results = await self.session.browse_next(params)
+            head = next(iter(results))
```
oroulet, line 453: yes we should check that next does not return None

oroulet, line 453: but something else i wrong here. The original code only handled results[0] and this is correct since we only sent one request to server. Now that check is gone...

oroulet, line 453: I propose we just revert to original code, I do not see anything wrong in it (apart from getting a strange error if result does not contain at least one result., it should but...)

oroulet:
I reverted change in that other MR: https://github.com/FreeOpcUa/opcua-asyncio/pull/1534.
Thank for you help @joshbode !

joshbode:
Sounds good, @oroulet :)

PR #1545, opened 2024-01-04
crypto change subject alternate errors to warning

### LINKED ISSUE #1544: Lower Certificate Error in Logs
Hi! 

I have been seeing the following `ERROR` log message when using `asyncua`:
```
[ERROR] [asyncua.crypto.uacrypto]: certificate does not contain the application uri (urn:freeopcua:client). Most applications will reject a connection without it.
[ERROR]  [asyncua.crypto.uacrypto]: certificate does not contain the hostname in DNSNames eeeccc7d53d4. Some applications will check this.
```

However, for my specific application, the certificates I am using work and allow communication with the OPCUA Server. Therefore, I think this log level should be `WARNING` instead of `ERROR` since it does not cause an error in ALL applications.

Thanks!

schroeder-:
Only log a warning if the subject alternate name is not as the opc ua specs.
Fixes #1544


alliehunter:
Hi! I was wondering when this would be released to use? I am not seeing this change on the latest version. Thank you!

PR #1555, opened 2024-01-21
Configure ThreadLoop to be daemon

eskildsf:
Configuring a sync Client object and connecting it to an OPC-UA host will cause the calling script to hang on exit if disconnect() is not called. By setting the ThreadLoop to be a daemon resolves this issue.

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -34,7 +34,7 @@ class ThreadLoopNotRunning(Exception):
 
 class ThreadLoop(Thread):
     def __init__(self, timeout: Optional[float] = 120) -> None:
-        Thread.__init__(self)
+        Thread.__init__(self, daemon=True)
         self.loop = None
         self._cond = Condition()
         self.timeout = timeout


oroulet:
Is it safe?
not correctly closing things is often a very bad ideas. But maybe in that particular case it is worth it... not sure

eskildsf:
Hi @oroulet 

I don't think I can say "yes" confidently. You probably have a better basis to make that call.

I will say though that it is quite useful for the Python session not to hang if you have an exception which causes a crash. That is probably the most relevant usecase since I do agree that in the best of all worlds the programmer would make the call to `disconnect` deliberately.

Br Eskild

still-wondering:
hi, @eskildsf and @oroulet,

Maybe a bit safer way to fix it would be to also add a method to a `ThreadLoop` like this (in addition to the `daemon=True`):
```py
    def _at_exit(self) -> None:
        if self.is_alive():
            self.stop()
```
and register it with `atexit` in `ThreadLoop.start()`:
```diff
     def start(self) -> None:
         with self._cond:
+            atexit.register(self._at_exit)
             Thread.start(self)
             self._cond.wait()
```
This will cause the interpreter not to wait for the thread before exiting, but the atexit callback will be called which will join the thread eventually. It is still not perfect and some tasks may not finish (resulting e.g. in the logs below), but this will let the current batch of callbacks in the loop to finish, and the loop (and therefore the thread) to stop gracefully.
<details>
  <summary>Failure logs</summary>

  ```
  [ERROR] 2026-07-27T12:48:22+0000 asyncio: Task was destroyed but it is pending!
  task: <Task pending name='Task-3' coro=<Client._connection_supervisor() running at /root/my-opcua-asyncio/asyncua/client/client.py:873> wait_for=<Future pending cb=[Task.task_wakeup()]>>
  [ERROR] 2026-07-27T12:48:22+0000 asyncio: Task was destroyed but it is pending!
  task: <Task pending name='Task-4' coro=<Client._stale_watchdog_loop() running at /root/my-opcua-asyncio/asyncua/client/client.py:917> wait_for=<Future pending cb=[Task.task_wakeup()]>>
  Exception ignored in: <coroutine object Client._connection_supervisor at 0x7f241b3d9b10>
    File "/root/my-opcua-asyncio/asyncua/client/client.py", line 882, in _connection_supervisor
    File "/usr/lib/python3.13/asyncio/tasks.py", line 712, in sleep
  RuntimeError: no running event loop
  [ERROR] 2026-07-27T12:48:22+0000 asyncio: Task was destroyed but it is pending!
  task: <Task pending name='Task-2' coro=<Client._renew_channel_loop() running at /root/my-opcua-asyncio/asyncua/client/client.py:1091> wait_for=<Future pending cb=[Task.task_wakeup()]>>
  ```

</details>

<details>
  <summary>Code to reproduce</summary>

  ```py
  opcua_client = Client(...)
  opcua_client.connect()
  raise RuntimeError() 
  ```

</details>

PR #1566, opened 2024-02-07
Add `set_attribute_value_setter()` method to server

FMeinicke:
This method allows to set a custom setter function on a per-attribute basis. This setter function will then be called by `write_attribute_value()` instead of writing the attribute value directly. This allows for custom logic to be implemented by the user (such as checking if the value to be written is within a certain range) and if not, an appropriate error can be raised from that setter, preventing the write and notifying the client about the occurred error.

Also, see discussion #1563 for some more background info about the use case of this method.


oroulet:
let's try...

PR #1571, opened 2024-02-08
Add support for coroutine functions as attribute value setters

FMeinicke:
The previous implementation only allowed sync functions to be registered as an attribute value setter. This limits the setter function in a way such that it cannot read from other nodes using `Node.read_value()`, for example. Allowing coroutine functions to be set as attribute value setters enables these functions to also use coroutine functions themselves.

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -792,7 +792,10 @@ class AddressSpace:
             return ua.StatusCode(ua.StatusCodes.BadTypeMismatch)
 
         if attval.value_setter is not None:
-            attval.value_setter(node, attr, value)
+            if asyncio.iscoroutinefunction(attval.value_setter):
+                await attval.value_setter(node, attr, value)
+            else:
+                attval.value_setter(node, attr, value)
         else:
             attval.value = value
         attval.value_callback = None


AndreasHeine:
i remember in the past that writing should be sync in order to be consistent!? long time ago... @oroulet do you remember?

AndreasHeine:
you mean like the range from a AnalogItem?

FMeinicke:
Thank you for your suggestions. I'll try to make it work that way instead.

AndreasHeine:
guess via nodeid as part of the setter:

```python
def setter(node_data: NodeData, attr: ua.AttributeIds, value: ua.DataValue) -> None:
    range_high = "ns=1;i=1000"
    range_high = "ns=1;i=1000"

    # we need to access servers addressspace (as long as we read only it would be ok)
    high_data = server.iserver.aspace.get(range_high)
    low_data = server.iserver.aspace.get(range_low)

    if value.Value.Value > some_other_value.Value.Value:
        raise ua.userrors.BadOutOfRange()
    node_data.attributes[attr].value = value
```

maybe a wrapper class for those variable types would be a good idea to implement the semantics

oroulet:
> # we need to access servers addressspace (as long as we read only it would be ok)
    high_data = server.iserver.aspace.get(range_high)
    low_data = server.iserver.aspace.get(range_low)
    
    Should be ok to write as long as you are using syn calls

oroulet:
fancy.... and when you look at code, there is zero reason for them to be async, they do not await anything.
InternalServer.write_attribute_value must be async for compatibiity with client version but you can try to make a MR to make that method sync in AddressSpace, then you can call it directly..

FMeinicke:
> and when you look at code, there is zero reason for them to be async, they do not await anything.

As I see it, there is a reason for `AddressSpace.write_attribute_value()` to be async: It's because it calls the `AttributeValue`'s `datachange_callbacks` which are async:

https://github.com/FreeOpcUa/opcua-asyncio/blob/8cf8682a019bacc7e67c42fec72531bf3e87c31e/asyncua/server/address_space.py#L800-L804

So, I don't think there is an easy way to make `AddressSpace.write_attribute_value()` sync. Or am I missing something here?

oroulet:
arrg did not see that one. I am not sure how to change that. . but is writing in another node a good idea when writing into a  node? on server side you should better use subscription and react to a value change to change another node value.
Reading to ensure some value validatino has a clear use case

FMeinicke:
> on server side you should better use subscription and react to a value change to change another node value.

Very good point. I somehow didn't think of using subscriptions for this use case. I agree that this is indeed the better way instead of writing to another node from within the value setter of a node.

Issue #1313, opened 2023-05-22
Server shelf_file not working


harriv:
**Describe the bug** <br />

When using the shelf_file in server init, an exception is raised:

> Traceback (most recent call last):
>   File "test_server.py", line 20, in <module>
>     asyncio.run(main(), debug=True)
>   File "Python311\Lib\asyncio\runners.py", line 190, in run
>     return runner.run(main)
>            ^^^^^^^^^^^^^^^^
>   File "Python311\Lib\asyncio\runners.py", line 118, in run
>     return self._loop.run_until_complete(task)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "Python311\Lib\asyncio\base_events.py", line 653, in run_until_complete
>     return future.result()
>            ^^^^^^^^^^^^^^^
>   File "test_server.py", line 9, in main
>     await server.init(shelf_file="test")
>   File "opc_env\Lib\site-packages\asyncua\server\server.py", line 122, in init
>     await self.iserver.init(shelf_file)
>   File "opc_env\Lib\site-packages\asyncua\server\internal_server.py", line 80, in init
>     await self.load_standard_address_space(shelffile)
>   File "opc_env\Lib\site-packages\asyncua\server\internal_server.py", line 122, in load_standard_address_space   
>     is_file = await asyncio.get_running_loop().run_in_executor(
>               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "Python311\Lib\concurrent\futures\thread.py", line 58, in run
>     result = self.fn(*self.args, **self.kwargs)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "Python311\Lib\pathlib.py", line 1267, in is_file
>     return S_ISREG(self.stat().st_mode)
>                    ^^^^^^^^^
> AttributeError: 'str' object has no attribute 'stat'
> 

My code:

```
from asyncua import Server, ua
from asyncua.common.methods import uamethod
from asyncua.common.structures104 import new_enum

import asyncio

async def main():
    server = Server()    
    await server.init(shelf_file="test")
    server.set_endpoint("opc.tcp://0.0.0.0:54840/server/")
    server.set_server_name("Test server")
    uri = "http://opcua.local"
    idx = await server.register_namespace(uri)
    main = await server.nodes.objects.add_object(idx, "Main")
[...]

PR #1574, opened 2024-02-12
test: Verify loading from shelf files raises an AttributeError

AiyionPrime:
> This confirms @harriv's finding in #1313.

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -187,6 +187,15 @@ def test_sync_server_get_node(server, idx):
     assert vars[0].read_value() == 6.7
 
 
+@pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


AiyionPrime:
@oroulet any objections against merging this test?

oroulet:
I am not really following here, you want to merge a test that is disabled? What is your goal?

AiyionPrime:
This tests that storing the shelf indeed fails, as @harriv tested manually.
The following PR will resolve the issue.

It's worth noting, that this does not disable the test, but marks it as expected failure.
One of the next PRs will then resolve the issue and remove the line of expected failure.

But it's not disabled, it runs and show the problem.


When I started this I expected the problem to be bigger than it effectively was.
This PR is ready to merge and I will rebase the others accordingly afterwords.

PR #1576, opened 2024-02-13
fix: Allow creating shelf files again

AiyionPrime:


oroulet:
I am fine with these things, I have never used that shelv thing, so if it work for you I am fine with merging it

tests/test_sync.py
```diff
@@ -187,9 +187,6 @@ def test_sync_server_get_node(server, idx):
     assert vars[0].read_value() == 6.7
 
 
-@pytest.mark.xfail(
-        raises=AttributeError, reason="asyncua introduced a regression, likely when we switched to pathlib", strict=True
```
AiyionPrime, line 191: Now this strict=True would have led your pipeline to fail, if I resolved the issue, without removing this `xfail` marker again.

All in all this is a rather clean structure for test driven development with pytest, which I've learned to value over time.

PR #1577, opened 2024-02-13
test: Migrate deactivated unit to pytest

AiyionPrime:
There exists a unittest, which could attest that loading shelf files does currently not work.
This migrates it from unittest to the new pytest format and alters the shelf_file invocation which was provided in `__init__()` before and is now done in `init()`.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -3,6 +3,9 @@ Run common tests on server side
 Tests that can only be run on server side must be defined here
 """
 import asyncio
+import shelve
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


AiyionPrime:
@oroulet if there's anything unclear, let me know.
Otherwise I think this is good to go as well.

It verifies loading a shelf does currently not work either, as the old test did before, but allows it to execute and verify it really isn't; similar to the test in #1574 before.

AiyionPrime:
@oroulet Would you mind reviewing this?

PR #1601, opened 2024-03-10
deprecation: replce utcnow() with now(UTC)

oroulet:


alexrudd2:
`datetime.UTC` is only available in [Python 3.11](https://docs.python.org/3/library/datetime.html#datetime.UTC)

PR #1623, opened 2024-04-24
ci:use pytest-xdist to parallelize tests

alexrudd2:
https://pypi.org/project/pytest-xdist/


oroulet:
Does it handle having some servers run on the same port?
At least it looks like the failed tests are releated to something different....

alexrudd2:
It doesn't do anything automatically (besides run tests in parallel).  So the tests might have to be changed to use different ports.

In the past I've used `@pytest.mark.parameterize` to do this but I'm not sure how these tests are setup.

alexrudd2:
What about binding to port `0`, which will cause the OS to assign a random available port? After the server is created, the assigned port can be determined with `srv.bserver.port` and given to the client.

PR #1631, opened 2024-04-29
Update python versions in setup.py

alexrudd2:


oroulet:
We should add 3.12 then...

alexrudd2:
I figured that would go in https://github.com/FreeOpcUa/opcua-asyncio/pull/1528

Issue #116, opened 2019-12-17
Memory balloons after "Subscription Subscription(id:<id>) has expired."


dannmartens:
We've noticed that in our production environment, where we run multiple instances (19 right now) of asyncua-based OPC UA servers, some instances enter some strange mode from which recovery is difficult. A symptom of this runaway mode is memory ballooning until the limit is reached.    

From the moment the issue manifests itself, the logs are blasted with the following statements, following "the session id is not valid" error:
```
2019-12-17 15:26:46 [ERROR] asyncua.server.uaprocessor - sending service fault response: The session id is not valid. (BadSessionIdInvalid)
2019-12-17 15:31:25 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:93) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:25 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:86) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:25 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:93) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:25 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:86) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:25 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:93) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:25 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:86) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:26 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:93) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:26 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:86) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:26 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:93) has expired, publish cycle count(101) > lifetime count (100)
2019-12-17 15:31:26 [WARNING] asyncua.server.internal_subscription - Subscription Subscription(id:86) has expired, publish cycle count(101) > lifetime count (100)
[...]

PR #1655, opened 2024-06-11
Add server-side session watchdog

chrisjbremner:
As discussed [here](https://github.com/FreeOpcUa/opcua-asyncio/issues/116#issuecomment-2155642372), to comply with [section 5.6.2](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.6.2) of the spec, the server should close a session that is inactive within its session timeout:

> [Sessions](https://reference.opcfoundation.org/search/17?t=Sessions) are terminated by the [Server](https://reference.opcfoundation.org/search/17?t=Server) automatically if the [Client](https://reference.opcfoundation.org/search/17?t=Client) fails to issue a [Service](https://reference.opcfoundation.org/search/17?t=Service) request on the [Session](https://reference.opcfoundation.org/search/17?t=Session) within the timeout period negotiated by the [Server](https://reference.opcfoundation.org/search/17?t=Server) in the [CreateSession](https://reference.opcfoundation.org/search/17?t=CreateSession) [Service](https://reference.opcfoundation.org/search/17?t=Service) response. This protects the [Server](https://reference.opcfoundation.org/search/17?t=Server) against [Client](https://reference.opcfoundation.org/search/17?t=Client) failures and against situations where a failed underlying connection cannot be re-established. [Clients](https://reference.opcfoundation.org/search/17?t=Clients) shall be prepared to submit requests in a timely manner to prevent the [Session](https://reference.opcfoundation.org/search/17?t=Session) from closing automatically.

I no longer think this will actually resolve anything in https://github.com/FreeOpcUa/opcua-asyncio/issues/116, but I think it's useful nonetheless if we want to comply with the spec.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -1,6 +1,6 @@
 import logging
 from enum import Enum
-from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple
+from typing import Iterable, Optional, List, Tuple, TYPE_CHECKING
 
 from asyncua import ua
 from asyncua.common.session_interface import AbstractSession
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -1,6 +1,9 @@
+import asyncio
 import logging
+from datetime import datetime, timedelta
 from enum import Enum
-from typing import Iterable, Optional, List, Tuple, TYPE_CHECKING
+from functools import wraps
+from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple
 
 from asyncua import ua
 from asyncua.common.session_interface import AbstractSession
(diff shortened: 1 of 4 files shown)


asyncua/server/internal_session.py
```diff
@@ -59,13 +67,44 @@ async def get_endpoints(self, params=None, sockname=None):
     def is_activated(self) -> bool:
         return self.state == SessionState.Activated
 
+    def update_last_action_time(func):
```
oroulet, line 70: I understand what you want to do, but I have bad experience with that kind of decorators. they make stack trace a mess and completlu breaks typing checks. if we want t omerge that we at least need typing and check things with mypy (which probbaly only does not complain because since you do not have typing on decorator so it disabled itself)

oroulet: maybe it is better to update some timestamp at lower level when we receive a request which is dispatched to thata internal session. IS that possible?

chrisjbremner: Good idea, I moved it into the UA processor

PR #1654, opened 2024-06-11
Stop expired subscriptions

chrisjbremner:
Addresses https://github.com/FreeOpcUa/opcua-asyncio/issues/1267 and https://github.com/FreeOpcUa/opcua-asyncio/issues/116. As discussed in [this comment](https://github.com/FreeOpcUa/opcua-asyncio/issues/116#issuecomment-2155642372), in order to comply with [section 5.13.1.1](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.13.1.1) part 8 of the spec, we should close the subscription and delete all monitored items:

[...]

PR #1656, opened 2024-06-12
Add callback to delete subscriptions

chrisjbremner:
As discussed [here](https://github.com/FreeOpcUa/opcua-asyncio/pull/1654#issuecomment-2162116152), we need to clean up some subscription objects when a subscription deletion within `internal_subscription.py` due to an exceeded `RevisedLifetimeCount`. In this PR, I implement a callback that cleans up the subscription objects in the subscription service that is called when the subscription is stopped in this fashion.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -85,7 +85,13 @@ class InternalSession(AbstractSession):
         if InternalSession._current_connections < 0:
             InternalSession._current_connections = 0
         self.state = SessionState.Closed
-        await self.delete_subscriptions(list(self.subscription_service.subscriptions.keys()))
+        await self.delete_subscriptions(
+            [
+                id
+                for id, sub in self.subscription_service.subscriptions.items()
+                if sub.session_id == self.session_id
+            ]
+        )
 
     def activate_session(self, params, peer_certificate):
         self.logger.info('activate session')
@@ -195,7 +201,7 @@ class InternalSession(AbstractSession):
         return await self.iserver.method_service.call(params)
 
     async def create_subscription(self, params, callback, request_callback=None):
-        result = await self.subscription_service.create_subscription(params, callback, request_callback=request_callback)
+        result = await self.subscription_service.create_subscription(params, callback, self.session_id, request_callback=request_callback)
         return result
 
     async def create_monitored_items(self, params: ua.CreateMonitoredItemsParameters):
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -196,7 +196,12 @@ class InternalSession(AbstractSession):
         return await self.iserver.method_service.call(params)
 
     async def create_subscription(self, params, callback, request_callback=None):
-        result = await self.subscription_service.create_subscription(params, callback, request_callback=request_callback)
+        result = await self.subscription_service.create_subscription(
+            params,
+            callback,
+            request_callback=request_callback,
+            delete_callback=self.delete_subscriptions,
+        )
         self.subscriptions.append(result.SubscriptionId)
         return result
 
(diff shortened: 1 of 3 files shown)


asyncua/server/internal_session.py
```diff
@@ -44,7 +44,6 @@ def __init__(self, internal_server: "InternalServer", aspace: AddressSpace, subm
         self.state = SessionState.Created
         self.session_id = ua.NodeId(self._counter)
         InternalSession._counter += 1
-        self.subscriptions = []
```
chrisjbremner, line 47: This didn't appear to be referenced anywhere, so it felt save to use the subscription service's version

asyncua/server/internal_subscription.py
```diff
@@ -116,7 +129,14 @@ async def publish_results(self, requestdata=None):
                                 self._publish_cycles_count, self.data.RevisedLifetimeCount)
             # FIXME this will never be send since we do not have publish request anyway
             await self.monitored_item_srv.trigger_statuschange(ua.StatusCode(ua.StatusCodes.BadTimeout))
-            await self.stop()
```
chrisjbremner, line 119: Using `self.stop()` directly was tricky due to the context of the subscription task and `stop` trying to directly await the task. Instead, I opted to decouple the stopping of the subscription from the task by setting `self._closing` which will immediately terminate the task because it gets checked after exiting this function.

asyncua/server/subscription_service.py
```diff
@@ -26,15 +26,23 @@ def __init__(self, aspace: AddressSpace):
...
         result.RevisedMaxKeepAliveCount = params.RequestedMaxKeepAliveCount
         self._sub_id_counter += 1
         result.SubscriptionId = self._sub_id_counter
-        internal_sub = InternalSubscription(result, self.aspace, callback, request_callback=request_callback)
+        internal_sub = InternalSubscription(
+            result,
+            self.aspace,
+            callback,
+            request_callback=request_callback,
+            delete_callback=lambda: self.subscriptions.pop(result.SubscriptionId, None),
```
chrisjbremner, line 44: Alternatively the subscription service could be passed to each internal subscription and the dict could be modified there, but this seemed lighter.

chrisjbremner:
I tested this locally by setting my RevisedLifetimeCount to 0 on my subscription and changing the `>` in [this line](https://github.com/FreeOpcUa/opcua-asyncio/blob/6dce2409c7ba75e11c6a4a96178e68b9cb7e2b9a/asyncua/server/internal_subscription.py#L114) to `>=`, which then triggered this error, and I confirmed the subscription was successfully "deleted". I did note that before my changes here, the memory leak was not fully fixed by #1654 because `await self.stop()` did not fully behave the way I expected due to a suppressed `CancelledError`

asyncua/server/internal_session.py
```diff
@@ -86,7 +85,13 @@ async def close_session(self, delete_subs=True):
         if InternalSession._current_connections < 0:
             InternalSession._current_connections = 0
         self.state = SessionState.Closed
-        await self.delete_subscriptions(self.subscriptions)
+        await self.delete_subscriptions(
```
chrisjbremner, line 88: Instead of maintaining a list of subscriptions in two places that would need to be updated, store the session ID in the internal subscription so that it can be fetched directly. Makes managing the lists much easier at the expense of a slight efficiency hit on a function that's not called very often.

asyncua/server/internal_session.py
```diff
@@ -86,7 +85,13 @@ async def close_session(self, delete_subs=True):
...
+            [
+                id
+                for id, sub in self.subscription_service.subscriptions.items()
+                if sub.session_id == self.session_id
+            ]
+        )
```
oroulet, line 94: I would really make a method get_subscription_by_id(id) or similar. but OK maybe this is fine..

chrisjbremner:
Would it make sense to do a new release soon to include these changes?

PR #1661, opened 2024-06-19
Resolve timezone deprecation warnings

chrisjbremner:
I see lots of deprecation warnings when running tests in Python 3.12:

```
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    await self.current_time_node.write_value(datetime.utcnow())
```

Instead of the suggested `datetime.datetime.now(datetime.UTC)`, I chose to use `datetime.datetime.now(datetime.timezone.utc)` since the latter is compatible with Python 3.7 whereas the former is only compatible with Python 3.11.


setup.py
```diff
@@ -23,7 +23,7 @@
     packages=find_packages(exclude=["tests"]),
     provides=["asyncua"],
     license="GNU Lesser General Public License v3 or later",
-    install_requires=["aiofiles", "aiosqlite", "python-dateutil", "pytz", "cryptography>40.0.1", "sortedcontainers", "importlib-metadata;python_version<'3.8'", "pyOpenSSL>23.2.0", "typing-extensions"],
+    install_requires=["aiofiles", "aiosqlite", "python-dateutil", "pytz", "cryptography>42.0.0", "sortedcontainers", "importlib-metadata;python_version<'3.8'", "pyOpenSSL>23.2.0", "typing-extensions"],
```
chrisjbremner, line 26: Should already be this since `not_valid_after_utc` is used [here](https://github.com/FreeOpcUa/opcua-asyncio/blob/8537ae2660e5982f686e28220019489a512db34b/asyncua/crypto/uacrypto.py#L334), but this function was added in `cryptography==42.0.0`

oroulet:
good work. I never found the 3.7 compatible hack

oroulet:
but looks like you modified the autogenerate files instead of modifying the script generating them https://github.com/FreeOpcUa/opcua-asyncio/blob/master/schemas/generate_protocol_python.py

PR #1663, opened 2024-06-19
Patch statemachine to work with example

tmcguc:
We modified the library so the state machine works with the example provided

### DIFF
--- a/my-changes.patch
+++ b/my-changes.patch
@@ -1,0 +1,36 @@
+From 7a5311b4f1094dff434df43acc46ead4d1b2d7a0 Mon Sep 17 00:00:00 2001
+From: Dr Terrence McGuckin <terrence@ephemeron-labs.com>
+Date: Wed, 19 Jun 2024 18:55:18 -0400
+Subject: [PATCH] changed sytate machine to work with example
+
+---
+ asyncua/common/statemachine.py | 6 ++++--
+ 1 file changed, 4 insertions(+), 2 deletions(-)
+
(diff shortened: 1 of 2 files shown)


my-changes.patch
```diff
@@ -0,0 +1,36 @@
+From 7a5311b4f1094dff434df43acc46ead4d1b2d7a0 Mon Sep 17 00:00:00 2001
```
oroulet, line 1: what is that file?

tmcguc:
That was added by mistake.  I’ll remove it.

Dr. Terrence McGuckin
(they/them)
215 531 4703
Founder
Ephemeron Labs
3525 I street
Suite 07
Philadelphia, PA 19134

AndreasHeine:
@oroulet @tmcguc 

As sidenote:
StateMachineType is not Abstract. So it can be instantiated!

StateVariableType: https://reference.opcfoundation.org/Core/Part5/v104/docs/B.4.3
HasProperty | Variable | Id | BaseDataType | PropertyType | Mandatory

>BaseDataType!!! so it can be any BaseDataType / Subtype of it e.g. NodeId

[Id ](https://reference.opcfoundation.org/search/21?t=Id)is a name which uniquely identifies the current state within the [StateMachineType](https://reference.opcfoundation.org/search/21?t=StateMachineType). A subtype may restrict the [DataType](https://reference.opcfoundation.org/search/21?t=DataType).

I chose NodeId at that time because the State-ObjectType-Instance has a unique id (node-id) in the Addressspace!

i am not sure if the change is valid!?

PR #1669, opened 2024-07-01
Uaerr

oroulet:


asyncua/ua/uaerrors/_base.py
```diff
@@ -107,3 +107,15 @@ class UaStructParsingError(UaError):
 
 class UaInvalidParameterError(UaError):
     pass
+
+
+class UaClientRenewChannelError(UaError):
+    pass
+
+
+class UaClientConnectionLostError(UaError):
```
sigmunau, line 116: Maybe have this also inherit ConnectionError for any existing code out there that excepts that?

asyncua/client/client.py
```diff
@@ -531,13 +532,22 @@ async def check_connection(self) -> None:
...
+                    await self._renew_channel_task
+                except Exception as ex:
+                    raise UaClientRenewChannelError() from ex
         if self._monitor_server_task is not None:
             if self._monitor_server_task.done():
-                await self._monitor_server_task
+                try:
+                    await self._monitor_server_task
+                except Exception as ex:
+                    raise UaClientConnectionLostError() from ex
```
sigmunau, line 544: Maybe change the order of these two blocks. If the monitor task determines that the connection is lost, any errors in the renew channel task is most likely due to the lost connection?

PR #1690, opened 2024-07-31
Fixed function signature in `Node`

tanekere:
Fixed function signature of `read_array_dimensions` in asyncua/common/node.

ArrayDimentions is always a `list` of `int`s because OPC UA has support for multidimentional arrays. This is specified in the OPC UA standard at the end of [this](https://reference.opcfoundation.org/Core/Part6/v105/docs/5.2.2.16) page. But the return type in asyncua was is specified as `int`. This can generate false type-checker errors. Fix is to just change signature. The fucntion already correctly returns a list of integers.

Let me know if you want me to add a changelog entry also 😊


oroulet:
thanks, unfortunately we still have many of these wrong typing things..

Issue #1686, opened 2024-07-24
Received DataChangeNotification without Value has changed


Scheibengraber:
**Describe the bug** <br />
We receive data change notificications from a Siemens S7 1517 (Firmware 2.9.4) OPC UA server with asyncua version 1.1.5.

Console output:
```
C:\>python scratchfile_opcua_client.py
Requested session timeout to be 3600000ms, got 30000ms instead
Revised values returned differ from subscription values: CreateSubscriptionResult(SubscriptionId=3410194732, RevisedPublishingInterval=200.0, RevisedLifetimeCount=1000, RevisedMaxKeepAliveCount=150)
datachange_notification for node ns=5;i=6010: DataValue(Value=Variant(Value=False, VariantType=<VariantType.Boolean: 1>, Dimensions=None, is_array=False), StatusCode_=StatusCode(value=0), SourceTimestamp=datetime.datetime(2024, 7, 24, 15, 44, 11, 97178, tzinfo=datetime.timezone.utc), ServerTimestamp=datetime.datetime(2024, 7, 24, 15, 44, 11, 97178, tzinfo=datetime.timezone.utc), SourcePicoseconds=None, ServerPicoseconds=None)
datachange_notification for node ns=5;i=6010: DataValue(Value=Variant(Value=False, VariantType=<VariantType.Boolean: 1>, Dimensions=None, is_array=False), StatusCode_=StatusCode(value=0), SourceTimestamp=datetime.datetime(2024, 7, 24, 15, 49, 21, 96451, tzinfo=datetime.timezone.utc), ServerTimestamp=datetime.datetime(2024, 7, 24, 15, 49, 21, 96451, tzinfo=datetime.timezone.utc), SourcePicoseconds=None, ServerPicoseconds=None)
```

**To Reproduce**<br />
You can use the following script to reproduce the issue. Please exchange the variables host, port and node id "ns=5;i=6010" with your node id in your OPCUA server. 

```python
"""OPCUA client"""
import asyncio
import logging

from typing import List

from asyncua import Client, Node, ua
from asyncua.ua.uatypes import StatusCode

NodeList = List[str]

LOGGER = logging.getLogger()


class SubscriptionHandler:
    """The SubscriptionHandler is used to handle the data that is received for the subscription."""

    async def datachange_notification(self, node: Node, value, data):
        """Callback for asyncua Subscription. This method will be called when the Client received a data change message from the Server."""
[...]

PR #1696, opened 2024-08-13
set interval to 50 as default

AndreasHeine:
#1686


oroulet:
why? I cannot get the reasonning from the thread

AndreasHeine:
A Client shall define a sampling interval of 0 if it subscribes for Events.

from: https://reference.opcfoundation.org/Core/Part4/v104/docs/5.12.1.2

PR #1698, opened 2024-08-22
Prevent `asyncio.wait_for` swallowing task cancellation

aronhoff:
I've been running up against an issue in a project I'm working on, where I cannot cancel an asyncio task while it is waiting on an asyncua call.  The `CancelledError` simply gets swallowed and the task continues as if nothing happened

I traced it down. It turns out, there is a bug in the implementation of `asyncio.wait_for`, making it swallow the cancellation if the task was cancelled from the outside. See references below.

The bug is only fixed in Python 3.12, existing in one way or another since 3.7. The library wait_for2 provides a fixed implementation for the affected versions. 

This PR adds wait_for2 as a dependency replaces calls to `asyncio.wait_for` with `wait_for2.wait_for` pre-3.12.

https://github.com/python/cpython/issues/87555
https://github.com/python/cpython/issues/86296
https://github.com/python/cpython/issues/107547


aronhoff:
Note that in 3.12, `asyncio.wait_for` seems to be using an alternative implementation based on `asyncio.timeout`: https://github.com/python/cpython/issues/96764.

`asyncio.timeout` was only introduced in 3.11. Backport [async-timeout](https://pypi.org/project/async-timeout/) exists for earlier versions, but also seems to have unresolved issues regarding swallowed cancellations:
https://github.com/aio-libs/async-timeout/issues/229
https://github.com/aio-libs/async-timeout/pull/235

In my own project, setting `asyncio.wait_for = wait_for2.wait_for` at the start of the script, running in 3.11, has fixed the cancellation issues.

oroulet:
Did you really manage to reliably stop a task using CancelledError? My experience over several years is that they are always swallowed somehwere, if not a local error in code base then a bug in Python itself as you found out ;-)

PR #1714, opened 2024-10-02
add type hints around crypto module

jbeckerwsi:


asyncua/client/ua_client.py
```diff
@@ -88,7 +88,7 @@ def _process_received_data(self, data: bytes) -> None:
                     return
                 msg = self._connection.receive_from_header_and_body(header, buf)
                 self._process_received_message(msg)
-                if header.MessageType == ua.MessageType.SecureOpen:
+                if header.MessageType == ua.MessageType.SecureOpen and isinstance(msg,ua.Message):
```
oroulet, line 91: that check for isinstance cannot be correct. if that is because of mypy then annotate the "recevie_from_header_xx" over

asyncua/common/connection.py
```diff
@@ -51,7 +56,7 @@ def is_chunk_count_within_limit(self, sz: int) -> bool:
             _logger.error("Number of message chunks: %s is > configured max chunk count: %s", sz, self.max_chunk_count)
         return within_limit
 
-    def create_acknowledge_and_set_limits(self, msg: ua.Hello) -> ua.Acknowledge:
+    def create_acknowledge_and_set_limits(self, msg: "ua.Hello") -> "ua.Acknowledge":
```
oroulet, line 59: that too cannot be correct.  should be possible to use `from __future__ import annotations` stuff to avoid adding "

oroulet:
@jbeckerwsi conflicts and some comments

PR #1715, opened 2024-10-03
implement ua.Boolean, such that it can be instantiated by ua.Boolean(val)

jbeckerwsi:


oroulet:
@jbeckerwsi for some reason a test if failing

oroulet:
and now you got conflict of course...

PR #1718, opened 2024-10-07
Fix SimpleRoleRuleset

cziebuhr:
I had issues using a Siemens S7 PLC as OPC UA client reading values out of an asyncua server. It turned out that it used RegisterNodes, which is currently only allowed with write permission.

According to https://reference.opcfoundation.org/Core/Part4/v105/docs/5.8.5, the RegisterNodes Service can be used by Clients to register the Nodes that they know they will access repeatedly (e.g. Write, Call).

The service itself neither reads or writes data, it's just an optional optimization which is implemented as noop by asyncua.

### DIFF
--- a/asyncua/crypto/permission_rules.py
+++ b/asyncua/crypto/permission_rules.py
@@ -1,8 +1,7 @@
 from asyncua import ua
 from asyncua.server.users import UserRole
 
-WRITE_TYPES = [
-    ua.ObjectIds.WriteRequest_Encoding_DefaultBinary,
+ADMIN_TYPES = [
     ua.ObjectIds.RegisterServerRequest_Encoding_DefaultBinary,
     ua.ObjectIds.RegisterServer2Request_Encoding_DefaultBinary,
     ua.ObjectIds.AddNodesRequest_Encoding_DefaultBinary,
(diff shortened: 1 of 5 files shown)


cziebuhr:
I found another related issue:
When using SecurityPolicy None, SimpleRoleRuleset has not been used, allowing all requests even for anonymous users. Unfortunately, SecurityPolicy None (class SecurityPolicy) is implemented in ua/uaprotocol_hand.py, all other polices are implemented in crypto/security_policies.py. I now moved the usage of SimpleRoleRuleset to server/server.py, to only have it in one place. Also adding it to ua/uaprotocol_hand.py would have led to a circular import. SecurityPolicy() is also used both in server and client code, and shouldn't pull in server code via UserRole if used from client code. Personally, I would have also moved classes CryptographyNone, SecurityPolicy, and SecurityPolicyFactory to crypto/security_policies.py (not addressed in this PR), because they are internal classes and not defined by OPC UA spec.

As a consequence, I also allowed writing values for regular users:
- It was possible before SimpleRoleRuleset was implemented
- It was possible when using SecurityPolicy None
- It is still not possible for regular users if writable flag is not set
- It would otherwise be only possible for admins, for which writable flag isn't needed anyway
- Tests have been failing after the fix above

PR #1720, opened 2024-10-14
Add policy URI to each endpoint token's SecurityPolicyUri

haotruongnhat:
The opc-ua client backend open62541 scans for endpoint tokens' security URI on connection. This PR to resolve the case that open62541 to skip token which is missing that URI.

### DIFF
--- a/asyncua/server/server.py
+++ b/asyncua/server/server.py
@@ -495,6 +495,7 @@ class Server:
             idtoken = ua.UserTokenPolicy()
             idtoken.PolicyId = "anonymous"
             idtoken.TokenType = ua.UserTokenType.Anonymous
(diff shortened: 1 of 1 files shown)


cziebuhr:
According to https://reference.opcfoundation.org/Core/Part4/v105/docs/7.42, the security policy for the SecureChannel is used if this value is null or empty, so this should not have been an issue here. But I will report other findings in that area in the next days.

haotruongnhat:
@cziebuhr The UAExpert software uses the official OPC-UA backend, which does follow the document stated above. Therefore, the connection works fine with this tool.

However, in the open-sourced `open62541` backend, it checks only the securityPolicyUri in each available token, and skips the token if it is null or empty. It's quite tricky.

Issue #1722, opened 2024-10-15
generate_private_key produces key that can't be loaded with load_private_key


rth:
I run into a case where, a private key generated with `generate_private_key` cannot be loaded with `load_private_key`.

The minimal example is,
```py
from asyncua.crypto.uacrypto import load_private_key
from asyncua.crypto.cert_gen import generate_private_key, dump_private_key_as_pem
from pathlib import Path


async def main():
    key_file = Path("/tmp/cert.key")
    key = generate_private_key()
    key_file.write_bytes(dump_private_key_as_pem(key))

    key2 = await load_private_key(key_file)    
	

if __name__ == '__main__':
    import asyncio
    asyncio.run(main())
```

which produces the exception below,

```py
Traceback (most recent call last):
  File "check_private_key.py", line 16, in <module>
    asyncio.run(main())
  File "lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "check_private_key.py", line 11, in main
    key2 = await load_private_key(key_file)    
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/asyncua/crypto/uacrypto.py", line 80, in load_private_key
    return serialization.load_der_private_key(content, password=password, backend=default_backend())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 494, in _handle_key_loading_error
    raise ValueError(
[...]

PR #1723, opened 2024-10-15
TST Add test to generate and load a private key

rth:
The minimal example from https://github.com/FreeOpcUa/opcua-asyncio/issues/1722 as a unit test to see if fails or passes in CI.

Edit: Please let me know if you want such test, I don't have the impression that this combination is currently explicitly tested. Otherwise I can close this MR.


oroulet:
Tests are good

PR #1724, opened 2024-10-15
Add pre-commit-config and improve linting configuration

rth:
I made a PR recently and it was a bit frustrating due to lint errors. So I had to fix both mypy and ruff each time just to get some feed back from tests in CI.

This improves a bit the linting using experience with [pre-commit](https://pre-commit.com/). 
You can do,
```
pip install pre-commit
pre-commit install
```
and it would run linters (mypy, ruff)  locally when creating a new commit (on changed files). It's also possible to run linters for all files with `pre-commit run -a`.

This ensures that all the lint dependencies use the same version locally and in CI. Also it would display all lint errors in CI at once (rather than displaying ruff errors, having to fix them, only to see the mypy errors as currently)

I also moved linting to a separate job in CI, so it doesn't block tests if it fails. As there is no need to run them for different python versions.


asyncua/common/xmlimporter.py
```diff
@@ -482,7 +482,7 @@ def _set_attr(self, atttype, fargs, attname: str, val):
                     self._set_attr(sub_atttype, subargs, attname2, v2)
             if "Encoding" in subargs:
                 del subargs["Encoding"]
-            fargs[attname] = atttype(**subargs)
+            fargs[attname] = atttype(**subargs)  # type: ignore[operator]
```
rth, line 485: Otherwise mypy fails here with,
```
asyncua/common/xmlimporter.py:485: error: "DataclassInstance" not callable  [operator]
```
not really sure why

oroulet:
let's try

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - id: check-yaml
+        args: [--unsafe]
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: 'v0.6.9'
+    hooks:
+      - id: ruff
+        args: ['--fix']
```
oroulet, line 15: oops I merged it but that line looks very strange... why fix? there is not reason to run fix for linting, isnt'it?

rth: No that like is correct. That way you can use the same config locally and in CI.  Locally you run with a fix and it fixes. And it CI it will still exit with the exit status > 0, so the CI would fail and you would see what it would have fixed.

rth: If there are any difficulties with this don't hesitate to mention me.

oroulet: then you hav ruff check --diff, which is meant exactly for this, isnt'it?

rth: Well it's how they say it could be used with pre-commit https://github.com/astral-sh/ruff-pre-commit?tab=readme-ov-file#using-ruff-with-pre-commit

Without it, it means you have to manually fix those linting errors (or manually run `ruff check --fix`). Up to you.

About `ruff check --diff` not sure, I just copy past whatever the ruff team says the `.pre-commit-config.yaml` for ruff should be :)

PR #1726, opened 2024-10-16
move to pyproject.toml

oroulet:


pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+dependencies = [
+    "aiofiles",
+    "aiosqlite",
+    "cryptography>42.0.0",
+    "pyOpenSSL>23.2.0",
+    "python-dateutil",
+    "pytz",
+    "sortedcontainers",
+    "typing-extensions",
+    "wait_for2==0.3.2",
```
rth, line 35: maybe
```suggestion
    "wait_for2==0.3.2; python_version < "3.12",
```
since for Python 3.12+ it will never be imported https://github.com/FreeOpcUa/opcua-asyncio/blob/add6eddaab8ed90ffbfaf128c2456bda302ed86e/asyncua/common/utils.py#L153

cf https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers

pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+uahistoryread = "asyncua.tools:uahistoryread"
+uals = "asyncua.tools:uals"
+uaread = "asyncua.tools:uaread"
+uaserver = "asyncua.tools:uaserver"
+uasubscribe = "asyncua.tools:uasubscribe"
+uawrite = "asyncua.tools:uawrite"
+
+[project.urls]
+Homepage = "http://freeopcua.github.io/"
+
```
rth, line 67: Could be nice to add,
```
Repository = "https://github.com/FreeOpcUa/opcua-asyncio"
```
So in particular, it would display a link to the github repo on PyPI.

image

pyproject.toml
oroulet: thanks

pyproject.toml
oroulet: done

oroulet:
looks ok @rth ? I am thinking about merging that asap

rth:
The resulting `pyproject.toml` LGTM. It's nice to be able to move most config there. Thanks @oroulet !

rth:
@oroulet  I can make a PR to move [setup.cfg](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/setup.cfg), pytest.ini and [ruff.toml](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/ruff.toml) also inside pyproject.toml, unless you were already working on doing that?

oroulet:
Feel free to do it. My main goal was to start using uv and doing automated formatting.

PR #1734, opened 2024-10-25
Move ruff.toml, pytest.ini and setup.cfg to pyproject.toml

rth:
Follow up on https://github.com/FreeOpcUa/opcua-asyncio/pull/1726 cc @oroulet


setup.cfg
```diff
@@ -1,21 +0,0 @@
-[flake8]
-max-line-length = 160
-exclude = uaproto*_auto_add.py, *_auto.py, standard_address_space_services.py
-ignore = E501, W503, W504
-per-file-ignores =
-    __init__.py: F401, F403
-    examples/*: F841, E402
-[pycodestyle]
-max-line-length = 160
```
rth, line 9: removed the flake8 and pycodestyle config as it's replaced by ruff format.

pyproject.toml
```diff
@@ -72,6 +72,50 @@ Repository = "https://github.com/FreeOpcUa/opcua-asyncio"
...
 include = ["/asyncua"]
 
+[tool.pytest]
+log_cli = false
+log_level = "INFO"
+testpaths  = ["tests"]
+asyncio_mode = "auto"
 
 [tool.ruff]
-line-length = 120
```
rth, line 77: Note that this line was previously ignored since `ruff.toml` file existed.

If a line length of 120 is set currently (or even 160) there is a bunch of errors. So probably if this is to be fixed, might be easier to start using `ruff format`...

pyproject.toml
```diff
@@ -72,6 +72,50 @@ Repository = "https://github.com/FreeOpcUa/opcua-asyncio"
...
+line-length = 999
+extend-exclude = [
+    "docs",
+    "nodeset",
+    "schemas",
+    "*_auto.py",
+    "event_objects.py",
+    "standard_address_space_services.py"
+]
+target-version = "py37"
```
oroulet, line 91: should be 3.9 now, evertyhing else is not supported by Python

pyproject.toml
oroulet: very strnage I did a reformating of many things in a recent commit. 
but ok I can change that later

oroulet:
merge anyway, we can fix the other things later

PR #1735, opened 2024-10-26
set line length to 120

oroulet:


.github/workflows/python-package.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           uv tool install pre-commit
       - name: Lint with ruff
         run: |
+          uvx ruff format --diff
```
rth, line 34: It might be better to include formatting in pre-commit, so it would also run it locally on commit which is more convenient see https://github.com/astral-sh/ruff-pre-commit?tab=readme-ov-file#using-ruff-with-pre-commit

oroulet: I have 0 ideas how to add that to pre-commit. So you can do it.
But I feel it is a bit strange to use pre-commit to do linting. Isnt't much clearer to have the check we do in plain test here?

Issue #1594, opened 2024-02-27
Connecting to a real PLC: UASocketProtocol:ServiceFault, BadSessionNotActivated


hirsche:
**Describe the bug** <br />

I am connecting to an OPC UA server of a B&R PLC but receiving the error: UASocketProtocol:ServiceFault (BadSessionNotActivated ...). The same code works fine when connecting to an opcua-asyncio server and not to the B&R PLC. However, the implementation of the B&R PLC seem also to be correct, because when connecting with UA Expert everything works as expected and also those PLCs are professional grade products which are up-to-date regarding the software stack.

Interestingly, when the PLC is configured to allow security None, also the opcua-asyncio client framework can access the real B&R PLC without any issues.

Any help would be really appreciated, although I know it might be difficult to reproduce without an actual B&R PLC. 

**To Reproduce**<br />
Code to reproduce it is more or less the same like in [client-with-encryption.py](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/examples/client-with-encryption.py) in the examples folder. Nevertheless, here is the code:
```python
        client = Client(url=url)
        client.set_user('***')
        client.set_password('***')

        host_name = socket.gethostname()
        client_app_uri = f"urn:{host_name}:client"
        certificate: Path = Path(Path(__file__).parent.parent / "./key/cert.pem")
        private_key: Path = Path(Path(__file__).parent.parent / "./key/key.pem")

        client.application_uri = client_app_uri

        await setup_self_signed_certificate(private_key,
                                            certificate,
                                            client_app_uri,
                                            host_name,
                                            [ExtendedKeyUsageOID.CLIENT_AUTH],
                                            {
                                                'countryName': '***',
                                                'stateOrProvinceName': '***',
                                                'localityName': '***',
                                                'organizationName': '***',
                                                'commonName': '***'
                                            })

        await client.set_security(
            SecurityPolicyBasic256Sha256,
            private_key=str(private_key),
            certificate=str(certificate))
```

**Expected behavior**<br />
A secure connection

**Exception Stack**<br />
```
INFO:asyncua.client.ua_client.UaClient:activate_session
[...]

PR #1769, opened 2025-01-08
Fixes for using activate_session to change users.

mver-al:
- The updated ServerNonce was not saved after activate_session, which means that subsequent activate_sessions would fail with BadIdentityTokenInvalid.
- The self._password attribute of Client was checked in _add_user_auth function, instead of its own password argument.

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -143,12 +143,12 @@ class Client:
         """
         self._username = username
 
-    def set_password(self, pwd: str) -> None:
+    def set_password(self, pwd: str | None) -> None:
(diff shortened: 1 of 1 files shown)


mver-al:
Only thing I'm not fully sure about is whether the `_username` and `_password` actually have a purpose currently as member variables... They also aren'treset when logging in anonymously in the current situation...

mver-al:
I have a feeling #1594 could be related to this as well: the challenge for certificate_auth is also based on the server_nonce.

oroulet:
thanks

PR #1777, opened 2025-01-21
Update connection limit logging to check for changes

Eitol:
Added conditionals to log limit updates only when changes are detected, reducing unnecessary log noise. This applies to both server and client limit updates in the connection logic.

On the server these fake limit update logs cause noise and are mostly false positives.

Example:
image


oroulet:
Sorry I oversaw that notification

Issue #654, opened 2021-08-30
browse_nodes might fail with "BadNodeNotInView" due to ViewDescription.Timestamp being set 


educatedguessing:
I just encountered an opc ua server refusing my browse / browse_nodes attempts, raising a ""BadNodeNotInView"" exception.

After some searching I found what I believe to be the equivalent issue [here](https://github.com/gopcua/opcua/issues/342).

Digging deeper, I noticed an explicit handling of the View.timestamp in node.get_references. Possibly dealing with the same issue

https://github.com/FreeOpcUa/opcua-asyncio/blob/ecc3b1adbcf9366ae0934f5e3485bf8c294d44fc/asyncua/common/node.py#L395-L399

Applying this also for browse_nodes fixed the issue and caused no problems with any other server I'm using.

During this process I also read [here](https://reference.opcfoundation.org/v104/Core/docs/Part4/7.39/) that the current view should be returned if "timestamp is null and viewVersion is 0". 

I'm wondering if the defaults for ViewDescription.Timestamp might be better set using 0/win_epoch than utcnow.

https://github.com/FreeOpcUa/opcua-asyncio/blob/ecc3b1adbcf9366ae0934f5e3485bf8c294d44fc/asyncua/ua/uaprotocol_auto.py#L4947-L4951

(There is a good chance I'm not aware of all the implications.)

Anyway, it took me some time to resolve my issue and I wanted to at least discuss it openly for others to find.

Thank you for the great library
Ed

antoine-galataud:
I faced the same strange behavior while comparing UAExpert and opcua-asyncio requests against a UAGateway server using wireshark. The first was succeeding in browse requests, the second one failing with `BadViewTimestampInvalid` response code.

Replacing: 

```python
Timestamp: UtcTime = field(default_factory=datetime.utcnow)
```

by 

```python
Timestamp: UtcTime = field(default_factory=lambda: datetime(1601, 1, 1, tzinfo=timezone.utc))
```

fixes the issue.

cacamille3:
We faced the same issue. 
In our case, the OPC UA Server was returning BadViewTimestampInvalid: The view timestamp is not available or not supported (OPC 10000-4 - UA Specification Part 4 - Services 1.05.00 - Table 182 - Common Service Results Code )
We didn't fix it in the library, but instead, we had the ViewDescription.Timestamp overwriten in the caller with the same DateTime 1601-01-01. 

[...]

PR #1779, opened 2025-01-24
ViewDescription default Timestamp value should be OPC UA TimeBase

cacamille3:
OPC UA TimeBase is 1601-01-01 UTC
Otherwise, some OPC UA Server not supporting this Browse optional feature might return BadViewTimestampInvalid when the Timestamp is not set to null

#654

### DIFF
--- a/asyncua/ua/uaprotocol_auto.py
+++ b/asyncua/ua/uaprotocol_auto.py
@@ -8,7 +8,7 @@ from enum import IntEnum, IntFlag
 from typing import List, Optional, Type
 from dataclasses import dataclass, field
 
-from asyncua.ua.uatypes import FROZEN
+from asyncua.ua.uatypes import FROZEN, FILETIME_EPOCH_AS_UTC_DATETIME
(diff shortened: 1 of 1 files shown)


cziebuhr:
You can't just do changes in uaprotocol_auto.py, as it gets autogenerated.

https://reference.opcfoundation.org/Core/Part6/v105/docs/5.1 has a table with default values per built-in type. I think it makes sense to use all those defaults in all places where default values gets created:
- common/structures104.py:get_default_value
  - Boolean: "False" (instead of "True")
  - DateTime: "FILETIME_EPOCH_AS_UTC_DATETIME"
- ua/uatypes.py:get_default_value
  - ByteString: null (instead of b"")
  - DateTime: FILETIME_EPOCH_AS_UTC_DATETIME
 
(List may not be complete)

oroulet:
but you can simply move that code uaprotocol_hand.py or similar that overrides autogenerate classes. We do that already for a few classes

PR #1782, opened 2025-01-30
Add typing metadata and include `py.typed` in sdist

Eitol:
I submitted this pull request to enhance the opcua-asyncio library's compatibility with static type checkers like mypy. While running mypy in my project, I encountered the following error:
```text
Skipping analyzing "asyncua.ua": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

This issue arises because mypy requires either inline type annotations or the presence of a py.typed file in the package to recognize it as typed. By adding the py.typed file and including it in the source distribution (sdist), we ensure that mypy and other type checkers can properly analyze the asyncua module, thereby improving the developer experience and code reliability.


oroulet:
the main issue is that typing is far from perfect. but we can try again to fix it...

oroulet:
done in another MR

PR #1786, opened 2025-02-06
fix: NodeId change for class DataValue

torarvid:
According to ObjectIds, [DataValue = 23](https://github.com/FreeOpcUa/opcua-asyncio/blob/447bdf7596ba07c0aac239919e28471b0b3efd45/asyncua/ua/object_ids.py?plain=1#L28), and not 25 as it was before this commit. [25 is DiagnosticInfo](https://github.com/FreeOpcUa/opcua-asyncio/blob/447bdf7596ba07c0aac239919e28471b0b3efd45/asyncua/ua/object_ids.py?plain=1#L30), which can also be seen on line 1137 in the uatypes.py file.

(this is my first PR in this repo, and I found it while "just browsing around", so feel free to tell me if I'm completely off here 😉)

Edit: Now using `ObjectIds.DataValue` instead of `23` to make it more obvious.

### DIFF
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -1087,7 +1087,7 @@ class DataValue:
     :vartype ServerPicoseconds: int
     """
 
-    data_type = NodeId(25)
+    data_type = NodeId(23)
 
     Encoding: Byte = field(default=0, repr=False, init=False, compare=False)
     Value: Optional[Variant] = None
--- a/asyncua/ua/uatypes.py
+++ b/asyncua/ua/uatypes.py
@@ -15,6 +15,8 @@ from datetime import datetime, timedelta, timezone
 from enum import IntEnum
 from typing import Any, Generic, List, Optional, Union
 
+from asyncua.ua.object_ids import ObjectIds
+
 # hack to support python < 3.8
 if sys.version_info.minor < 10:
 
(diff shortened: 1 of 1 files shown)


oroulet:
how can such a big bug been around for so long. I am not really getting it. is that attribute not used?

torarvid:
Btw, I noticed this "just by luck", since I thought it was surprising that the value `25` was used in both DataValue and DiagnosticInfo (they are close together in the file).

But it also got me wondering: why are plain numbers used here? Would it not be easier to spot/prevent bugs like these if the ObjectIds were used? So:

```python
    data_type = NodeId(Int32(ObjectIds.DiagnosticInfo))
```

instead of
```python
    data_type = NodeId(25)
```

oroulet:
historical reason I suppose. Also you probably do not need to cast to Int32. it is done at lower level I suppose.
Crazy that it works with wrong data type for DataValue.... but OK DataValue are probably mostly sent at places where the data type is hardcoded

oroulet:
ok then add it. The warning msut come from mypy or the facebook alternative

torarvid:
@oroulet Oh, did you mean I should change the PR so that instead of using `23` directly, I should just use the ObjectId? Sure, I can do that 👍🏼 

(And I think actually it's my Language Server — pyright — that's giving the Int32 error, not mypy)

Issue #1319, opened 2023-05-31
Unknown datatype when calling load_data_type_definitions


tomj-99:
When I connect to an OPC UA server as a client and call `await client.load_data_type_definitions()`, I get this error:

`error:  Unknown datatype for field: StructureField(Name='MyCustomTypeArrayField', Description=LocalizedText(Locale=None, Text=None), DataType=NodeId(Identifier=3008, NamespaceIndex=3, NodeIdType=<NodeIdType.FourByte: 1>), ValueRank=1, ArrayDimensions=[0], MaxStringLength=0, IsOptional=False) in structure:MyCustomStructureType, please report`

When calling `await client.load_type_definitions() ` instead, it works fine but gives a warning `Deprecated since spec 1.04, call load_data_type_definitions`. 

The field 'MyCustomTypeArrayField' is an array of a custom data type (ns=3 i=3008, with only basic opc ua data types as fields) with an array size of 0 that is set dynamically at runtime.

![image](https://github.com/FreeOpcUa/opcua-asyncio/assets/132424909/f7066f77-439c-4875-a4c3-86499a8fa1e5)

The server that I connect to uses the .NET SDK v3.3.0 from UnifiedAutomation which should actually support spec 1.05.

versions:
Python 3.11.0
ayncua 1.0.2

tomj-99:
@schroeder- yes, that exists. I just had a closer look at the log and found this:

`Structure type StructureDefinition(DefaultEncodingId=NodeId(Identifier=5018, NamespaceIndex=3, NodeIdType=<NodeIdType.FourByte: 1>), BaseDataType=NodeId(Identifier=22, NamespaceIndex=0, NodeIdType=<NodeIdType.TwoByte: 0>), StructureType_=<StructureType.StructureWithSubtypedValues: 3>, Fields=[...]) not implemented` 

I shortened it a little bit, the important part seems to be `StructureType_=<StructureType.StructureWithSubtypedValues: 3>`, I just went through the code of `load_data_type_definitions` and found in `structures104.py` in the function `make_structure_code` these lines:

```
if sdef.StructureType not in (ua.StructureType.Structure, ua.StructureType.StructureWithOptionalFields, ua.StructureType.Union):
        raise NotImplementedError(f"Only StructureType implemented, not {ua.StructureType(sdef.StructureType).name} for node {struct_name} with DataTypdeDefinition {sdef}")
```

so `StructureWithSubtypedValues `does not seem to be supported

[...]

PR #1790, opened 2025-03-01
Add support for reading structure with subtyped values

olofos:
This PR adds support for loading data types definitions for types that have StructureType set to StructureWithSubtypedValues, and to read the corresponding values. It does this by simply wrapping the types that should be allowed to have subtypes in a typing.Annotated, which ends up giving the right result because of how ua.uatypes.type_allow_subclass is defined. I'm not sure if this is the most proper way of doing this

As far as I can tell this is not enough to *write* such a value, or at least I haven't manage to figure out how to set that up. Since I wasn't able to write any values I also weren't able to add any tests that check reading structures with subtypes. I have however successfully tested loading data types and reading values from an external server

There is also the UnionWithSubtypedValues structure type which I'm ignoring here because I haven't seen it used anywhere

This patch requires python 3.9 but that seems to already be the minimum according to pyproject.toml even though the readme says 3.8

Also see issue #1319

### DIFF
--- a/asyncua/common/structures104.py
+++ b/asyncua/common/structures104.py
@@ -179,6 +179,7 @@ def make_structure_code(data_type, struct_name, sdef, log_error=True):
         ua.StructureType.Structure,
         ua.StructureType.StructureWithOptionalFields,
         ua.StructureType.Union,
+        ua.StructureType.StructureWithSubtypedValues,
     ):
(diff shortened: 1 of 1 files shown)


oroulet:
Ideally that feature should have a test but since it does not seem to break anything, I will merge

PR #1794, opened 2025-03-07
Fix Nodeclass in ReferenceDescriptions of UA-Nodeset

### LINKED ISSUE #1529: Browsing for HasTypeDefinition references is broken
Browsing a server built with opcua-asyncio for HasTypeDefinition references is broken.

When specifying a NodeClassMask of only ObjectType or VariableType, no results are returned.

If the NodeClassMask is set to 0 (all NodeClasses should be returned), then 1 result is returned, however it incorrectly identifies the NodeClass as `NodeClass.DataType`. (I suspect this is the cause of the 0 length result when a mask is supplied).

This code written against Eclipse Milo dev/1.0 branch reproduces, though you should be able to reproduce from any OPC UA SDK: https://gist.github.com/kevinherron/f85ce3b0eebe84008e59a5192f6b404d

Results:
```
Browsing with NodeClass mask: 24
  No references found
Browsing with NodeClass mask: 0
  Reference[0]:
    NodeId: ExpandedNodeId{ns=0, id=2138, serverIndex=0}
    BrowseName: QualifiedName{name=ServerStatusType, namespaceIndex=0}
    DisplayName: LocalizedText{text=ServerStatusType, locale=null}
    NodeClass: DataType
    TypeDefinition: ExpandedNodeId{ns=0, id=0, serverIndex=0}
```

Wireshark capture: [browse_has_type_definition.zip](https://github.com/FreeOpcUa/opcua-asyncio/files/13696531/browse_has_type_definition.zip)

cziebuhr:
Fixes #1529.

Add references after all nodes has been processed, when we know the respective NodeClass of the TargetNode. Reused schema version 1.05.01, to not introduce additional changes to the address space.

I also fixed the generation of binary_address_space.pickle, which probably wasn't up-to-date before, unless the generator has been run twice. Not sure if this file is needed at all, at least it's not used by asyncua itself.

### DIFF
--- a/schemas/generate_address_space.py
+++ b/schemas/generate_address_space.py
@@ -98,6 +98,7 @@ class CodeGenerator:
         self.output_file = None
         self.part = self.input_path.parts[-1].split(".")[-2]
         self.parser = None
+        self.nodes = {}
 
     async def run(self):
         sys.stderr.write(f"Generating Python code {self.output_path} for XML file {self.input_path}\n")
@@ -122,6 +123,8 @@ class CodeGenerator:
                 self.make_method_code(node)
             else:
                 sys.stderr.write(f"Not implemented node type: {node.nodetype}\n")
+        for obj in self.nodes.values():
+            self.make_refs_code(obj, "   ")
         self.output_file.close()
 
     def writecode(self, *args):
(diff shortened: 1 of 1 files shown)


oroulet:
thanks!

PR #1802, opened 2025-03-29
Aware time in the examples

54ny7:
datetime.datetime.utcnow() is deprecated.
So I use ".now(UTC)"" and "from datetime import datetime, UTC". And fix typos. Optimize imports


oroulet:
linting is failing

54ny7:
Passed.

Issue #621, opened 2021-07-15
Unicode Decode Error When Reading Node Value


sanssecours:
# Description

The library seems to have problem when reading string values that contain incorrect UFT-8 byte sequences. It would be nice, if there would be an option to replace these byte sequences with a replacement character such as `�`.

# Steps to Reproduce the Problem

1. Paste the following Python code

   ```py
   from asyncio import run
   from pathlib import Path
 
   from asyncua import Client
   from asyncua.ua import NodeId
 
 
   async def read_value():
       cert = Path(__file__).parent.joinpath("cert.der")
       key = Path(__file__).parent.joinpath("key.pem")
       uri = "opc.tcp://SuperCoolUsername:SuperSafePassword@localhost:4840"
       security = f"Basic256Sha256,SignAndEncrypt,{cert},{key}"
       client = Client(uri)
       await client.set_security_string(security)
 
       async with client as client:
           return (await client.read_values([
               client.get_node(NodeId("/Channel/ProgramInfo/actPartProgram", 2))
           ])).pop()
 
 
   if __name__ == '__main__':
       print(run(read_value()))
   ```

2. Replace the values `SuperCoolUsername` and `SuperSafePassword` with actual values that work for your OPC UA server
3. Save the text as `error.py`
4. Generate the key `key.pem` and certificate `cert.der` with a slightly modified version of [`generate_certificate.sh`](https://github.com/FreeOpcUa/opcua-asyncio/blob/master/examples/generate_certificate.sh) and put them into the same folder as the script.
5. Execute the Python script

   ```sh
   python error.py
   ```

# Expected Result

The script prints the value of the node `/Channel/ProgramInfo/actPartProgram`:

```

MCHOME
TLCH1H("M3016FR",-90) ;Werkzeugwechsel h�ndisch



TLCOSOF
MCMILLS1
G54
D1
G95 S3=294 M3=3 F=0.45  ;M3=8
;MCICOTON

G0 X1=160 Y1=0 C1=150
G0 Z1=-11
G1 X1=-20
G0 X1=0
G1 Y1=25
```

The value seems to contain an Umlaut character (`ä`) (in ISO-8859-1 encoding). The screenshot below shows the value in the Prosys OPC UA Browser:

<img width="1482" alt="Prosys" src="https://user-images.githubusercontent.com/691989/125792725-099c5269-61d1-42d1-8509-9c93ee8ae81b.png">

[...]

PR #1812, opened 2025-04-03
Fixes encoding related issues with mutated vowels - e.g. cannot retrieve information of converted nodeids based on umlauts

lkaupp:
Now, converted NodeIds can be used to request more information. E.g., in the XmlExporter or somewhere else. It is now possible to work with German umlauts / mutated vowels if present in the OPC UA model. A cover-up implementation of https://github.com/FreeOpcUa/opcua-asyncio/issues/621


oroulet:
linting failed...

PR #1819, opened 2025-04-11
Make base user manager abstract

pbouill:


.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
oroulet, line 1: you make .gitignore ignore that file, then you commit it?

oroulet:
Can you close one of your MR? why 2?

PR #1820, opened 2025-04-15
Make base address space abstract

pbouill:
adding abstract classes for user manager, address space in order to allow for better implementation of iserver instantiation (allow iserver to use custom address space class)


oroulet:
fine but I think these changes should be a few commits. one for each new abstract classes. Not 19 with some random rebases

oroulet:
to be honest I am afraid of all these changes. A lot of chnages that  makes code harder to follow and the use case is not really clear for me.
Also what are the impact on performamce on all the calls redirections?

PR #1832, opened 2025-04-28
Fix small bug if imported xml has also the stardard namespace included

lkaupp:
This minor adjustment resolves a bug that occurs when importing an XML file with the standard "http://opcfoundation.org/UA/" namespace. Without this fix, the subsequent code would incorrectly shift all namespaces one level deeper, resulting in the following changes:

Level 1 namespaces being treated as Level 0
Level 2 namespaces being treated as Level 1 and so on.

### DIFF
--- a/asyncua/common/xmlimporter.py
+++ b/asyncua/common/xmlimporter.py
@@ -54,6 +54,14 @@ class XmlImporter:
         if not present the namespace is registered.
         """
         xml_uris = self.parser.get_used_namespaces()
+
+        # Check if first namespace is the standard namespace and
+        # drop it, to prevent the following code from failing and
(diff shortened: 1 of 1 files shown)


oroulet:
small linting issue

lkaupp:
@oroulet could you guide me? idk why the linter is failing. Is there a linter, which i can install locally to prevent these base checks from triggering?

oroulet:
This is the formater. Try 'ruff format '

oroulet:
Maybe some documentation is missing...

lkaupp:
@oroulet reformated with ruff, hopefully everything is fine.

PR #1833, opened 2025-04-28
Enhancement: Chunked Build ETree Function

lkaupp:
To mitigate server overload issues caused by the original build_etree function, a new chunked version has been implemented. This updated function reconnects after processing a specified number of chunks, preventing random disconnects and ensuring successful extraction of large numbers of variables (tested with >30.000 variables).

Key Improvements:

1. Chunked Processing: The new function processes data in chunks, reconnecting after each chunk to prevent server overload.
2. Configurable Chunk Size: The chunk size can be defined, allowing for customization based on specific use cases.
3. Namespace Extraction: The updated function can now extract all namespaces, not just the ones in use, providing more comprehensive XML data.

### DIFF
--- a/asyncua/common/xmlexporter.py
+++ b/asyncua/common/xmlexporter.py
@@ -55,7 +55,7 @@ class XmlExporter:
         if self._export_values:
             self.logger.warning("Exporting values of variables is limited and can result in invalid xmls.")
 
-    async def build_etree(self, node_list):
+    async def build_etree(self, node_list, add_all_namespaces = False):
         """
         Create an XML etree object from a list of nodes;
         Namespaces used by nodes are always exported for consistency.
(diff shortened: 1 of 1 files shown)


lkaupp:
@oroulet whats wrong with this build? does the built reconnect violates the tests?

oroulet:
> @oroulet whats wrong with this build? does the built reconnect violates the tests?

I am not following the comment about reconnect but pypy3.10 has been randomly  failing some pipelines lately
and pypy 3.12 does not work at all. Someone should look at that at some point

lkaupp:
Well I added a reconnect to the novel introduced function of build_etree_chunked. I though this might happend here. However, thank you for clarifing things. Have a great day :)

PR #1834, opened 2025-04-28
Pull Request: Added Support for Custom String Encoding

lkaupp:
#### Description
This pull request introduces the ability to set a custom string encoding for communication with non-standard servers. This change addresses numerous issues related to vowel mutations and umlauts in the German-speaking area.

#### Changes
* Added an option to specify a custom string encoding, with UTF-8 as the default value.
* Updated the client and server to accept instructions for using alternative encodings for strings, including node IDs.
* Provided support for encodings like CP1252, which can be used as an alternative to UTF-8.
* Added fallback for readonly communication with surrogates

#### Benefits
* Resolves issues with vowel mutations and umlauts in the German-speaking area.
* Enhances compatibility with non-standard servers that use custom string encodings.
* Allows for more flexibility in communication between the client and server.

#### Example Use Case
To use a custom encoding like CP1252, simply instruct the client or server to use this encoding for strings, including node IDs. This can be particularly useful when communicating with servers that do not support UTF-8.

### DIFF
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -35,7 +35,7 @@ _logger = logging.getLogger(__name__)
 
 T = TypeVar("T")
 
-_string_encoding: str = contextvars.ContextVar("ua_string_encoding", default="utf-8")
+_string_encoding = contextvars.ContextVar("ua_string_encoding", default="utf-8")
 
 def get_string_encoding() -> str:
     return _string_encoding.get()


lkaupp:
forget to add the type serialiser .. update incoming

schroeder-:
From the specs only UTF-8 is allowed. [Spec](https://reference.opcfoundation.org/Core/Part6/v104/docs/5.2.2.4)
If we are the problem we should make sure we are always sending UTF-8.
If the problem is another server/server, tell them to comply to the spec.

lkaupp:
What should i say. I currently work with a Siemens OPC UA Server that runs a windows encoding. The encoding issues that pop up over the time show that german umlauts are a serious problem in encoding - but this should not be the case right? utf-8 has german umlauts / mutated vowel in it. The only way this problem could happen in the first place is that the server using different encoding. That was verified our server runs on cp1552 encoding. The server sends and expects strings encoded using this encoding. In this and the old repo multiple issues got raised due to the german umlaut encoding.

To address this, I'm trying to add the ability to change the encoding in the low-level API. Currently, I'm focusing on node ID parsing, as it's impossible to subscribe to a node that has been surrogate-encoded or replaced and saved to a file (as currently implemented). However, to achieve this, I've had to make more changes than I initially expected. I hope to push a fully functional version by the end of the week, which will use UTF-8 as the standard encoding but also allow for other encodings to be used if necessary.

One concern I have is whether the incorrect encoding is also affecting the implemented XML protocol, aside from the subscription issues / nodeid issues. From what I can tell, the binary protocol doesn't seem to be affected by this issue. I'd appreciate any feedback or guidance on how to proceed with this change.

asyncua/ua/ua_binary.py
```diff
@@ -71,19 +71,23 @@ def unpack(data):
         return data.read(length)
 
 
+class LowApiConfig:
```
oroulet, line 74: that name is so strange... What is supposed to mean?

lkaupp: this name is open to debate, feel free for another suggestion. This is the least effort, without intervene with your great design of the library, and I added the String_Encoding as first element. In the future, maybe there are other configuration parameters which should be altered specifically e.g. for custom objects that have a different array encoding / byte encoding. The config is there to change the parameters on the low-level communcation. Therefore, my suggestion of a "lowApiConfig" for this class.

lkaupp:
Any concerns left, that i can implement for you to merge this into the master? I am willing to assist at any point. I want to give some of our findings back to the community. This lib is great and I want to contribute my part to the success story.

asyncua/client/client.py
```diff
@@ -41,7 +42,7 @@ class Client:
     _password: Optional[str] = None
     strip_url_credentials: bool = True
 
-    def __init__(self, url: str, timeout: float = 4, watchdog_intervall: float = 1.0):
+    def __init__(self, url: str, timeout: float = 4, watchdog_intervall: float = 1.0, encoding="utf-8"):
```
oroulet, line 45: tihs will be a very rarely used option, leave it as property, no need to be part of constructor

asyncua/server/server.py
```diff
@@ -84,7 +85,7 @@ class Server:
         server listens on some internal IP.
     """
 
-    def __init__(self, iserver: InternalServer = None, user_manager=None):
+    def __init__(self, iserver: InternalServer = None, user_manager=None, encoding="utf-8"):
```
oroulet, line 88: same comment as for the client. 
Also do we want that in the server? Why making a server not supporting the standard?

asyncua/ua/ua_binary.py
oroulet: Maybe just remove that class and use directly the member String_Encoding...
also StringEncoding... mixing up camel case and underscore is not allowed by convention in Python

oroulet:
> Any concerns left, that i can implement for you to merge this into the master? I am willing to assist at any point. I want to give some of our findings back to the community. This lib is great and I want to contribute my part to the success story.

I wrote some comments. Fix them and we can merge that if that is needed in the real world

asyncua/server/server.py
lkaupp: super simple. If you have to mirror the servers behaviour to track the faults in implementations, you need the same encoding as the original server. Otherwise xmlexported values cannot be reimported and debuggt using the client. The information extracted cannot be automatically casted into the new encoding - not feasable with thousands of variables. Here the server needs to speak the same languqge as the client.

asyncua/client/client.py
lkaupp: how we should set the variable, using a subsequent function? Defining the encoding in the constructor as with any other client/server library would make more sense to me as a season programmer.

lkaupp:
> > Any concerns left, that i can implement for you to merge this into the master? I am willing to assist at any point. I want to give some of our findings back to the community. This lib is great and I want to contribute my part to the success story.
> 
> I wrote some comments. Fix them and we can merge that if that is needed in the real world

Done :). If it should be set in a subsequest function request, np. I could also add it as a kwargs argument.

asyncua/client/client.py
oroulet: not sure if I understand correctly but in python all attributes are public by default. and that class has already many many atrtibutes...

so in python you would do

```
c = Client()
c.encoding = "whatever" # for the very few special cases

```

I suppose in java/C++ you would do

```
c = Client()
c.set_encoding(whater)
```

and in rust

`let c = Client::new().encoding("whatewerr");`

asyncua/server/server.py
oroulet: can be valid for testing but definitely use property and nor constructor argument

asyncua/ua/ua_binary.py
oroulet: @lkaupp you still need to fix that. LowApiConfig really means nothing to my brain.
Remove that calss and keep the config as a property in that file

string_encoding: st = SSSSS

and follow PEP8 for the naming

oroulet:
> > > Any concerns left, that i can implement for you to merge this into the master? I am willing to assist at any point. I want to give some of our findings back to the community. This lib is great and I want to contribute my part to the success story.
> > 
> > 
> > I wrote some comments. Fix them and we can merge that if that is needed in the real world
> 
> Done :). If it should be set in a subsequest function request, np. I could also add it as a kwargs argument.

@lkaupp if that is done, it is not pushed...

Alexander-Zender:
@oroulet I made the requested changes. It would be amazing if you could check it out.

lkaupp:
@oroulet is it ok with the changes Alex made?

PR #1835, opened 2025-04-29
Feat abstract iserver

pbouill:
improve server typehinting, allow users to create their own address space classes (and user manager) and create/initialize the server object with these custom components

### DIFF
--- a/tests/test_connections.py
+++ b/tests/test_connections.py
@@ -115,7 +115,9 @@ async def test_session_watchdog():
     client.session_timeout = 1000  # 1 second
     await client.connect()
     client._closing = True  # Kill the keepalive tasks
-    await asyncio.sleep(2)  # Wait for the watchdog to terminate the session due to inactivity
+    await asyncio.sleep(
+        delay=client.session_timeout / 1000 + 1
+    )  # Wait for the watchdog to terminate the session due to inactivity
     with pytest.raises(BadSessionNotActivated):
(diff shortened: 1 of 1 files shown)


asyncua/server/base.py
```diff
@@ -0,0 +1,339 @@
...
+        self,
+        iserver: AbstractInternalServer,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        certificate: Optional[Certificate] = None,
+    ) -> Optional[User]:
+        pass
+
+
+class AbstractInternalServer(ABC):
```
oroulet, line 41: Internal Server is most of the implentation of the Ua server. How can it make sense t be able to abstract it?

asyncua/server/server.py
```diff
@@ -84,22 +87,52 @@ class Server:
...
+        iserver: Optional[AbstractInternalServer] = None,
+        aspace: Optional[AbstractAddressSpace] = None,
+        user_manager: Optional[AbstractUserManager] = None,
+        url: str = "opc.tcp://localhost:4840/freeopcua/server/",
+        uri: str = "urn:freeopcua:python:server",
+        product_uri: str = "urn:freeopcua.github.io:python:server",
+        name: str = "FreeOpcUa Python Server",
+        manufacturer_name: str = "FreeOpcUa",
+        socket_address: Optional[Tuple[str, int]] = None,
+        discovery_period: int = 60,
```
oroulet, line 101: why sudently expose some parameters as constructor args when they are anyway available on the object itself as property?

PR #1868, opened 2025-09-16
Handling ErrorMessage response by propagating the exception to futures

40csmith40:
### Issue
When a response message is returned as a `ua.ErrorMessage` type, callbacks cannot be utilized as there is no `request_id` to access the callback. The server may respond immediately with a failure which currently will log the error and disconnect the socket, but the request future hangs until the asyncio `TimeoutError` from the `wait_for` is raised. 

This is especially apparent when the `send_hello` fails (try putting a typo in your the protocol for the endpoint, but with a valid hostname and port, like "opc.XXX://<your endpoint>"). When this happens the server immediately responds with `BadTcpEndpointUrlInvalid` but the exception never makes it to the request and it eventually fails on a timeout which is misleading. 

### Solution
If a fatal server failure happens, the exact UA exception is raised for any pending futures. This causes them to finish immediately and properly raise the exception and message that caused the fatal error rather than always raising a timeout.

### DIFF
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -115,6 +115,7 @@ class UASocketProtocol(asyncio.Protocol):
                 data = bytes(buf)
             except ua.UaStatusCodeError as e:
                 self.logger.error("Got error status from server: {}".format(e))
+                self._fail_all_pending(e)
                 self.disconnect_socket()
                 return
             except Exception:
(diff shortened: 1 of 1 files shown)


oroulet:
I really think we should merge your code. It is a good solution to an issue that has been here for a long time. I am just afraid of side effects.

OK I think we merge. If that create issue it must be quite rare, we would need a non fatal error from server, a call excuting exactly during that time that should haev succeeded and is fatal to the application.

oroulet:
3.13 tests never pass , did they ever pass locally on your branch?

oroulet:
really good work

Issue #1843, opened 2025-06-12
Client certificate chain support


hrycko-mb:
Hi, I was looking for support of certificate chain for client side both for user authentication and secure channel, but could not find any direct mentions.

Is this feature implemented? If not, is it planned or at least known about? Are there any implementations or PoCs? 

Thank you in advance for the answers.

PR #1871, opened 2025-09-17
Client certificate chain support

still-wondering:
Following the issue #1843, this pull request adds support for OPC UA client to include the chain of intermediate certificates when sending leaf authentication/communication certificate. 

This, e.g., allows the client to connect to the server with the certificate issued by the intermediate certificate which is not available to the server.


oroulet:
honestly I do not know that stuff, so I will jsut trust you...

PR #1891, opened 2025-11-13
Split 'standard_address_space_services.py' into chunks

jprogin:
We want to use asyncua on RevPI devices. However, simply running `pip install asyncua` was causing a memory overflow and the device was needing a hardware reset. A turn around was to limit virtual memory usage (something like `ulimit -v 300000`) before running the pip installation, but later the simple `import asyncua` was causing the memory overflow again.

Investigations revealed that the origin of the issue was the huge size of the file `standard_address_space_services.py` file. It seems that Python on RevPI was consuming the whole memory available during the load of this file. 

Splitting this file into sub function chunks, each loaded separately and then called by a unique function, solves this issue. This behavior has been implemented into `schemas/generate_address_space.py` and the `update_ua_nodeset.py` has been called.


jprogin:
I suppose that the tests are not ok due to the version of the Node Set used during the generation...

Any help appreciated.

PR #1897, opened 2025-12-03
refactor!: custom-types are hidden under aliases on ua

sebastianpfischer:
In a nutshell, when a data-type is registered, it is registered on ua with a unique id _. The registered class has an __alias__ attribute too. Because, nodes are unique (at least per opcua namespeces), I can assign the right data-type to the right node.

I extended ua with 3 core methods that should be used to find any struct stored on ua:

get_custom_struct_via_nodeid() <- can be used like: ua.get_custom_struct_via_nodeid(await value_node.read_data_type())
get_custom_struct_with_perfect_match() <- that can be used get_custom_struct_with_perfect_match('MyStruct', ['field1', 'field2') that would return the structure that matches your name and contains the field names you need.
get_custom_struct_via_name() <- my least favorite, it returns the first struct that matches the given name.

### DIFF
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -623,7 +623,7 @@ def extensionobject_to_binary(obj):
         encoding = 0
         body = None
     else:
-        name = obj.__class__.__alias__ if hasattr(obj.__class__, '__alias__') else obj.__class__.__name__ 
+        name = obj.__class__.__alias__ if hasattr(obj.__class__, '__alias__') else obj.__class__.__name__
         type_id = ua.extension_object_typeids[name]
         encoding = 0x01
         body = struct_to_binary(obj)
--- a/asyncua/ua/ua_binary.py
+++ b/asyncua/ua/ua_binary.py
@@ -42,7 +42,17 @@ def get_safe_type_hints(cls, extra_globals=None):
     # Filter out properties from the class dict to avoid shadowing
     localns = {k: v for k, v in cls.__dict__.items() if not isinstance(v, property)}
 
-    return typing.get_type_hints(cls, globalns=globalns, localns=localns)
+    try:
+        resolved_fieldtypes = typing.get_type_hints(cls, globalns=globalns, localns=localns)
+    except AttributeError:
+        # Some recursion is happening and typing needs to be fixed.
+        # To do that, we will temporarily create the class as it is named (instead of the alias)
+        # Then delete it again
+        setattr(ua, cls.__name__, cls)
+        resolved_fieldtypes = get_type_hints(cls, {"ua": ua})
+        delattr(ua, cls.__name__)
+
+    return resolved_fieldtypes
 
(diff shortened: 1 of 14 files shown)


asyncua/ua/ua_binary.py
```diff
@@ -40,7 +40,17 @@ def get_safe_type_hints(cls, extra_globals=None):
...
-    return typing.get_type_hints(cls, globalns=globalns, localns=localns)
+    try:
+        resolved_fieldtypes = typing.get_type_hints(cls, globalns=globalns, localns=localns)
+    except AttributeError:
+        # Some recursion is happening and typing needs to be fixed.
+        # To do that, we will temporarily create the class as it is named (instead of the alias)
+        # Then delete it again
+        setattr(ua, cls.__name__, cls)
+        resolved_fieldtypes = get_type_hints(cls, {"ua": ua})
+        delattr(ua, cls.__name__)
```
oroulet: I think we had a try except before and I managed to remove it recently...
I do not get why this is necessary again.

that temporary hack is even more horrible...

sebastianpfischer:
@oroulet, maybe for the backward compatibility you can help me?
The problem is that the during node creation, the datatype is searched. If the datatype exist, the node is not created and the existing node is shared cross. By making the data-struct unique, I force the nodes to be created every time.

sebastianpfischer:
@oroulet, on my side, it looks like the linter is already fixed no? I just see pypy-3.11 failing.

oroulet:
arrg I had a second look and that is much more difficult than I though. We are using python type hints to decide what ua type a struct member is.  All typing is writting as ua.xxx. if you want to change that to refer to some ua data type or name, you need to change the entire encoding of ua structures in python. 
I do not understand yet how you managed to get your example works, it must work only for highest level because you still have code everywhere looking for class names in ua and setting it there as I pointed one case over.

sebastianpfischer:
> > > > @oroulet, your proposal is not working, even though I tried to use it as base... Meanwhile, I found a way for the backward compatibility.
> > > 
> > > 
> > > would be great to know why. fundamentally it should work. Current code is already hacky enough, I cannot merge anything...
> > 
> > 

> > I think it is because there is a connection done on node-level. You do not re-enter `_generate_object()` for the second object creation of the same name. That is what I meant in my message of the 05.12.2025.
> 
> my MR is supposed to check TypeId for the node so it should create a new object on second pass.

I commented on your PR.

oroulet:
I merged a similar solution just now

sebastianpfischer:
> I merged a similar solution just now

https://github.com/FreeOpcUa/opcua-asyncio/pull/1976/changes?

Thanks :)

When it comes to having 2 different structures that share the same name, how do I import them now from `ua.uatypes` ? I could not fully understand it out of the PR.

oroulet:
https://github.com/FreeOpcUa/opcua-asyncio/pull/1988

Issue #1918, opened 2026-01-23
Refactor dynamic struct generation to replace exec() with standard type creation


RinZ27:
Refactoring completed and merged in PR #1920. Closing this issue now. Thanks for the review!

PR #1920, opened 2026-01-24
Refactor structures104 to replace exec() with standard type creation

RinZ27:
Following the discussion in #1918, I've refactored `structures104.py` to eliminate the use of `exec()` when generating dynamic Python classes from UA data types.

The new implementation uses `dataclasses.make_dataclass` and `type()` to create classes and enums directly. This approach:
- Improves debuggability as classes are now standard Python objects.
- Enhances security by avoiding execution of arbitrary strings.
- Maintains full compatibility with the existing Type Hinting and Union property logic.

I've verified the changes with the existing test suite (`tests/test_custom_structures.py`), and all 25 tests passed. 

I've started with `structures104.py` as suggested. Let me know if you'd like me to apply similar changes to the older `structures.py` as well.

### DIFF
--- a/asyncua/common/structures104.py
+++ b/asyncua/common/structures104.py
@@ -196,7 +196,9 @@ def _make_union_property(name, idx):
     return property(getter, setter)
 
 
-def make_structure(data_type: ua.NodeId, struct_name: str, sdef: ua.StructureDefinition, log_error: bool = True) -> dict[str, type]:
+def make_structure(
+    data_type: ua.NodeId, struct_name: str, sdef: ua.StructureDefinition, log_error: bool = True
+) -> dict[str, type]:
     """
     given a StructureDefinition object, generate Python class
     """
@@ -235,9 +237,6 @@ def make_structure(data_type: ua.NodeId, struct_name: str, sdef: ua.StructureDef
 
     for idx, sfield in enumerate(sdef.Fields, start=1):
         fname = clean_name(sfield.Name)
-        if fname in seen_names:
-            _logger.warning("Field name %s is duplicated in structure %s, renaming to %s_%s", fname, struct_name, fname, idx)
-            fname = f"{fname}_{idx}"
         seen_names.add(fname)
         if sfield.DataType.NamespaceIndex == 0 and sfield.DataType.Identifier in ua.ObjectIdNames:
             if sfield.DataType.Identifier == 24:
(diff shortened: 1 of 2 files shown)


RinZ27 (commented):
I've addressed all the review comments. Specifically:
- Added return type hints to `get_default_value`, `make_structure`, and `make_enum` in `structures104.py`.
- Refactored type hint resolution into a new `resolve_type_hint` function in `ua_binary.py` to eliminate code duplication.
- Ensured `field_serializer` is consistent in handling its input.
- Confirmed that the `get_type_hints` call in `create_dataclass_serializer` is cached via `@functools.cache`, so it doesn't impact runtime parsing performance beyond the initial setup.
- Reverted unrelated changes in `udp.py` and `uaprocessor.py` as requested.
Verified with `tests/test_custom_structures.py` and everything is passing. Thanks for the feedback! @oroulet

RinZ27:
I've addressed all the review comments and further optimized the implementation. Specifically:
- Added return type hints to `get_default_value`, `make_structure`, and `make_enum` in `structures104.py`.
- Refactored type hint resolution into a new `resolve_type_hint` function in `ua_binary.py` to eliminate code duplication.
- Ensured `field_serializer` is consistent in handling its input.
- Further optimized `create_dataclass_serializer` by using the existing `get_safe_type_hints` helper. This function is cached via `@functools.cache`, ensuring parsing performance is not impacted.
- Reverted unrelated changes in `udp.py` and `uaprocessor.py` as requested.
Verified with `tests/test_custom_structures.py` and everything is passing. Ready for another look! @oroulet

oroulet:
I found out. you need to fix code adding parent fields:

```   def __add_recursion(sdef, desc):
        name = clean_name(desc.BrowseName.Name)
        if sdef:
            if parent_sdef:
                names = [f.Name for f in sdef.Fields]
                for sfield in reversed(parent_sdef.Fields):
                    if sfield.Name not in names:
                        sdef.Fields.insert(0, sfield)
            if isinstance(sdef, ua.StructureDefinition):
                dtypes.append(DataTypeSorter(desc.NodeId, name, desc, sdef))
            return _recursive_parse(
            ```

File: asyncua/common/structures104.py [hunk 1]
```diff
@@ -409,18 +408,20 @@ def __str__(self):
...
+        if isinstance(sdef, ua.StructureDefinition):
+            if isinstance(parent_sdef, ua.StructureDefinition):
+                names = [f.Name for f in sdef.Fields]
                 for sfield in reversed(parent_sdef.Fields):
-                    sdef.Fields.insert(0, sfield)
+                    if sfield.Name not in names:
+                        sdef.Fields.insert(0, sfield)
             if isinstance(sdef, ua.StructureDefinition):
                 dtypes.append(DataTypeSorter(desc.NodeId, name, desc, sdef))
             return _recursive_parse(
```
oroulet: REcursive_parse() return nothing. So return can be removed and typing can be -> None

oroulet:
ok. we try...

PR #1929, opened 2026-02-11
fix: resolve NameError for Byte during binary encoding of SessionSecurityDiagnosticsDataType

zarch:
# NameError: `Byte` is not defined during deserialization of `SessionSecurityDiagnosticsDataType`

## Summary

When `uals` (or any client code) reads a node whose value contains a `SessionSecurityDiagnosticsDataType` extension object, deserialization crashes with:

```
NameError: name 'Byte' is not defined. Did you mean: 'bytes'?
```

## Root Cause

The file `asyncua/ua/uaprotocol_auto.py` uses `from __future__ import annotations` (line 1), which turns **all** annotations into lazily-evaluated strings. It also imports `Byte` directly:

```python
from asyncua.ua.uatypes import SByte, Byte, Bytes, ByteString, ...
```

Most dataclass fields in this file use quoted `'ua.Xxx'` style annotations (e.g. `SessionId: 'ua.NodeId'`), but `SessionSecurityDiagnosticsDataType` has one field with a **bare** annotation:

```python
# line 10059
Encoding: Byte = field(default=0, repr=False, init=False, compare=False)
```

Because of `from __future__ import annotations`, this is stored as the string `"Byte"` rather than the actual `Byte` class.

When the deserializer in `ua_binary.py` calls `get_type_hints()` to resolve these string annotations, it passes `{"ua": ua}` as `globalns`:

```python
# ua_binary.py, _create_dataclass_deserializer()
resolved_fieldtypes = get_type_hints(objtype, {"ua": ua})
```

The problem: passing a custom `globalns` **replaces** the module's own globals entirely. The namespace `{"ua": ua}` can resolve `ua.NodeId`, `ua.String`, etc., but it does **not** contain the bare name `Byte`. So `eval("Byte", {"ua": ua}, ...)` raises `NameError`.

## Fix

Pass `globalns=None` so that `get_type_hints()` automatically uses the module globals of the class being inspected (which includes `Byte` and all other direct imports). Provide `{"ua": ua}` as `localns` instead, so that `ua.Xxx`-style annotations still resolve:

```python
# Before (broken)
resolved_fieldtypes = get_type_hints(objtype, {"ua": ua})

# After (fixed)
resolved_fieldtypes = get_type_hints(objtype, None, {"ua": ua})
```

The same fix applies to `get_safe_type_hints()` used by the serializer path.

## Additional changes

`asyncua/tools.py` — `_lsprint_1` and `_lsprint_long` now retry once on `UaError` before falling back to a graceful error message, so that  `uals` can continue listing remaining nodes when a single read fails (e.g. due to a truncated server response).

## Test plan

- Added `test_session_security_diagnostics_roundtrip` in `tests/test_unit.py` that round-trips a `SessionSecurityDiagnosticsDataType` through  `struct_to_binary` / `struct_from_binary`
- Fails on `master` with `TypeError: issubclass() arg 1 must be a class` (caused by `get_type_hints` failing to resolve `Byte`)
- Passes on this branch
- All 70 existing unit tests continue to pass

### DIFF
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -910,6 +910,27 @@ def test_struct_104() -> None:
     assert m == m2
 
 
+def test_session_security_diagnostics_roundtrip():
+    """Regression test: SessionSecurityDiagnosticsDataType has a bare
+    'Encoding: Byte' annotation (not quoted as 'ua.Byte'). With
+    from __future__ import annotations in uaprotocol_auto.py, this becomes
+    a forward reference that get_type_hints() must resolve using the
+    module globals. Previously, passing {"ua": ua} as globalns replaced
+    the module globals entirely, causing NameError: name 'Byte' is not defined.
+    """
+    obj = ua.SessionSecurityDiagnosticsDataType(
+        ClientUserIdOfSession="testuser",
+        ClientUserIdHistory=["user1", "user2"],
+        AuthenticationMechanism="Anonymous",
+        TransportProtocol="opc.tcp",
+        SecurityPolicyUri="http://opcfoundation.org/UA/SecurityPolicy#None",
+        ClientCertificate=b"\x00\x01\x02",
+    )
+    data = struct_to_binary(obj)
+    obj2 = struct_from_binary(ua.SessionSecurityDiagnosticsDataType, ua.utils.Buffer(data))
+    assert obj == obj2
+
+
 def test_builtin_type_variant():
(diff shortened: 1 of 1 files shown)


zarch:
Commit 560455e it is useful to handle errors like:
`asyncua.common.utils.NotEnoughData: Not enough data left in buffer, request for 6146, we have 41`

While using `uals` command.

oroulet:
you need to fix linting before any review

PR #1932, opened 2026-02-18
Fix uaread with username/password

basyskom-swid:
Credential + security handling is now the same as for all other tools


basyskom-swid:
@oroulet Sorry, I was away from the computer. Thanks!

Issue #1940, opened 2026-03-04
_create_type_deserializer -> RecursionError: maximum recursion depth exceeded in comparison


AndreasHeine:
Error on Clientside deserialization of an ISA95JobOrderDataType

As part of a Subscription of an ISA95JobOrderStatusEventType (https://reference.opcfoundation.org/ISA95JOBCONTROL/v200/docs/6.2.6)

which containes the following DataType Structures:

ISA95ParameterDataType:
https://reference.opcfoundation.org/ISA95JOBCONTROL/v200/docs/6.3.10

ISA95PropertyDataType:
https://reference.opcfoundation.org/ISA95JOBCONTROL/v200/docs/6.3.9

e.g.:
<img width="1532" height="953" alt="Image" src="https://github.com/user-attachments/assets/b34939df-b83b-4105-a80e-8167afb56f2d" />

it appears that there is a potential issue with recursive datatype definitions which contain ther own type in it self

AndreasHeine:
> on master?

asyncua==1.1.8

AndreasHeine:
@oroulet FYI: asyncua==1.1.8 has not fully the issue!

1.1.8 -> Only if you Use the Recusive type, if you specifie it in the JobOrder
master -> always, even if not used (Optional Field)

PR #1950, opened 2026-03-10
fix(ua_binary): avoid recursive list codec construction for mutual da…

AndreasHeine:
fix for: #1940


oroulet:
merged of it. might some stuff missing

Issue #1848, opened 2025-07-02
Changing system time


troutgoingupstream:
What would be the right course of action to refresh both server and client in a situation when I change the system time? We noticed a bunch of issues with that especially when moving backwards more than one hour.
We found out, that as a minimum we have to refresh all subscriptions on the client side, as opcua uses asyncio.sleep internally which waits for the original timestamp, and we didn’t get opc updated until the time „catched up” the original time.
But now we see 
asyncua.server.uaprocessor: sending service fault response: The session cannot be used because ActivateSession has not been called. (BadSessionNotActivated) 
error. 
Is the complete server and all clients shutdown and restart an only option/walkaround right now to mitigate that?

troutgoingupstream:
It’s on a device disconnected from the internet and without access to the local ntp server. Initial time set on this device might be wrong, so manual adjustment might be necessary.

kohtala:
There is still datetime being used.

kohtala:
The remaining datetime is now fixed with commit 237979cd2e7a91f8071acc6f2b176d4339bef11c.

If there is nothing else using wall clock, I think this can be closed.

PR #1969, opened 2026-05-10
Auto reconnect v2

oroulet:
@AndreasHeine since your MR has so many things I extracted the session thing and made claude clean things up.
Here is a proposition


AndreasHeine:
@oroulet looks great!

as long as we get to the point where we can have a session restored/reused/recreated + subscription transefered without data loss (republish) on a transport losses, i am fine with it!

https://reference.opcfoundation.org/Core/Part4/v104/docs/6.7
image

beyond that:
- a client can have multiple sessions
- subscription handler should decouple from usercode

the nasty part is just that everything is related...

oroulet:
@AndreasHeine another commit added to ad auto-reconnection.

What do you mean by subscription handler should be decoupled from user code?
if you mean not stop receviing loop I agree with that.

oroulet:
also concerning that multi-session thing. I suppose it is possible, It just sounds so crazy over engineered. Are some people really using that ?

AndreasHeine:
multi session in gateway solutions is inevitable due to roles and permissions bound to an identity maintaining multiple client connections (securechannels) is overhead

AndreasHeine:
> @AndreasHeine another commit added to ad auto-reconnection.
> 
> What do you mean by subscription handler should be decoupled from user code? if you mean not stop receviing loop I agree with that.

yes, usercode should not have negative impact on the execution of asyncua (even blocking code in the subhandler!)

AndreasHeine:
just as a reminder for ua_session: 
transfer_subscriptions + republish

is important for a successfull reconnect in order to continue at the last published notification

oroulet:
@AndreasHeine Can you precise? 
Is that a bug in that branch?
btw you are welcome to review and test. The main ideas are from your branch so you should recognise things

AndreasHeine:
Subscription Republish:

```python
 async def republish(
        self, subscription_id: int, retransmit_sequence_number: int
    ) -> ua.NotificationMessage:
        """Request republish of specific notification."""
        self.logger.debug("republish")
        request = ua.RepublishRequest()
        request.Parameters.SubscriptionId = subscription_id
        request.Parameters.RetransmitSequenceNumber = retransmit_sequence_number
        data = await self._send_session_request(request)
        response = struct_from_binary(ua.RepublishResponse, data)
        self.logger.debug(response)
        response.ResponseHeader.ServiceResult.check()
        # RepublishResponse can expose NotificationMessage directly (spec-compliant)
        # or behind a Parameters object in older generated protocol code.
        params = getattr(response, "Parameters", None)
        if params is not None:
            return params.NotificationMessage
        return response.NotificationMessage
```

so in a nutshell you inform the server after you transefered the subscription to your new session that he should resend from the last known (to the client) sequence_number

image

oroulet:
@AndreasHeine fixed. other things?

AndreasHeine:
image

is this intended?

oroulet:
something with a difference between client and server. I suppose we could try to put it back, but it only cosmetic

oroulet:
fixed anyway

AndreasHeine:
> something with a difference between client and server. I suppose we could try to put it back, but it only cosmetic

it is a guard for the node class to ensure nodeclass works from server and client side!

AndreasHeine:
@oroulet  i asked Copilot for a quick review of the implementation against the OPC UA Spec.: 

Reference: https://reference.opcfoundation.org/specs/OPC-10000-4/6.7 

The biggest difference is that this client does not implement the spec-preferred reconnect path exactly.

The spec says the client should first open a new SecureChannel and then re-activate the existing Session, and only create a new Session if ActivateSession fails. This implementation always tears the old session down and then reruns the full initial connect sequence, which always includes CreateSession and ActivateSession on a new session. You can see that in [client.py:393](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [client.py:768](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [client.py:799](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html). So on that point it is materially different from Part 4/6.7.

Other notable differences versus the spec:

Subscription recovery is aligned only with the “new session” branch of the spec. After reconnect it tries TransferSubscriptions, and if that fails it recreates the subscription and monitored items. That logic is in [client.py:818](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [subscription.py:450](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html). That matches the spec’s fallback path, but because the client always creates a new session, it uses this path every time instead of first trying to keep the existing session alive.

Republish handling is weaker than the spec text. The spec says to call Republish in a loop from the next expected sequence number until the server returns Bad_MessageNotAvailable. This implementation only republishs the sequence numbers advertised in [AvailableSequenceNumbers](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) from TransferSubscriptions, and stops on the first failure. See [subscription.py:478](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html). That is not the full spec algorithm.

There is no automatic ResendData or “read all monitored values” recovery path if republish cannot fill gaps. The spec explicitly recommends that fallback. I do not see that logic in the reconnect path.

The security-failure branch is incomplete relative to the spec. The spec says that if reconnect fails because of certificate/security changes, the client shall call GetEndpoints to refresh security information. This reconnect loop retries with backoff, but it does not automatically branch into GetEndpoints on Bad_CertificateInvalid or related security failures. See [client.py:768](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

What is aligned with the spec is that reconnect reuses the same client configuration: same security policy object, same credentials, same application identity held on the [Client](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) instance, and it blocks normal user requests while reconnect is in progress. That behavior is in [client.py:637](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [client.py:768](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

So the short answer is: yes, there is a real spec difference, and the main one is important. This implementation favors a simpler “new session plus transfer/recreate subscriptions” recovery strategy over the spec-preferred “new SecureChannel plus re-activate existing Session” strategy. That makes the code simpler, but it can lose some of the spec’s intended lossless recovery behavior.

AndreasHeine:
@oroulet re run against the Spec.:


Remaining gaps versus OPC UA Part 4, 6.7 are mainly these:

1. Connection monitoring strategy is still not spec-preferred
The spec prefers subscription keep-alive as primary health signal and recommends subscribing to ServerStatus.State.
Current implementation relies on periodic state reads and a stale-subscription watchdog instead. See [client.py:745](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [client.py:723](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Impact: works in practice, but is less aligned with normative “preferred mechanism” wording.

> Ok from my PoV

2. No ResendData fallback after unrecoverable notification gaps
The spec recommends ResendData (or reading all monitored values) if Republish cannot recover missing notifications.
Current flow republishes, then falls back to subscription recreation if recovery is incomplete. See [subscription.py:480](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [subscription.py:488](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [subscription.py:472](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Impact: potential temporary data divergence after hard gaps, especially for clients maintaining a strict cache.

> Ok from my PoV, current values gets send after recreation....

3. No built-in “stale overwrite” protection when replaying data
The spec warns clients not to overwrite newer local data with older republished values.
The library forwards republished notifications through normal publish_callback without an ordering guard beyond sequence processing. See [subscription.py:486](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [subscription.py:231](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Impact: application handlers must enforce timestamp/sequence monotonicity if that matters.

> Would be better to fix that i guess!?

4. Certificate/security refresh is endpoint-level only
On security failures, endpoint refresh is present and good, but there is no explicit automated policy for cert trust rollover handling beyond retrying after GetEndpoints. See [client.py:786](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [client.py:866](vscode-file://vscode-app/c:/Users/Anwender/AppData/Local/Programs/Microsoft%20VS%20Code/974500e64f/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Impact: still may require caller-side trust-store update workflow before reconnect can succeed.

> Ok from my PoV

asyncua/server/uaprocessor.py
```diff
@@ -189,6 +189,8 @@ async def _process_message(self, typeid, requesthdr, seqhdr, body):
                     raise ua.uaerrors.BadUserAccessDenied
 
         self.session_last_activity = datetime.now(timezone.utc)
```
kohtala, line 191: This datetime.now needs to go. It is used by the _session_watchdog_loop to timeout the session based on wall-clock time instead of the monotonic clock. There is issue #1848 about these timeout failures.

The session.touch should take care of it.

asyncua/server/internal_session.py
```diff
@@ -92,24 +111,59 @@ async def create_session(self, params: ua.CreateSessionParameters, sockname: tup
...
+        self._last_activity = time.monotonic()
+
+    def _start_timeout_watchdog(self) -> None:
+        if self.session_timeout is None or self.session_timeout <= 0:
+            return
+        if self._timeout_task is not None and not self._timeout_task.done():
+            return
+        self._timeout_task = asyncio.create_task(self._timeout_loop())
+
+    async def _timeout_loop(self) -> None:
```
kohtala, line 142: Is there some disadvantage with using `loop.call_later` in place of this task? The `touch()` could cancel the old `TimerHandle` and setup a new call_later. These unnecessarily waking loops can pull the processor unnecessarily out from power saving on mostly idle servers.

kohtala (commented):
I'm studying a session timeout issue and came from #1848 to see if this will fix the issue. I'd like to leave a couple of review comments for your consideration.

asyncua/server/internal_session.py
oroulet: the only reason it is a task, is that async task are easier on the brain than callback chains. that is the entire point of async as far I understood.
Also since we sleep in that loop, is it really different on hardware level from using timers?

asyncua/server/uaprocessor.py
oroulet: good point

oroulet: done

asyncua/server/internal_session.py
kohtala: Probably not that big difference considering power usage in practice.

This anyway raises some bad feelings I'm trying to put my finger on. I see you add a flag to prevent re-entry into the close_session from received message while the timeout is already closing the session. I'm not sure if there could be other paths from other messages messing with the session during the session close. Like CreateSubscription.

I usually try to move this kind of things all into the same task. If I read the code right, there is a task _process_received_message_loop for processing all protocol messages. The call_later callback for a timeout could post an item on that message queue to be handled in the same task with the received messages. Having the connection and session objects owned by a single task would prevent re-entry from multiple tasks and prevent otherwise difficult to avoid and find concurrency bugs.

I have not made myself familiar enough with the code to give any strong opinion. Perhaps you have something in place that prevents this. I hope this is helpful and not just waste of your time.

oroulet: this is the internal session, it is only closing when the server closes, if an event is missed here, this is not a big deal, we are anyway going to miss many things when closed...
and not you are not wasting any time, you made me fix a big issue right over and I saw you had an interesting MR that will need porting over that branch here

asyncua/server/uaprocessor.py
kohtala: Was the datetime supposed to be removed? This was merged to master, but the UaProcessor on master still sets session_last_activity and in the _session_watchdog_loop task closes the connection based on wall-clock time rather than using time.monotonic.

oroulet: I must have fucked up some rebasing. Will have a look

kohtala: If you had a commit with it, you probably find it in the reflog.

oroulet: in fact I was missing many commits, I merged them now. That issue should be fixed now

PR #1974, opened 2026-05-19
Implement PublishRequest limit, no subscriptions and timeout responses

kohtala:
The PublishRequest does not result in BadNoSubscription when there are no subscriptions. It also does not return a response at all with a timeout while it should return a ServiceFault with BadTimeout. Also, there is no limit to queued PublishRequest when it should result in BadTooManyPublishRequests for the oldest PublishRequest.

The publish request itself also unnecessarily had nested loops over all subscriptions and acks potentially affecting performance with large number of subscriptions and acks.

The ModifySubscriptionRequest was not part of the default roleruleset therefore resulting in BadUserAccessDenied.

The DeleteSubscription exception checking tested wrong variable.

The PublishRequest is missing tests in this PR. I'm not sure how such tests should be implemented.

### DIFF
--- a/asyncua/server/internal_session.py
+++ b/asyncua/server/internal_session.py
@@ -325,7 +325,7 @@ class InternalSession(AbstractSession):
 
         return subscription_result
 
-    def publish(self, acks: Iterable[ua.SubscriptionAcknowledgement] | None = None) -> None:
+    def publish(self, acks: Iterable[ua.SubscriptionAcknowledgement] | None = None) -> int:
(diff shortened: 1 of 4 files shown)


kohtala:
I added tests for the BadNoSubscription and BadTooManyPublishRequests, but the BadTimeout still needs some thought how to reproduce it for a test as we keep the PublishRequests in queue past the timeout until they are needed.

kohtala:
The BadTimeout is also tested.

PR #1978, opened 2026-05-24
Defer standard server address-space import until use

### LINKED ISSUE #1977: Defer generated standard address-space import until server initialization
Good day, @oroulet,

I have an OPC client-only application that I am currently auditing. I noticed that a clean `import asyncua` currently imports the generated server-address-space modules, even when no server is initialized.

The import path appears to be:
```python
import asyncua
  -> asyncua.__init__
  -> asyncua.server.Server
  -> asyncua.server.internal_server.InternalServer
  -> asyncua.server.standard_address_space.standard_address_space
  -> asyncua.server.standard_address_space.standard_address_space_services
  -> asyncua.server.standard_address_space.standard_address_space_services_1 ... _25
```
This means generated server-only address-space modules are loaded for client-only users and for code paths that only import the package. The result is an unnecessary baseline memory and import work for use cases that do not initialize a server. (Correct me if I am wrong, but I could not find a reason why the address-space has to be loaded when `asyncua` is initially imported rather than, say, at server initialization).

I measured this locally with a clean interpreter using:
```shell
PYTHONDONTWRITEBYTECODE=1 uv run python -X importtime check_asyncua_import.py 2> importtime.log
```
```python
# check_asyncua_import.py
#!/usr/bin/env python3
import importlib
import resource
import sys
import time

WATCH_PREFIXES = (
    "asyncua.server.standard_address_space.standard_address_space",
    "asyncua.server.standard_address_space.standard_address_space_services",
)


def rss_mb() -> float:
    # Linux reports ru_maxrss in KiB.
    rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    return rss / 1024


def main() -> int:
    start_rss = rss_mb()
    start = time.perf_counter()

    importlib.import_module("asyncua")

    elapsed_ms = (time.perf_counter() - start) * 1000
    end_rss = rss_mb()

    loaded = sorted(name for name in sys.modules if any(name.startswith(prefix) for prefix in WATCH_PREFIXES))

    print(f"Python: {sys.executable}")
    print(f"asyncua: {importlib.import_module('asyncua').__file__}")
    print(f"import asyncua time: {elapsed_ms:.1f} ms")
    print(f"max RSS increase: {end_rss - start_rss:.1f} MB")

    if loaded:
        print("\nGenerated standard address-space modules loaded during import:")
        for name in loaded[:50]:
            print(f"  - {name}")
        if len(loaded) > 50:
            print(f"  ... {len(loaded) - 50} more")
        return 1

    print("\nOK: generated standard address-space modules were not loaded during import.")
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

Eagerly loading the server-address-space import shows:
```shell
import asyncua time: 183.4 ms
max RSS increase: 27.9 MB

Generated standard address-space modules loaded during import
...
```

While deferring the import:
```shell
import asyncua time: 175.2 ms
max RSS increase: 17.7 MB  # a 37% decrease!

OK: generated standard address-space modules were not loaded during import.
```

The wall-clock import-time improvement is modest (only about 10ms) largely due to remaining import costs being elsewhere (especially the generated UA protocol modules). However, the change removes an unnecessary server-only generated-module import from the baseline `import asyncua` path and reduced clean-import RSS growth by about 10MB in this local check.

I should also note that my corporate machine is quite powerful, so wall-clock may be larger on others' PCs.

**PROPOSED CHANGE**

The fix is quite simple, actually. We just need to move the `standard_address_space` import from the module scope in `asyncua.server.internal_server` to `InternalServer.load_standard_address_space()`, immediately before it is used.

Conceptually:
```python
    async def load_standard_address_space(self, shelf_file: Path | None = None) -> None:
        if shelf_file:
            if shelf_file.is_file() or shelf_file.with_suffix(".db").is_file():  # noqa: ASYNC240
                # import address space from shelf
                self.aspace.load_aspace_shelf(shelf_file)
                return
        # import address space from code generated from xml
        from .standard_address_space import standard_address_space ### moved from the module-level import to here

        await asyncio.to_thread(standard_address_space.fill_address_space, self.node_mgt_service)
        # import address space directly from xml, this has performance impact so disabled
        # importer = xmlimporter.XmlImporter(self.node_mgt_service)
        # importer.import_xml("/path/to/python-asyncua/schemas/Opc.Ua.NodeSet2.xml", self)
        if shelf_file:
            # path was supplied, but file doesn't exist - create one for next start up
            await asyncio.to_thread(self.aspace.make_aspace_shelf, shelf_file)
```

This should preserve existing behavior:
- `Server.init()` still loads the standard address space.
- shelf/cache behavior is unchanged.
- public imports such as `from asyncua import Server` are unchanged.
- generated address-space code is still imported before `fill_address_space()` is called.

I will happily submit a small PR for this and you can accept or decline the change at your leisure.

Best

dakotahorstman:
This PR defers importing the generated standard server address-space modules until `InternalServer.load_standard_address_space()` actually needs them.

Currently, `asyncua.server.internal_server` imports `standard_address_space` at module import time. Because `asyncua.__init__` imports `Server`, this means a plain `import asyncua` also imports the generated server standard address-space modules, even when no server is initialized (such as in client-only applications).

The generated standard address space is only needed when a server loads the standard address-space, so this PR moves that import to the point of use.

Before this change, measuring RSS:
```shell
import asyncua time: 183.4 ms
max RSS increase: 27.9 MB
Generated standard address-space modules loaded during import
```

After this change:
```
import asyncua time: 175.2 ms
max RSS increase: 17.7 MB
OK: generated standard address-space modules were not loaded during import.
```

Fixes #1977

### DIFF
--- a/asyncua/server/internal_server.py
+++ b/asyncua/server/internal_server.py
@@ -26,7 +26,6 @@ from .address_space import AddressSpace, AttributeService, MethodService, NodeDa
 from .event_generator import EventGenerator
 from .history import HistoryManager
 from .internal_session import InternalSession
-from .standard_address_space import standard_address_space
 from .subscription_service import SubscriptionService
 from .user_managers import PermissiveUserManager, UserManager
 
@@ -155,6 +154,8 @@ class InternalServer:
                 self.aspace.load_aspace_shelf(shelf_file)
                 return
         # import address space from code generated from xml
+        from .standard_address_space import standard_address_space
+
         await asyncio.to_thread(standard_address_space.fill_address_space, self.node_mgt_service)
         # import address space directly from xml, this has performance impact so disabled
         # importer = xmlimporter.XmlImporter(self.node_mgt_service)
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -6,6 +6,8 @@ Tests that can only be run on server side must be defined here
 import asyncio
 import logging
 import shelve
+import subprocess
+import sys
 from datetime import timedelta
 from enum import EnumMeta
 from pathlib import Path
@@ -905,3 +907,30 @@ async def test_start_server_when_port_is_in_use(server: Server):
         await server2.start()
     # now it should still be possible to stop the server with exceptions
     await server2.stop()
+
+
+def test_import_asyncua_does_not_import_generated_standard_address_space_services():
+    """
+    Regression test verifying that importing top-level asyncua does not
+    import the server standard address-space.
+
+    See https://github.com/FreeOpcUa/opcua-asyncio/issues/1977
+    """
+    code = """
+import sys
+import asyncua
+
+loaded = [
+    name for name in sys.modules
+    if (
+        name.startswith("asyncua.server.standard_address_space.standard_address_space_services")
+        or name == "asyncua.server.standard_address_space.standard_address_space"
+    )
+]
+assert not loaded, loaded
+"""
+
+    subprocess.run(
+        [sys.executable, "-c", code],
+        check=True,
+    )
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -930,6 +930,8 @@ loaded = [
 assert not loaded, loaded
 """
 
+    # Subprocess required to avoid interference from other tests that
+    # may already have imported server modules.
     subprocess.run(
         [sys.executable, "-c", code],
         check=True,


oroulet:
Can't we do that even better by not importing the server namespace at all when only using client api? just wondering

oroulet:
seems hard. I merge that code

PR #1991, opened 2026-06-25
Return ServiceFault for bad subscription ID and return ack results in PublishResult

kohtala:
The Create/Modify/DeleteMonitoredItemsRequests returned Good service result with bad SubscriptionId and each operation result was BadSubscriptionIdInvalid. The OPC-UA specification however asks for the BadSubscriptionIdInvalid to be returned as a service result.

The PublishResponse Results was always empty while the specification asks for it to return the results of subscription acknowledgements.

### DIFF
--- a/asyncua/common/subscription.py
+++ b/asyncua/common/subscription.py
@@ -17,6 +17,7 @@ from typing import TYPE_CHECKING, Any, Protocol, overload
 from asyncua import ua
 from asyncua.client.ua_session import UaSession
 from asyncua.common.ua_utils import copy_dataclass_attr
+from asyncua.common.utils import ServiceError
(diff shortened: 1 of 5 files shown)


oroulet:
@kohtala some lint issues here

kohtala:
> @kohtala some lint issues here

@oroulet, I had missed what had happened earlier and only saw the error today. It is now fixed to pass.

VSCode is complaining the `uaprotocol_auto.py` typing references undefined `ua`. I was trying to add the import and it was fine in VSCode, but not for mypy. But mypy then does not complain about the undefined `ua`. Strange.

It might be the `ua` is not needed because the types are imported directly in the module.

PR #1997, opened 2026-07-17
Client reverse connect

still-wondering:
## Client-side reverse connect implementation
Adds support for the OPC UA **Reverse Connect** on the client side.

Spec Reference: https://reference.opcfoundation.org/specs/OPC-10000-6/7.1.3

### Changes
- `RCServer`: asyncio server accepting new connections, handling the `ReverseHello` and yielding the connection to be handled as a normal UA connection
- `RCClient`: a `Client` subclass using `RCServer`, instead of normal dial to connect
- Sync API: `SyncRCServer` and sync `RCClient`
- Some extra typing updates on the way 

### API usage example
Single RC client:
```python
client = RCClient(RCServer("127.0.0.1", 60555), rc_timeout=60) # only wait for incoming connections for 60s
# set up certs/auth as needed here
async with client:
    # client is connected to the server here, use as a usual client
```

Multiple RC clients (callbacks-like handling):
```python
tasks: list[asyncio.Task] = [] # for python>=3.11 use asyncio.TaskGroup
# use exit stack for dynamic context management
async with AsyncExitStack() as stack, RCServer("127.0.0.1", 60555) as server:
    while True:
        client = RCClient(server, rc_timeout=None) # disable timeout to wait for connections indefinitely 
        # set up certs/auth here as needed
        await stack.enter_async_context(client) # connect - wait for reverse connection from server
        tasks.append(asyncio.create_task(callback(client)))
```


oroulet:
what is difference between that branch and the one here: https://github.com/FreeOpcUa/opcua-asyncio/pull/1947

still-wondering:
hi @oroulet, the #1947 implements the reverse connection for the OPC UA server while this branch implements the reverse connection for the OPC UA client, so they should be complementary

asyncua/client/rc/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Client reverse connect."""
```
oroulet: rc is too opaque as a name. call that reverse for example

oroulet: or reverse_connection

still-wondering: changed to `reverse_connect`, is that all right?

oroulet:
some claude review:

```
## Review findings

### Critical

**1. Memory-exhaustion DoS via attacker-controlled `body_size`** — `asyncua/client/reverse_connect/rc_server.py:134`
`RCProtocol` buffers incoming data until the full `body_size` from the header has arrived, with no cap. Any peer that can reach the RC listening port can send `b'RHE' + b'F' + struct.pack('<I', 0xFFFFFFFF)` and stream data: `header_from_binary` applies no limit, `_parse_rc_hello` keeps returning `None`, and `rec_buf` grows toward ~4 GB per connection before the 4095-char URI/URL checks ever run. `slow_connection_timeout` defaults to `None`, so there is no time bound either. A legal ReverseHello body is at most ~8.2 KB — reject `header.body_size` above a small cap right after `_parse_header`. Note also that rebuilding `Buffer(bytes(self.rec_buf))` on every chunk makes parsing quadratic in chunk count.

**2. Sync `RCClient` + `SyncRCServer` never deliver a connection** — `asyncua/sync.py:579`
The sync `RCClient` and `SyncRCServer` each create their own `ThreadLoop`, so the shared `asyncio.Queue` is used across two event loops: the server and `RCProtocol.put_nowait` run on loop A while `connect()` awaits `next_rc()`/`Queue.get` on loop B (woken via non-threadsafe `call_soon` from the wrong loop). Reproduced on the branch: the connection is queued within 0.3 s but `connect()` always ends in `TimeoutError` (or hangs forever with `rc_timeout=None`). `RCClient` should reuse `rc_server.tloop` when handed a `SyncRCServer`.

### Bugs

**3. `RCServer.stop()` never wakes waiters blocked in `next_rc()`** — `asyncua/client/reverse_connect/rc_server.py:247`
`stop()` swaps in a fresh queue but coroutines already parked in `next_rc()`/`Que that nothing will ever feed or cancel. In the shared-server pattern (exactlyFileNotFoundError, and the new MessageType/ChunkType enums keep per-member annotations (Hello: bytes = b'HEL') which the typing spec treats as non-members — mypy is only quiet because misc
is globally disabled.

A few lower-priority cleanups (tloop-bootstrap duplication in sync.py, a third hand-rolled framing loop, narrating comments, a missing -> None in the example) were cut by the 10-finding cap. If you want, I can draft these as review comments on the PR.

✻ Cogitated for 26m 11s

※ recap: I reviewed PR #1997 (reverse-connect feature) and reported 10 verified findings, the worst being a memory-exhaustion DoS on the listener and a broken sync wrapper. Next: say if you want them posted as PR comments. (disable recaps in /config)

❯ show me all findings as a list. I can put that as one big comment on MR

Here's the full review as one paste-ready markdown comment (raw source in the fence below):

## Review findings

### Critical

**1. Memory-exhaustion DoS via attacker-controlled `body_size`** — `asyncua/client/reverse_connect/rc_server.py:134`
`RCProtocol` buffers incoming data until the full `body_size` from the header has arrived, with no cap. Any peer that can reach the RC listening port can send `b'RHE' + b'F' + struct.pack('<I', 0xFFFFFFFF)` and stream data: `header_from_binary` applies no limit, `_parse_rc_hello` keeps returning `None`, and `rec_buf` grows toward ~4 GB per connection before the 4095-char URI/URL checks ever run. `slow_connection_timeout` defaults to `None`, so there is no time bound either. A legal ReverseHello body is at most ~8.2 KB — reject `header.body_size` above a small cap right after `_parse_header`. Note also that rebuilding `Buffer(bytes(self.rec_buf))` on every chunk makes parsing quadratic in chunk count.

**2. Sync `RCClient` + `SyncRCServer` never deliver a connection** — `asyncua/sync.py:579`
The sync `RCClient` and `SyncRCServer` each create their own `ThreadLoop`, so the shared `asyncio.Queue` is used across two event loops: the server and `RCProtocol.put_nowait` run on loop A while `connect()` awaits `next_rc()`/`Queue.get` on loop B (woken via non-threadsafe `call_soon` from the wrong loop). Reproduced on the branch: the connection is queued within 0.3 s but `connect()` always ends in `TimeoutError` (or hangs forever with `rc_timeout=None`). `RCClient` should reuse `rc_server.tloop` when handed a `SyncRCServer`.

### Bugs

**3. `RCServer.stop()` never wakes waiters blocked in `next_rc()`** — `asyncua/client/reverse_connect/rc_server.py:247`
`stop()` swaps in a fresh queue but coroutines already parked in `next_rc()`/`Queue.get` stay on the old queue that nothing will ever feed or cancel. In the shared-server pattern (exactly `multiple_clients_per_server` in the example), a client with `rc_timeout=None` hangs forever when another task stops the server; with a finite timeout the caller gets a misleading `TimeoutError` instead of "server stopped".

**4. `SyncRCServer` never stops its own `ThreadLoop`; interpreter can't exit**
`close_tloop=True` is set for a self-created `ThreadLoop` but neither `stop()` nor `__exit__` reads it. `ThreadLoop` is a non-daemon `Thread` running `loop.run_forever()`, so the process hangs after the last line of the script. Compare `sync.Client.disconnect`, which does the `if self.close_tloop: self.tloop.stop()` dance.

**5. `attach_socket` swallows a fatal ERR delivered in `leftover_data`** — `asy
If the server sends ReverseHello immediately followed by ERR (e.g. `BadTcpServerTooBusy`), `p.data_received(leftover_data)` routes it to `_process_received_message`, which closes the transport and raises `UaStatusCodeError` — but `_process_received_data` catches it and `_fail_all_pending` iterates an empty callback map, so nothing propagates. `attach_socket` then reports `SOCKET_OPEN` on a closed transport, and the subsequent hello fails with a generic `TimeoutError`, the server's status code visible only in logs. Same outcome if the peer FINed before the handoff, since `connection_lost` went to the discarded `RCProtocol`.

**6. `auto_reconnect` silently degrades RCClient to a forward dial** — `asyncua/client/reverse_connect/rc_client.py:46`
`rc_client.connect(auto_reconnect=True)` overrides the `__init__` default. On transport loss the supervisor calls `_reconnect_with_backoff → _try_reactivate_existing_session → connect_socket()`, which dials the host/port taken from the ReverseHello `EndpointUrl` instead of waiting for a new reverse connection via `_connect_sequence`. For a server behind NAT/firewall (the whole point of reverse connect) this retries forever and never reconnects. `RCClient` should reject `auto_reconnect` or route reconnection through `_connect_sequence`.

**7. Server ownership sampled once at `__init__`; `start()` has no double-start guard** — `asyncua/client/reverse_connect/rc_client.py:42`
Constructing several `RCClient(server)` instances before awaiting `server.start()` makes every client capture ownership (since `is_listening` was false), so each `connect()` enters the `RCServer` context and calls `start()` on the already-listening server — `create_server` rebinding the same port raises `OSError: address already in use`. On other interleavings the first client's context exit `stop()`s the shared listener and strands the other clients in `next_rc()`. Re-check `is_listening` at connect time and guard `start()` against double-start with a clear `RuntimeError`.

**8. Example certificate path doesn't exist** — `examples/client-reverse-connection.py:11`
`CERT_BASE = Path(__file__).parent / 'examples' / 'certificates' / 'chain'` resolves to `examples/examples/certificates/chain`; the real files live at `examples/certificates/chain/`. The shipped example raises `FileNotFoundError` in `set_security`. Should be `Path(__file__).parent / 'certificates' / 'chain'`. Related: the `load_client_certificate`/`load_private_key` calafter `set_security` additionally switch the session to certificate user-auth,  not intend.

### Design / cleanup

**9. Dead `ready_clients.full()` backpressure branch** — `asyncua/client/reverse_connect/rc_server.py:107`
`self._ready_clients = asyncio.Queue[ReverseConnection]()` (line 216) has no `maxsize`, so `full()` is always `False` and the "queue full, dropping connection" branch can never trigger. A long-running shared server with slow consumers parks every accepted transport in `pause_reading` in the queue forever — fd growth until exhaustion, and the remote server sees N stale accepted connections. Either delete the branch or add the intended `maxsize` parameter.

**10. Annotated Enum members are non-members per the typing spec** — `asyncua/ua/uaprotocol_hand.py:37`
`MessageType`/`ChunkType` keep per-member annotations after the Enum conversion (`Invalid: bytes = b'INV'`). Per the typing spec an annotated assignment in an Enum body is a plain attribute, not a member — pyright flags it today, and mypy only stays quiet because `misc` is globally disabled. Drop the annotations (`Hello = b'HEL'`). Also note `Header.MessageType` is annotated as the enum while `header_from_binary` (`ua_binary.py:858`) still assigns raw wire bytes — equality works via the bytes mixin, but identity checks or `match` statements would break on the receive path.

### Minor (noted, not blocking)

- `tloop` bootstrap logic in `sync.py` is duplicated from `sync.Client`/`sync.Server` — could be factored out.
- The RC framing loop is a third hand-rolled copy of the header/body reassembly logic that already exists in client and server transports.
- Several comments narrate what the next line does; repo convention is no comments unless they explain a non-obvious why.
- Missing `-> None` on the example's async functions.
- Pre-existing: `AsymmetricAlgorithmHeader.__str__` prints `size1`/`size2` swapped (not introduced by this PR).

Two suspected issues were checked and dropped as false positives: a missing `wait_closed()` after closing the listener (listening sockets close synchronously here) and a supposed `wait_for`/`Queue.get` cancellation race dropping a queued connection (verified on Python 3.10–3.14 that the item is never lost).
it.
- The sync wrapper doesn't work (sync.py:579): sync RCClient and SyncRCServer eop, so the shared asyncio.Queue is touched from two event loops. Reproduced onthe branch: the reverse connection is queued on one loop but connect() waits on the other and always times out. The client should reuse the server's tloop.

Lifecycle/shutdown bugs: RCServer.stop() swaps in a fresh queue without waking coroutines already blocked in next_rc(), so shared-server waiters hang forever; and SyncRCServer sets close_tloop=True but never reads it, so the non-daemon loop thread keeps the interpreter alive after the script ends.

Error-handling and design gaps: attach_socket swallows a fatal ERR delivered ine socket open, turning the server's status code into a generic timeout;
auto_reconnect=True on an RCClient silently falls back to forward-dialing the Refeats reverse connect for a server behind NAT; and server ownership is sampledonce at RCClient.__init__, so constructing clients before starting a shared server ends in a confusing EADDRINUSE.

Smaller items: the ready_clients.full() backpressure branch is dead code (queue has no maxsize), the shipped example computes a cert path under examples/examples/… and crashes with
FileNotFoundError, and the new MessageType/ChunkType enums keep per-member anno') which the typing spec treats as non-members — mypy is only quiet because misc is globally disabled.

A few lower-priority cleanups (tloop-bootstrap duplication in sync.py, a third hand-rolled framing loop, narrating comments, a missing -> None in the example) were cut by the 10-finding cap. If you want, I can draft these as review comments on the PR.

✻ Cogitated for 26m 11s

※ recap: I reviewed PR #1997 (reverse-connect feature) and reported 10 verified findings, the worst being a memory-exhaustion DoS on the listener and a broken sync wrapper. Next: say if you want them posted as PR comments. (disable recaps in /config)

❯ show me all findings as a list. I can put that as one big comment on MR

Here's the full review as one paste-ready markdown comment (raw source in the f

## Review findings

### Critical

**1. Memory-exhaustion DoS via attacker-controlled `body_size`** — `asyncua/client/reverse_connect/rc_server.py:134`
`RCProtocol` buffers incoming data until the full `body_size` from the header has arrived, with no cap. Any peer that can reach the RC listening port can send `b'RHE' + b'F' + struct.pack('<I', 0xFFFFFFFF)` and stream data: `header_from_binary` applies no limit, `_parse_rc_hello` keeps returning `None`, and `rec_buf` grows toward ~4 GB per connection before the 4095-char URI/URL checks ever run. `slow_connection_timeout` defaults to `None`, so there is no time bound either. A legal ReverseHello body is at most ~8.2 KB — reject `header.body_size` above a small cap right after `_parse_header`. Note also that rebuilding `Buffer(bytes(self.rec_buf))` on every chunk makes parsing quadratic in chunk count.

**2. Sync `RCClient` + `SyncRCServer` never deliver a connection** — `asyncua/sync.py:579`
The sync `RCClient` and `SyncRCServer` each create their own `ThreadLoop`, so the shared `asyncio.Queue` is used across two event loops: the server and `RCProtocol.put_nowait` run on loop A while `connect()` awaits `next_rc()`/`Queue.get` on loop B (woken via non-threadsafe `call_soon` from the wrong loop). Reproduced on the branch: the connection is queued within 0.3 s but `connect()` always ends in `TimeoutError` (or hangs forever with `rc_timeout=None`). `RCClient` should reuse `rc_server.tloop` when handed a `SyncRCServer`.

### Bugs

**3. `RCServer.stop()` never wakes waiters blocked in `next_rc()`** — `asyncua/client/reverse_connect/rc_server.py:247`
`stop()` swaps in a fresh queue but coroutines already parked in `next_rc()`/`Queue.get` stay on the old queue that nothing will ever feed or cancel. In the shared-server pattern (exactly `multiple_clients_per_server` in the example), a client with `rc_timeout=None` hangs forever when another task stops the server; with a finite timeout the caller gets a misleading `TimeoutError` instead of "server stopped".

**4. `SyncRCServer` never stops its own `ThreadLoop`; interpreter can't exit** — `asyncua/sync.py:525`
`close_tloop=True` is set for a self-created `ThreadLoop` but neither `stop()` nor `__exit__` reads it. `ThreadLoop` is a non-daemon `Thread` running `loop.run_forever()`, so the process hangs after the last line of the script. Compare `sync.Client.disconnect`, which does the `if self.close_tloop: self.tloop.stop()` dance.

**5. `attach_socket` swallows a fatal ERR delivered in `leftover_data`** — `asyncua/client/ua_client.py:572`
If the server sends ReverseHello immediately followed by ERR (e.g. `BadTcpServerTooBusy`), `p.data_received(leftover_data)` routes it to `_process_received_message`, which closes the transport and raises `UaStatusCodeError` — but `_process_received_data` catches it and `_fail_all_pending` iterates an empty callback map, so nothing propagates. `attach_socket` then reports `SOCKET_OPEN` on a closed transport, and the subsequent hello fails with a generic `TimeoutError`, the server's status code visible only in logs. Same outcome if the peer FINed before the handoff, since `connection_lost` went to the discarded `RCProtocol`.

**6. `auto_reconnect` silently degrades RCClient to a forward dial** — `asyncua/client/reverse_connect/rc_client.py:46`
`rc_client.connect(auto_reconnect=True)` overrides the `__init__` default. On transport loss the supervisor calls `_reconnect_with_backoff → _try_reactivate_existing_session → connect_socket()`, which dials the host/port taken from the ReverseHello `EndpointUrl` instead of waiting for a new reverse connection via `_connect_sequence`. For a server behind
NAT/firewall (the whole point of reverse connect) this retries forever and neveld reject `auto_reconnect` or route reconnection through `_connect_sequence`.

**7. Server ownership sampled once at `__init__`; `start()` has no double-start guard** — `asyncua/client/reverse_connect/rc_client.py:42`
Constructing several `RCClient(server)` instances before awaiting `server.start()` makes every client capture ownership (since `is_listening` was false), so each `connect()` enters the `RCServer` context and calls `start()` on the already-listening server — `create_server` rebinding the same port raises `OSError: address already in use`. On other interleavings the first client's context exit `stop()`s the shared listener and strands the other clients in `next_rc()`. Re-check `is_listening` at connect time and guard `start()` against double-start with a clear `RuntimeError`.

**8. Example certificate path doesn't exist** — `examples/client-reverse-connection.py:11`
`CERT_BASE = Path(__file__).parent / 'examples' / 'certificates' / 'chain'` resolves to `examples/examples/certificates/chain`; the real files live at `examples/certificates/chain/`. The shipped example raises `FileNotFoundError` in `set_security`. Should be `Path(__file__).parent / 'certificates' / 'chain'`. Related: the `load_client_certificate`/`load_private_key` calls
after `set_security` additionally switch the session to certificate user-auth,  not intend.

### Design / cleanup

**9. Dead `ready_clients.full()` backpressure branch** — `asyncua/client/reverse_connect/rc_server.py:107`
`self._ready_clients = asyncio.Queue[ReverseConnection]()` (line 216) has no `maxsize`, so `full()` is always `False` and the "queue full, dropping connection" branch can never trigger. A long-running shared server with slow consumers parks every accepted transport in `pause_reading` in the queue forever — fd growth until exhaustion, and the remote server sees N stale accepted connections. Either delete the branch or add the intended `maxsize` parameter.

**10. Annotated Enum members are non-members per the typing spec** — `asyncua/ua/uaprotocol_hand.py:37`
`MessageType`/`ChunkType` keep per-member annotations after the Enum conversion (`Invalid: bytes = b'INV'`). Per the typing spec an annotated assignment in an Enum body is a plain
attribute, not a member — pyright flags it today, and mypy only stays quiet becled. Drop the annotations (`Hello = b'HEL'`). Also note `Header.MessageType` isannotated as the enum while `header_from_binary` (`ua_binary.py:858`) still assigns raw wire bytes — equality works via the bytes mixin, but identity checks or `match` statements would break on the receive path.

### Minor (noted, not blocking)

- `tloop` bootstrap logic in `sync.py` is duplicated from `sync.Client`/`sync.St.
- The RC framing loop is a third hand-rolled copy of the header/body reassembly client and server transports.
- Several comments narrate what the next line does; repo convention is no comme-obvious why.
- Missing `-> None` on the example's async functions.
- Pre-existing: `AsymmetricAlgorithmHeader.__str__` prints `size1`/`size2` swapR).

Two suspected issues were checked and dropped as false positives: a missing `wait_closed()` after closing the listener (listening sockets close synchronously here) and a supposed `wait_for`/`Queue.get` cancellation race dropping a queued connection (verified on Python 3.10–3.14 that the item is never lost).

``

still-wondering:
@oroulet, review incorporated:

Out of scope:
• **1** — Memory-exhaustion DoS via unbounded `body_size`. This is a real problem, but it also exists in normal client and normal server. As mentioned in other comment, 
all 3 of them have a message parsing loop implemented almost the same way 3 times. IMHO, it would be better to address this in a separate PR, as it will way to many changes (on top of the current ones).
The same also goes for message limits as per the specification.

Fixed:

• **2** — Sync `RCClient`+`SyncRCServer` cross-loop deadlock. Sync `RCClient` can now only interact with `SyncRCServer` via `SyncRCServer.create_client()`, which passes server's tloop into the `RCClient` it builds, so sync client and sync server now always run on the same event loop/thread. Tests added.
• **3** — `RCServer.stop()` never woke `next_rc()` waiters. `stop()` now tracks pending `next_rc()` tasks in `self._waiters` and cancels them, so both `rc_timeout=None` and finite-timeout callers unblock immediately instead of hanging or timing out misleadingly. Tests added.
• **4** — `SyncRCServer` leaked its `ThreadLoop` thread. `stop()` now closes the loop, same as `Client.disconnect`.
• **5** — `attach_socket` swallowed a fatal ERR in `leftover_data`. Added `UASocketProtocol.attach_transport()`, which registers a future in the callback map before feeding in leftover data, then re-raises the exception `UASocketProtocol._fail_all_pending()` set on it, if any. Tests added.
• **6** — `auto_reconnect=True` silently degraded to forward-dialing. `RCClient.connect()` always passes `auto_reconnect=False` to `Client.connect()`, ignoring the arguments.
• **7** — Server ownership race / no double-start guard. `RCServer.start()` now sets `_is_listening = True` synchronously before the `await create_server(...)` call, raising a `RuntimeError` on double-start. Tests added.
• **9** — Dead `ready_clients.full()` backpressure branch. Added a `max_pending_clients` parameter passed into `asyncio.Queue(maxsize=...)` allowing to limit the number of ready clients. Tests added.
• **8** — Example certificate path and **10** — Annotated Enum members / raw bytes into enum field. Fixed.


Other comments were either duplicates, not-related to the PR or AI hallucinations.

PR #1998, opened 2026-07-19
Limit concurrent client requests

### LINKED ISSUE #1987: Concurrent WriteRequests on a shared Client session timeout in asyncua 2.0
# Concurrent WriteRequests on a shared Client session timeout in asyncua 2.0

## Environment

* asyncua: 2.0
* Python: 3.14.5
* OpenSSL: 3.6.3
* Server: OPC UA server (PLC-backed in our case)
* Transport: opc.tcp

## Description

After upgrading to asyncua 2.0, concurrent Write service calls issued through a single shared `Client` instance consistently timeout.

The same application logic worked previously on an earlier asyncua release.

Single writes and single low-level `WriteRequest`s continue to work correctly.

## Observed behavior

The application performs an initialization step that writes a large number of node values.

When multiple tasks execute concurrently and share the same asyncua `Client`, requests eventually timeout inside:

```python
results = await client.uaclient.write(params)
```

with:

```text
asyncio.exceptions.CancelledError

...

TimeoutError

...

Exception: Unhandled exception while sending request to OPC UA server
```

The timeout occurs inside:

```text
asyncua/client/ua_client.py
asyncua/client/ua_session.py
```

## Minimal write path

```python
params = ua.WriteParameters()
params.NodesToWrite = write_values

results = await client.uaclient.write(params)
```

where `write_values` contains standard `ua.WriteValue` instances:

```python
ua.WriteValue(
    NodeId=ua.NodeId.from_string(nodeid),
    AttributeId=ua.AttributeIds.Value,
    Value=data_value,
)
```

## What works

A single low-level write succeeds:

```python
dv = ua.DataValue()
dv.Value = ua.Variant(False, ua.VariantType.Boolean)

wv = ua.WriteValue(
    NodeId=ua.NodeId.from_string(nodeid),
    AttributeId=ua.AttributeIds.Value,
    Value=dv,
)

params = ua.WriteParameters()
params.NodesToWrite = [wv]

results = await client.uaclient.write(params)
```

Result:

```text
StatusCode(value=0)
```

## What fails

Concurrent tasks sharing the same client:

```python
await asyncio.gather(
    task_1(),
    task_2(),
    task_3(),
    ...
)
```

where each task eventually executes:

```python
await client.uaclient.write(params)
```

against the same `Client` instance.

This eventually produces request timeouts.

## Workaround

Serializing the writes fixes the issue completely:

```python
for task in tasks:
    await task()
```

or equivalently:

```python
write_lock = asyncio.Lock()

async with write_lock:
    await client.uaclient.write(params)
```

After serialization, all writes complete successfully and initialization finishes normally.

## Additional information

The same application behavior worked prior to upgrading to asyncua 2.0. No application-side changes were made to the write logic other than updating dependencies.

Because:

* Single writes succeed
* Low-level `WriteRequest`s succeed
* Serializing requests succeeds
* Concurrent requests on a shared client/session timeout

it appears that either:

1. Concurrent service calls on a shared `Client`/session are no longer supported, or
2. There may be a regression in request handling/correlation when multiple writes are in flight simultaneously.

## Question

Is concurrent use of:

```python
client.uaclient.write(...)
```

on a shared `Client`/session expected to be supported in asyncua 2.0?

If not, could the documentation clarify that service calls should be serialized per client/session?

If it is expected to be supported, this appears to be a regression compared to previous asyncua versions.

sankalpsthakur:
## Summary

- keep a high private `UaClient._max_concurrent_requests` default of 100
- limit transport requests with one asyncio semaphore
- let `UaClient.set_max_concurrent_requests()` own both configuration and semaphore replacement
- expose `Client.set_max_concurrent_requests()` as a thin public delegation layer
- expose the same setter on sync `Client` for API parity

## Why

Some OPC UA servers and PLCs stop responding when many large service requests are pipelined at once. A high default preserves normal concurrency while allowing applications to lower the limit for constrained servers.

Fixes #1987.

## Validation

- request-limiter behavior tests
- sync Client setter parity test
- Ruff / mypy on changed modules

### DIFF
--- a/asyncua/client/client.py
+++ b/asyncua/client/client.py
@@ -235,6 +235,10 @@ class Client:
         """
         self._locale = locale
 
+    def set_max_concurrent_requests(self, max_concurrent_requests: int) -> None:
+        """Set the maximum number of requests sent concurrently."""
+        self.uaclient.set_max_concurrent_requests(max_concurrent_requests)
+
     async def set_security_string(self, string: str) -> None:
         """
         Set SecureConnection mode.
--- a/asyncua/client/ua_client.py
+++ b/asyncua/client/ua_client.py
@@ -392,12 +392,15 @@ class UaClient:
     with code that used to call them on `UaClient` directly.
     """
 
+    _max_concurrent_requests = 100
+
     def __init__(self, timeout: float = 1.0) -> None:
         """
         :param timeout: Timout in seconds
         """
         self.logger = logging.getLogger(f"{__name__}.UaClient")
         self._timeout = timeout
+        self._request_semaphore = asyncio.Semaphore(self._max_concurrent_requests)
         self.security_policy: security_policies.SecurityPolicy = security_policies.SecurityPolicyNone()
         self.protocol: UASocketProtocol | None = None
         self._pre_request_hook: Callable[[], Awaitable[None]] | None = None
@@ -515,6 +518,11 @@ class UaClient:
     def set_security(self, policy: security_policies.SecurityPolicy) -> None:
         self.security_policy = policy
 
+    def set_max_concurrent_requests(self, max_concurrent_requests: int) -> None:
+        """Set the maximum number of requests sent concurrently."""
+        self._max_concurrent_requests = max_concurrent_requests
+        self._request_semaphore = asyncio.Semaphore(max_concurrent_requests)
+
     def _make_protocol(self) -> UASocketProtocol:
         self.protocol = UASocketProtocol(self._timeout, security_policy=self.security_policy)
         self.protocol.pre_request_hook = self._pre_request_hook
@@ -585,9 +593,10 @@ class UaClient:
     async def _send_request(
         self, request: Any, timeout: float | None = None, message_type: ua.MessageType = ua.MessageType.SecureMessage
     ) -> Buffer:
-        if self.protocol is None:
-            raise ConnectionError("Connection is not open")
-        return await self.protocol.send_request(request, timeout, message_type)
+        async with self._request_semaphore:
+            if self.protocol is None:
+                raise ConnectionError("Connection is not open")
+            return await self.protocol.send_request(request, timeout, message_type)
 
     # --- back-compat: properties that previously lived on UaClient ---
 
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -310,6 +310,9 @@ class Client:
     def set_locale(self, locale: Sequence[str]) -> None:
         self.aio_obj.set_locale(locale)
 
+    def set_max_concurrent_requests(self, max_concurrent_requests: int) -> None:
+        self.aio_obj.set_max_concurrent_requests(max_concurrent_requests)
+
     @syncmethod
     def load_private_key(
         self, path: str, password: str | bytes | None = None, extension: str | None = None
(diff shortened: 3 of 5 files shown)


asyncua/client/ua_client.py
```diff
@@ -392,12 +398,18 @@ class UaClient:
     with code that used to call them on `UaClient` directly.
     """
 
-    def __init__(self, timeout: float = 1.0) -> None:
+    def __init__(self, timeout: float = 1.0, max_concurrent_requests: int | None = None) -> None:
```
oroulet: we do not want even more constructor args. add that as class variable

asyncua/client/ua_client.py
```diff
@@ -26,9 +26,15 @@
...
 from ..ua.uaprotocol_auto import OpenSecureChannelResult
+from ._context import _supervisor_owns_requests
 from .ua_session import SessionState, UaSession
 
 
+def _validate_max_concurrent_requests(value: int | None) -> None:
+    if value is not None and (isinstance(value, bool) or not isinstance(value, int) or value <= 0):
+        raise ValueError("max_concurrent_requests must be a positive integer")
+
+
```
oroulet: no need for that. in python we need to trust very stupid users

asyncua/client/client.py
```diff
@@ -82,6 +73,7 @@ def __init__(
         auto_reconnect: bool = False,
         reconnect_max_delay: float = 30.0,
         reconnect_request_timeout: float = 60.0,
+        max_concurrent_requests: int | None = None,
```
oroulet: same her. we do not want more constructor args

oroulet:
That MR is too complex. Ask you AI generator to simplify it. Also no need for it to be optional, just set  a high default, maybe that will help things get simpler and then maybe it will be acceptable

sankalpsthakur:
Simplified and force-pushed as `59ce20a`. High default of 100 via class member + semaphore around transport send; removed constructor args, validation, and special cases.

oroulet:
thanks for th quick reaction but you went a bit too far here ;-) put the max number of connection as class member of Uaclient and as setter metod in client as a minimum.

sankalpsthakur:
Updated in 861feb5: class member default remains 100, and `Client.set_max_concurrent_requests()` updates the value and semaphore. Covered by the concurrency regression test.

asyncua/client/client.py
```diff
@@ -235,6 +235,11 @@ def set_locale(self, locale: Sequence[str]) -> None:
         """
         self._locale = locale
 
+    def set_max_concurrent_requests(self, max_concurrent_requests: int) -> None:
+        """Set the maximum number of requests sent concurrently."""
+        self.uaclient.max_concurrent_requests = max_concurrent_requests
+        self.uaclient._request_semaphore = asyncio.Semaphore(max_concurrent_requests)
```
oroulet: good work. 
but here we are accessing a private variable. so probably best to move that method to UaClient and make max_concurrent_requests private too. then make mehtod here call the one from Uaclient. sorry it went a bit back and forth but things are almost good here

sankalpsthakur: Updated in `558838c`: `UaClient` owns `_max_concurrent_requests` and the semaphore via `set_max_concurrent_requests()`; `Client` only delegates.

oroulet:
thank you. I think this is an improvment.

sankalpsthakur:
Review threads addressed on `558838c`; CI green. `Client.set_max_concurrent_requests()` only delegates to `UaClient`.

Issue #1792, opened 2025-03-06
Bad from_string/to_string result for NodeId


irreg:
**Describe the bug** <br />
There are the following problems with the processing of from_string/to_string of NodeId
- Missing percent encoding for NamespaceUri when from_string/to_string
- Identifier cannot be parsed correctly when it contains ";"
- Also, unlike the specifications, nsu= and srv= are written after the Identifier, so if the Identifier happens to contain a confusing string like ;nsu=, parsing will be impossible.

https://reference.opcfoundation.org/Core/Part6/v105/docs/5.1.12

**To Reproduce**<br />
```py
from asyncua import ua

before_node = ua.NodeId(Identifier="foo;nsu=bar")
print(before_node) # ExpandedNodeId(Identifier='foo;nsu=bar', NamespaceIndex=0, NodeIdType=<NodeIdType.String: 3>, NamespaceUri='http://example.com/q?=foo;bar', ServerIndex=0)
after_node = ua.NodeId.from_string(before_node.to_string())
print(after_node) # UaStringParsingError
```

**Expected behavior**<br />
Probably the fix will be something like this:

```py
from urllib.parse import unquote, quote

class NodeId:
    @staticmethod
    def _from_string(string):
        elements = string
        identifier = None
        namespace = 0
        ntype = None
        srv = None
        nsu = None
        while elements:
            if not elements:
                continue
            k, v = elements.split("=", 1)
            k = k.strip()
            
            # Split only if not a string
            elements = ""
            if k != "s": 
                if ";" in v:
                    v, elements = v.split(";", 1)
                v = v.strip()
            
            if k == "ns":
                namespace = int(v)
            elif k == "i":
                ntype = NodeIdType.Numeric
                identifier = int(v)
            elif k == "s":
                ntype = NodeIdType.String
                identifier = v
            elif k == "g":
                ntype = NodeIdType.Guid
                identifier = uuid.UUID(f"urn:uuid:{v}")
            elif k == "b":
                ntype = NodeIdType.ByteString
                identifier = bytes(v, 'utf-8')
            elif k == "srv":
                srv = int(v)
            elif k == "nsu":
                nsu = unquote(v) # Decode parcent encoded string
        if identifier is None:
            raise UaStringParsingError(f"Could not find identifier in string: {string}")
        if nsu is not None or srv is not None:
            return ExpandedNodeId(identifier, namespace, ntype, NamespaceUri=nsu, ServerIndex=srv)
[...]

PR #2000, opened 2026-07-22
Fix _to_nodeid for integer identifiers > 255

### LINKED ISSUE #1831: _to_nodeid() fails with int arguments > 255
**Describe the bug** <br />
Calls to _to_node_id() fail with int arguments > 255.
A typical example is a call like
`self.get_referenced_nodes()`
where the type of the reference is provided in parameter refs as integer argument.
Meanwhile a lot of standard OPC UA references have integer IDs >> 255 (see example):

**To Reproduce**<br />
The example shows a search for references of type HasDictionaryEntry, standard reference type of OPC UA.
It fails because the integer id of the type is  HasDictionaryEntry = 17597.<br />
`                nodes = await self.get_referenced_nodes(refs = ua.ObjectIds.HasDictionaryEntry, direction = ua.BrowseDirection.Forward, nodeclassmask = ua.NodeClass.Object)
`
The root cause of the failure is, that the current implementation of the function _to_nodeid(), which is utilized when issuing the call in the example, only supports TwoByteNodeIds for input arguments of type int.<br />

**Expected behavior**<br />
The implementation of  _to_nodeid() should check the numeric range of the input argument and return an appropriate nodeid object, e.g. as depicted in the following code snippet with additional range check:<br />
`            
def _to_nodeid(nodeid: Union["Node", ua.NodeId, str, int]) -> ua.NodeId:
    if isinstance(nodeid, int):
            if nodeid <= 255:
                return ua.TwoByteNodeId(nodeid)
            elif nodeid <= 65535:
                return ua.FourByteNodeId(nodeid)
            else:
                return ua.NumericNodeId(nodeid)
    if isinstance(nodeid, Node):
        return nodeid.nodeid
    if isinstance(nodeid, ua.NodeId):
        return nodeid
    if isinstance(nodeid, str):
        return ua.NodeId.from_string(nodeid)
    raise ua.UaError(f"Could not resolve '{nodeid}' to a type id")
`

**Version**<br />
Python-Version: 3.13.3 <br />
opcua-asyncio Version: master branch, 1.1.6

sankalpsthakur:
## Summary

Fixes #1831

`_to_nodeid()` always built a `TwoByteNodeId` for `int` inputs. That encoding only allows identifiers `0..255`, so standard reference types such as `HasDictionaryEntry` (`17597`) failed with:

```
ValueError: TwoByteNodeId cannot have identifier > 255
```

when used with `get_referenced_nodes(refs=...)`, `add_reference`, etc.

This breaks dictionary and companion-spec browsing on real servers (Siemens,
Beckhoff TwinCAT, Prosys exports): walking nodes by standard reference type
ObjectId is a routine client pattern and should not require manually
constructing `FourByteNodeId`/`NumericNodeId` wrappers.

Related: #1792 (separate NodeId `from_string`/`to_string` parsing bugs in
the same `uatypes.py` module — not addressed here).

## Changes

- Route integer node ids through `ua.NodeId()`, which selects compact encoding by range
- Fix off-by-one in `NodeId` auto-encoding so bounds match the subclass validators:
  - TwoByte: Identifier `0..255` (was `< 255`)
  - FourByte: Identifier `0..65535`, NamespaceIndex `0..255` (was NamespaceIndex `< 255`)
- Unit regressions for small / mid / large ranges, `ObjectIds.HasDictionaryEntry`, and encoding bounds

## Repro

```python
nodes = await node.get_referenced_nodes(
    refs=ua.ObjectIds.HasDictionaryEntry,
    direction=ua.BrowseDirection.Forward,
)
# ValueError: TwoByteNodeId cannot have identifier > 255
```

## Tests

```bash
pytest tests/test_unit.py::test_to_nodeid_int_ranges tests/test_unit.py::test_nodeid_encoding_range_bounds -q
# 2 passed
```

### DIFF
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -42,7 +42,14 @@ def _check_results(results: list[ua.StatusCode], reqlen: int = 1) -> None:
 
 def _to_nodeid(nodeid: Node | ua.NodeId | str | int) -> ua.NodeId:
     if isinstance(nodeid, int):
-        return ua.TwoByteNodeId(nodeid)
+        # Pick the compact NodeId encoding that can hold the identifier.
+        # TwoByteNodeId only supports 0..255; larger standard reference type
+        # ids (e.g. HasDictionaryEntry = 17597) need FourByte/Numeric.
+        if nodeid <= 255:
+            return ua.TwoByteNodeId(nodeid)
+        if nodeid <= 65535:
+            return ua.FourByteNodeId(nodeid)
+        return ua.NumericNodeId(nodeid)
     if isinstance(nodeid, Node):
         return nodeid.nodeid
     if isinstance(nodeid, ua.NodeId):
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -42,9 +42,6 @@ def _check_results(results: list[ua.StatusCode], reqlen: int = 1) -> None:
 
 def _to_nodeid(nodeid: Node | ua.NodeId | str | int) -> ua.NodeId:
     if isinstance(nodeid, int):
-        # Use NodeId's range-based encoding (TwoByte/FourByte/Numeric).
-        # Plain TwoByteNodeId only allows 0..255; standard reference type ids
-        # such as HasDictionaryEntry (17597) need a wider form (GH-1831).
         return ua.NodeId(nodeid)
     if isinstance(nodeid, Node):
         return nodeid.nodeid
--- a/asyncua/common/node.py
+++ b/asyncua/common/node.py
@@ -42,14 +42,10 @@ def _check_results(results: list[ua.StatusCode], reqlen: int = 1) -> None:
 
 def _to_nodeid(nodeid: Node | ua.NodeId | str | int) -> ua.NodeId:
     if isinstance(nodeid, int):
-        # Pick the compact NodeId encoding that can hold the identifier.
-        # TwoByteNodeId only supports 0..255; larger standard reference type
-        # ids (e.g. HasDictionaryEntry = 17597) need FourByte/Numeric.
-        if nodeid <= 255:
-            return ua.TwoByteNodeId(nodeid)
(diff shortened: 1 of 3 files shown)


asyncua/common/node.py
```diff
@@ -42,7 +42,10 @@ def _check_results(results: list[ua.StatusCode], reqlen: int = 1) -> None:
 
 def _to_nodeid(nodeid: Node | ua.NodeId | str | int) -> ua.NodeId:
     if isinstance(nodeid, int):
-        return ua.TwoByteNodeId(nodeid)
+        # Use NodeId's range-based encoding (TwoByte/FourByte/Numeric).
+        # Plain TwoByteNodeId only allows 0..255; standard reference type ids
+        # such as HasDictionaryEntry (17597) need a wider form (GH-1831).
```
oroulet: comment is useless here

asyncua/ua/uatypes.py
```diff
@@ -472,9 +472,12 @@ class NodeId:
     def __post_init__(self) -> None:
         if self.NodeIdType is None:
             if isinstance(self.Identifier, int):
-                if self.Identifier < 255 and self.NamespaceIndex == 0:
+                # Match TwoByteNodeId / FourByteNodeId validation bounds:
+                # TwoByte: Identifier 0..255, NamespaceIndex == 0
+                # FourByte: Identifier 0..65535, NamespaceIndex 0..255
```
oroulet: same here. We cannot have 4 lines comments for every line of code. that 255 max issue is obvious

asyncua/common/node.py
sankalpsthakur: Removed in 426c4c67 — thanks.

asyncua/ua/uatypes.py
sankalpsthakur: Removed in 426c4c67 — thanks.

PR #2010, opened 2026-07-30
Keep sync Client ThreadLoop alive across disconnects

### LINKED ISSUE #1364: sync.Clinet is not reusable

miki5799:
Fixes #1364.
Supersedes #1503.

`sync.Client.disconnect()` currently also stops an internally created
`ThreadLoop`, so connecting the same Client again raises
`ThreadLoopNotRunning`.

This PR separates the connection lifecycle from the sync-wrapper lifecycle:

- `disconnect()` tears down the OPC UA connection but keeps the ThreadLoop alive
- `close()` disconnects and stops an internally owned ThreadLoop
- caller-provided ThreadLoops remain owned by the caller
- the Client context manager performs final cleanup through `close()`

Keeping the same ThreadLoop also preserves the loop references held by existing
`SyncNode` and other sync wrapper objects.

### DIFF
--- a/asyncua/sync.py
+++ b/asyncua/sync.py
@@ -18,6 +18,7 @@ from typing import Any, Literal, TypeVar, overload
 from cryptography import x509
 
 from asyncua import client, common, server, ua
+from asyncua.client.ua_client import UaClientState
 from asyncua.common import node, shortcuts, subscription, type_dictionary_builder, xmlexporter
 from asyncua.common.events import Event
 from asyncua.crypto import uacrypto
@@ -239,6 +240,9 @@ class Client:
     the sync client has one extra parameter: sync_wrapper_timeout.
     if no ThreadLoop is provided this timeout is used to define how long the sync wrapper
     waits for an async call to return. defualt is 120s and hopefully should fit most applications
+
+    Disconnecting keeps the ThreadLoop alive for later reconnects. Call close
+    when the client is no longer needed.
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


miki5799:
A backward-compatible 2.x alternative would be to add a keyword-only option:

```python
Client(..., stop_tloop_on_disconnect=True)
```

With the current behavior as the default, reusable clients would opt in with
`stop_tloop_on_disconnect=False` and call `close()` for final cleanup.

I chose not to include that option in the initial implementation because it
would add a transitional lifecycle flag to the public API. The direct
`disconnect()` / `close()` separation is smaller and represents the intended
long-term API more clearly.

The trade-off is that applications which manually use a final `disconnect()`
must migrate to `close()`; otherwise the internally owned non-daemon
ThreadLoop remains active. If preserving the current 2.x cleanup behavior is
preferred, I can switch the PR to the opt-in variant.

oroulet:
What is that supposed to fix? it does change API but doe snot offer any advantage. if you do something special, you can already just make your own threadlopp

miki5799:
> What is that supposed to fix? it does change API but doe snot offer any advantage. if you do something special, you can already just make your own threadlopp

Fair point, I was relying on an old version and haven't really considered that to be a solution. I guess the related issues are thereby resolved ? I'll close this PR

PR #2013, opened 2026-08-07
Implement write_values method for node value updates

fabian222005:
Added a new method to write values for multiple nodes, handling DataValue conversion.


oroulet:
even liting fails...

fabian222005:
sorry, I forgot to close the PR, I'm still working. Thanks

PR #2016, opened 2026-08-11
Fully migrate to `cryptography` from `pyOpenSSL`

still-wondering:
# Fully migrate to `cryptography` from `pyOpenSSL`
Replaces remaining usages of `pyOpenSSL` library for trust store handling with the corresponding functionality from `cryptography` library.

The replacement is not absolutely identical as `cryptography` takes a bit different approach to enforcing some constraints, but the changes should be practically the same for most of users.

## Key changes
-  `X509Store` -> `Store`,  `X509StoreContext` -> `ClientVerifier`(created by the `PolicyBuilder`)
- `cryptography`'s verifier checks for basic integrity, signatures, and the _date range_, therefore the manual check is not needed anymore
- `cryptography`'s policy builder by default enforces some extension checks according to the CA/B Forum guidelines. For the sake of compatibility those checks are disabled (see occurrences of `ExtensionPolicy.permit_all()`). But policy builder unconditionally enforces CA certificates to include the `basicConstraint` extension (with any criticality). It seems to me that if CA certificate contains that extension, then it must have the `ca` field set to true, therefore this behavior is also enforced
- policy builder is left as a public field on the trust store object, so that the caller can configure any extra checks on it if they want to do so
- `pyOpenSSL` library dependency is removed
- small modifications on how the certificates and CRLs are read

## Other notes
- as the policy builder can perform all those checks (when configured to), it would be nice to use them instead of the checks in the `CertificateValidator` class, but for now there is no way to reliably programatically distinguish the reason of the failed verification (verifier raises VerificationException for all the errors)
- it would be nice to load the certificates and CRLs here concurrently with the `TaskGroup` since python 3.11

### DIFF
--- a/asyncua/crypto/truststore.py
+++ b/asyncua/crypto/truststore.py
@@ -3,46 +3,57 @@ Functionality for checking a certificate based on:
 - trusted (ca) certificates
 - crl
 
-Use of cryptography module is prefered, but  doesn't provide functionality for truststores yet, so for some we rely on using pyOpenSSL.
-
 """
 
 import logging
 import re
-from datetime import datetime, timezone
 from pathlib import Path
 
+import anyio
 from cryptography import x509
-from OpenSSL import crypto
 
 from asyncua.crypto.uacrypto import get_content, load_certificate
 
 _logger = logging.getLogger("asyncuagds.validate")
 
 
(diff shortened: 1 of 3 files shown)


oroulet:
would have been great to split cosmetic changes from real changes.. but OK

still-wondering:
i see, will try to keep them separate in the future

Issue #1964, opened 2026-05-07
[Security] Chunk Buffer Not Released After max_chunk_count Rejection (CWE-401 Memory Leak)


Giancannella:
**Describe the bug**
[Security] When the server correctly rejects a message that exceeds max_chunk_count,  the ~100 MB buffer accumulated from the preceding 1601 chunks is NOT immediately  released from heap memory. Verified with top/htop: RAM remains elevated after the rejection and is only eventually reclaimed by the Python GC with no deterministic timing.

This is distinct from the FindServers/GetEndpoints exhaustion issues (filed separately):

This issue (post-rejection leak):

An attacker can rapidly cycle connections: send max_chunk_count-1 chunks → trigger  rejection → close → repeat. Residual buffers accumulate faster than the GC collects them, causing progressive memory exhaustion even with short-lived connections.

**To Reproduce**
1. Start a default asyncua server (e.g. `uaserver --populate`)
2. In a loop, open a connection, send exactly 1602 chunks (to trigger rejection), close it
3. Monitor RAM with top
4. Observe that RAM does not immediately drop to baseline after each rejection

**Expected behavior**
Upon rejecting a message that exceeds transport limits, the server should explicitly clear/free the accumulated chunk buffer before raising the exception:

**Version**
Python-Version: 3.11
opcua-asyncio Version: latest master (tested with asyncua 1.1.x)
OS: Ubuntu 24.04

Poseidonas:
I tried to reproduce this and could not. Measured with `tracemalloc`, driving `SecureConnection._receive()` with a distinct payload per chunk: 100.6 MB at the peak, and `+0.0 MB` immediately after the rejection — without calling `gc.collect()`. `self._incoming_parts = []` drops the last reference to the chunks and CPython frees them at that point; nothing else holds them.

What `top` shows afterwards is most likely the allocator keeping the pages rather than returning them to the OS, which looks the same from outside the process but is not a leak.

There is a real effect nearby, though: the connection was kept open after the rejection, so the same connection could accumulate and be rejected repeatedly without reconnecting — which is close to the cycling described here, minus the new connections. That part is in #2026.

Happy to look again at a case where the memory genuinely stays held.

Giancannella:
Thanks for the detailed investigation on #1964. The tracemalloc results are convincing — what I observed in top was indeed the allocator retaining pages rather than a true memory leak. I'm closing this issue as it doesn't represent a separate vulnerability.

Issue #1963, opened 2026-05-07
[Security] Unauthenticated Memory Exhaustion via GetEndpointsRequest Chunked Transport (DoS)


Giancannella:
**Describe the bug**
[Security] Same class of vulnerability as FindServersRequest (filed separately), but exploiting the GetEndpoints Discovery Service (NodeId 428). An unauthenticated attacker sends an oversized GetEndpointsRequest with an inflated endpointUrl field via chunked transport, never sending the final chunk 'F', causing the server to hold ~100 MB RAM per connection indefinitely.

This is a distinct issue from the FindServers report:

FindServersRequest: NodeId 422,  Vulnerable field: serverUris[0] (UA_String array)
GetEndpointsRequest: NodeId 428, Vulnerable field: endpointUrl (single UA_String) 

Both Discovery Services are available before any session or SecureChannel is established, regardless of SecurityMode (None, Sign, Sign+Encrypt).

**To Reproduce**
1. Start a default asyncua server (e.g. `uaserver --populate`)
2. Open N parallel TCP connections to port 4840
3. For each connection, complete the HEL/ACK handshake
4. Send a GetEndpointsRequest (NodeId 428) split into 1601 chunks of 65535 bytes, with endpointUrl.length declared as a large value (~100 MB)
5. Never send the final chunk (type 'F') — hold the connection open with time.sleep()
6. Monitor RAM with `watch -n 1 'ps aux | grep uaserver'`

```python
# Attack parameters used in testing:
MAX_CHUNK_SIZE  = 65535
MAX_CHUNK_COUNT = 1601
RAM_PER_CONN    = 100 * 1024 * 1024  # ~100 MB
# 50 connections → ~5 GB, 160 connections → OOM on 16 GB machine
```

**Expected behavior**
Server should enforce a maximum message size well below 100 MB and immediately disconnect any client exceeding it. GetEndpoints should not allow unauthenticated clients to allocate unbounded memory.

**Screenshots**

<img width="1244" height="252" alt="Image" src="https://github.com/user-attachments/assets/d42f7209-119b-4d5b-a8c0-db124b4d7583" />

**Version**
Python-Version: 3.11
opcua-asyncio Version: latest master (tested with asyncua 1.1.x)
OS: Ubuntu 24.04

Poseidonas:
[...]

Issue #1962, opened 2026-05-07
[Security] Unauthenticated Memory Exhaustion via FindServersRequest Chunked Transport (DoS)


Giancannella:
**Describe the bug**
[Security] An unauthenticated remote attacker can exhaust server memory via the FindServersRequest Discovery Service (NodeId 422) by sending up to 1601 intermediate chunks (type 'C') via chunked OPC UA Binary transport without ever sending the final  chunk (type 'F'). The server allocates up to ~100 MB RAM per connection before any check is performed. With the default max_connections=1000, this allows ~100 GB of theoretical RAM exhaustion without any authentication.

**To Reproduce**
1. Start a default asyncua server (e.g. `uaserver --populate`)
2. Open N parallel TCP connections to port 4840
3. For each connection, complete the HEL/ACK handshake
4. Send a FindServersRequest (NodeId 422) split into exactly 1601 chunks of 65535 bytes, with serverUris[0].length declared as a large value (e.g. 100*1024*1024)
5. Never send the final chunk (type 'F') — hold the connection open with time.sleep()
6. Monitor server RAM with top

```python
# Attack parameters used in testing:
MAX_CHUNK_SIZE  = 65535        # bytes per chunk
MAX_CHUNK_COUNT = 1601         # max before rejection
RAM_PER_CONN    = 100 * 1024 * 1024  # ~100 MB per connection
# With 50 open connections → ~5 GB RAM allocated, server becomes unresponsive
```

**Expected behavior**
The server should enforce a reasonable maximum message size (e.g. 2 MB) and reject or disconnect after a short timeout any client that does not complete a message. No unauthenticated client should be able to allocate more than a few MB on the server.

**Screenshots**

<img width="1244" height="252" alt="Image" src="https://github.com/user-attachments/assets/6713b6ce-dd1d-4256-aba9-943906afbed5" />

**Version**
Python-Version: 3.11
opcua-asyncio Version: latest master (tested with asyncua 1.1.x)
OS: Ubuntu 24.04

Poseidonas:
Reproduced on master: with the default limits a peer that never sends the final chunk gets 1601 intermediate chunks accepted and the server then holds 99.9 MB, before any session exists.

[...]

PR #2026, opened 2026-08-25
Bound a message while it is being reassembled, and close the connection after a transport error

Poseidonas:
Towards #1962 and #1963. Both describe the same transport-layer behaviour, so this is one change rather than two, and it also covers a third effect neither report mentions.

**What a peer can hold today**

Measured on master with the default limits: 1601 intermediate chunks are accepted for a message that is never completed, and `_incoming_parts` then holds 99.9 MB, before any session exists.

`TransportLimits.is_msg_size_within_limit()` exists but is never called, so nothing bounds a message while it is being reassembled — only the chunk count is checked, and because the default count is derived from the default size (`(100 MB // 65535) + 1`) the two coincide. Lowering `max_message_size` on its own therefore changes nothing: with it set to 2 MB the server still accumulates 99.9 MB. With this change the same setting accumulates 2.1 MB.

I have not changed the defaults — with them the chunk count still fires first, and picking a different default message size is your call.

**Repeating the same message on one connection**

Part 6, 7.1.3 has the server send an Error Message and close the transport connection. `UaProcessor.process()` sent the error and returned `True`, keeping the connection open, so the same connection could accumulate and be rejected over and over — measured at three rounds of 100 MB each without reconnecting.

It also left the trailing chunks of the refused message in the inbound queue, where they were parsed as if they were a new message:

```
Exception raised while processing message from client
  ...
  File "asyncua/ua/ua_binary.py", line 551, in nodeid_from_binary
ValueError: 33 is not a valid NodeIdType
```

Closing the transport alone did not stop that, since whatever `data_received` had already queued is processed after the close — so the message loop now stops when a message asks for the connection to close. That applies to the `BadUserAccessDenied` path as well, which had the same gap.

**What this does not change**

A conforming client already calls `disconnect_socket()` when it receives an Error Message, so nothing changes for it. Measured with a real client and an oversized write: `BadRequestTooLarge` is raised either way, the client ends up `DISCONNECTED` either way, and the context manager exits cleanly either way. Only a peer which ignores the error is affected.

**On #1964**

I could not reproduce it. With `tracemalloc` and a distinct payload per chunk: 100.6 MB at the peak, `+0.0 MB` immediately after the rejection without calling `gc.collect()` — `self._incoming_parts = []` drops the last reference and CPython frees it there and then. What `top` shows afterwards is the allocator not returning pages to the OS, which is a different thing. Happy to be shown a case where it does leak.

**Tests**

Five for the size limit (a message within the limit still arrives, `max_message_size=0` stays unlimited, the running total resets between messages and after an abort), and two for the close (no parse errors after the incident, and `process()` reports the connection should close after writing the error). Each fails on master. `mypy asyncua tests` and `ruff check .` are clean, and the suite is unchanged apart from those.


asyncua/common/connection.py
```diff
@@ -489,17 +490,29 @@ def receive_from_header_and_body(self, header: ua.Header, body: Buffer) -> Any:
...
+            self._drop_incoming_parts()
             _logger.error(
                 "Message size: %s is > chunk max size: %s", msg.MessageHeader.packet_size, self._limits.max_recv_buffer
             )
             raise ua.UaStatusCodeError(ua.StatusCodes.BadRequestTooLarge)
         self._check_incoming_chunk(msg)
+        # Check the size the message would reach before holding on to the body:
+        # a peer which never sends the final chunk otherwise keeps everything it
+        # sent so far, and the chunk-count limit alone only stops that once a
+        # full max_message_size has been accumulated.
```
oroulet: useless comment

asyncua/server/binary_server_asyncio.py
```diff
@@ -157,20 +157,26 @@ async def _process_received_message_loop(self) -> None:
                 # Connection was closed, end task
                 break
             try:
-                await self._process_one_msg(header, buf)
+                if not await self._process_one_msg(header, buf):
+                    # The connection is being closed; whatever else is queued
+                    # belongs to the message we just refused and would be parsed
+                    # as if it were a new one.
+                    break
```
oroulet, line 161: no need for a comment

asyncua/server/uaprocessor.py
```diff
@@ -140,7 +140,11 @@ async def process(self, header, body):
             err = ua.ErrorMessage(ua.StatusCode(e.code), str(e))
             data = uatcp_to_binary(ua.MessageType.Error, err)
             self._transport.write(data)
-            return True
+            # Part 6, 7.1.3: after an Error Message the transport connection is
+            # closed. Keeping it open lets a peer repeat the offending message
+            # on the same connection, and a conforming client has closed its
+            # own side on receiving the error anyway.
```
oroulet: that one is fine

Poseidonas:
Done, comments removed.

oroulet:
please merge the 2 last commits where they belong

Poseidonas:
Done, squashed into the two commits.

Issue #1972, opened 2026-05-14
OpenTelemetry tracing support — interest in native integration?


luke6Lh43:
Hi Team,

I've been working on OpenTelemetry instrumentation for asyncua and have a working prototype both as a manual approach (Medium article) and as a draft contrib package for OTel. I brought it to the OTel Python SIG, and their suggestion was to first check whether the asyncua team would be interested in native OTel integration within the library itself rather than as an external instrumentation package.

Native integration would mean:

OpenTelemetry as an optional dependency (asyncua[otel])
Spans emitted by asyncua directly on key operations (Node.read_value, Client.connect, etc.)
Users get tracing with minimal setup, no monkey-patching needed
Long-term sustainability — the library and its instrumentation evolve together

A few questions for the maintainers:

Is there interest in this direction in principle?
If yes, would you prefer it as a built-in feature or as an opt-in plugin/extension point?
Are there constraints (dependencies, supported Python versions, library size)?

[GH Issue in OTEL Python SIG](https://github.com/open-telemetry/opentelemetry-python-contrib/issues/4491)

[Medium article](https://medium.com/@lukasz.ciukaj/bringing-opentelemetry-to-opc-ua-manual-instrumentation-of-an-industrial-protocol-2cfa03475164)

Thanks for your work on asyncua — happy to discuss any way that fits the project.

oroulet:
maybe... how invasive is that. and how can it help?

kohtala:
As a user I'd be interested in connection, session, subscription gauges and some counters for what requests are received and their responses both for client and server.

I don't expect this would be much invasive. What some instrumentation I've done, it's been quite minimal API calls with insignificant cost for those who don't add some SDK to actually make the data available somewhere.

luke6Lh43:
Thanks for the feedback! I agree that metrics such as connection/session status, subscriptions, request counts, and response statistics would provide a lot of operational value alongside tracing.

From the OpenTelemetry perspective, I see traces and metrics as complementary. Traces help users understand the execution path and latency of individual OPC UA operations, while metrics provide the long-term operational health of clients and servers. If the integration remains optional and lightweight, I think both could be implemented with minimal overhead for users who don't enable OpenTelemetry.

Poseidonas:
[...]

PR #2027, opened 2026-08-25
Report requests, connection state and subscription lifecycle to an optional observer

Poseidonas:
Towards #1972.

The places asked for in that issue are all points the client already funnels through, so watching them needs no new plumbing: requests go through `UaClient._send_request`, connection state through `_set_state`, and the subscription lifecycle through `init()` / `recreate()` / `publish_callback`. `UaClient.observer` is unset by default and is offered each of:

```python
on_request(request_type, duration, error)
on_state_change(state)
on_subscription_event(event, subscription_id)   # created / recreated / deleted
on_notification(subscription_id, item_count)
```

Which covers what @kohtala asked for — connection and session state, subscription counts, request counts with their latency and outcome — and lets @luke6Lh43's OpenTelemetry work live outside the library rather than monkey-patching it.

**Nothing is imported, and nothing is measured when nobody is watching.** Each method is looked up before it is called, so an observer implements only what it needs and a partial one is not an error. An observer that raises is logged, not allowed to break the client. The clock is only read once an observer is set — a test pins that. Cost of a request with no observer, measured against master over 20k calls: 5.26 vs 5.31 µs.

Two things I got wrong first and fixed, in case they matter for review:

- the request callback ran inside `_request_semaphore`, so a slow observer held a concurrency slot and serialized requests behind it. It now runs outside, and a test asserts the slot is free while the observer runs.
- `recreate()` reported both `created` (from `init()`) and `recreated`, which would double-count on every reconnect. It reports one event now.

Twelve tests; each of the last two fails without its fix. `ruff check asyncua tests` and `mypy asyncua tests` are clean, and the suite is unchanged.

This is a starting point rather than a finished feature — the server side has the same kind of single choke points and could follow the same shape if this one is right. Happy to adjust names, arguments, or scope.


luke6Lh43:
Thanks @Poseidonas for putting this together and moving the idea forward! Really appreciate you taking this on. Please let me know how I can help with the PR or with any OpenTelemetry-specific aspects as this progresses.

kohtala:
In case the asyncua goes this way instead of directly using the opentelemetry-api, here my thought on this suggestion. Please take them as not requesting or demanding anything, but sharing thoughts I had.

I'd consider changing the implementation for a missing observer into providing an _ignorer_ while there is no-one observing. That is an observer that has empty methods that do nothing. Then at the instrumented places it can always expect an observer to be there. I don't see passing the parameters for nothing taking that much time that we'd need to try to avoid the call. 

I'd also consider leaving the exception handling be the responsibility of the observer, because some observers could be safe to trust to not throw exceptions. 

Similarly I'd consider leaving the timing of requests to the observer. The observer could provide a [context manager](https://docs.python.org/3/library/stdtypes.html#typecontextmanager) object and have the request run in the observer context. That would be easier to turn into tracing as well.

I also would consider leaving it to be the responsibility of the observer to be fast. No need to consider a slow observer. A typical implementation would just queue the observation for another thread to move forward and hence never block. Observations only would be lost if there are problems and the queue gets full.

The opentelemetry-api is much the same empty shell unless there is additional code provided to move the information forward. It does have a cost in the parameters that this kind of wrapper might help to reduce, if some cheaper parameters were added to offer access to dig out more information on the context when needed. The #2029 gives a comparison on the approaches.

Poseidonas:
Thank you both. The context-manager shape is better than what I have — it makes the request timing the observer's concern and turns into a span naturally — and a no-op observer instead of `getattr` checks is simpler too. I'm happy to rework #2027 along those lines, or to stand aside if the maintainers prefer the native route in #2029; either way the choke points are the same, so nothing is wasted. I'll wait for a word on direction before changing anything.

luke6Lh43:
From the OpenTelemetry perspective, I agree that the context-manager approach is a nice fit, especially for tracing, since the observer can naturally own the span lifecycle, timing, and error handling.

For me, the most important part is that whichever direction we take (#2027 observer hooks or native OTel API in #2029), the instrumentation points expose enough context to build meaningful OPC UA telemetry without having to reach back into asyncua internals or monkey-patch anything.

I’d also suggest keeping the hooks extensible enough to add OPC UA-specific attributes later as we better define the telemetry model and semantic conventions.

Happy to help with the OpenTelemetry side once the maintainers decide which direction they prefer.

oroulet:
I didnot like these getattibutes call everywhree. Claude proposed these changes https://github.com/FreeOpcUa/opcua-asyncio/pull/2041/changes

let me know what you think

oroulet:
also included all comments from @kohtala .
I think the observer api is nice. the other MR is just a proposition. let me know if that good enough as-is or if @Poseidonas wants to take over again

Poseidonas:
Thanks for taking the time to help and you're right about the getattr calls. Honestly, a few things in your version are better than what I had — the concrete `Observer` with a `NULL_OBSERVER` default reads much cleaner at the call sites, exposing it as a property on `AbstractSession` beats the way I was reaching into `_client`, and `observe_request` as a context manager is the natural fit for a span. I like the `init()`/`_create()` split too, and marking the subscription deleted before notifying.

I built both branches and ran the suites — yours is green (14), and the reconnect/HA/subscription tests pass as well. The one thing I'd want to bring over from my side is keeping the observer from being able to take the client down. These hooks are third-party tracing code running inline on the connection paths, so I tried them with an observer that throws, and in a few spots the exception carries straight through. When `on_state_change` throws it happens before the state listeners run and nothing wraps it, so the reconnect queue and the HA `refresh_handles` listener never see the transition and the client just stops reconnecting. When it throws as `observe_request` exits, a request that actually went through comes back to the caller as a failure. And when it throws in `on_subscription_event` during create, the subscription already exists on the server but the caller never gets the handle to delete it. On top of that, the branch drops the `try/except` that was already around the state listeners in `_set_state` — that one's separate from the observer, but the reconnect path leans on it, so I'd leave it in.

Two small things I noticed: there's an 11 MB `binary_address_space.pickle` that looks like it slipped in by accident, and neither of us tests cancellation on the request path.

If it's useful, I'm happy to take it from here — keep your shape, wire the guards back in, drop the pickle, and add the missing tests (cancellation, a throwing `on_notification`/`on_subscription_event`, zero-item notifications). Glad to credit both of us and @kohtala. Sound good, or would you rather keep it on your branch?

oroulet:
@Poseidonas all your comments are correct. I made that MR with AI as a very quick proposition, sorry. You can take over from now on

Poseidonas:
Pushed the reworked version. I kept your shape — the `Observer` base with `NULL_OBSERVER`, the `SubscriptionEvent` enum, the session `observer` property, `observe_request` as a context manager, and the `init()`/`_create()` split with the subscription marked deleted before it reports.

What I brought back is the fault isolation: every hook is wrapped so a raising observer is logged, never able to take the client down, and I left the existing `try/except` around the state listeners in place. `observe_request` is now concrete on the base — it times the call and hands the result to `on_request(request_type, duration, error)`, so a simple consumer overrides only `on_request` and gets the duration for free, while a span consumer still overrides `observe_request`. `NULL_OBSERVER` returns a `nullcontext`, so nothing is measured when no one is watching. `on_state_change` reports `state.value`, `on_subscription_event` reports the enum. Pickle's gone.

Twenty tests (a raising observer on each hook, request cancellation, a zero-item notification, delete-before-report); ruff and mypy clean, and the reconnect/HA/subscription suites still pass. Thanks again — your review and @kohtala's shaped this. Happy to adjust anything.

asyncua/client/ua_client.py
```diff
@@ -432,11 +434,18 @@ def clear_disconnect_request(self) -> None:
         """Clear the shutdown flag ahead of a fresh connect()."""
         self._disconnect_requested = False
 
+    def _observe(self, call: Callable[[], None]) -> None:
```
oroulet, line 437: the name of that parameter is shadowing some builtin or at least common method. Can you rename it?

asyncua/client/ua_client.py
```diff
@@ -599,10 +608,11 @@ async def close_secure_channel(self) -> None:
     async def _send_request(
         self, request: Any, timeout: float | None = None, message_type: ua.MessageType = ua.MessageType.SecureMessage
     ) -> Buffer:
-        async with self._request_semaphore:
-            if self.protocol is None:
-                raise ConnectionError("Connection is not open")
-            return await self.protocol.send_request(request, timeout, message_type)
+        with self.observer.observe_request(request):
```
oroulet, line 611: very nice API now ;-)

oroulet:
@Poseidonas fix one small comment and then we merge. 
@luke6Lh43 @kohtala it has everything you need for open telemetry?

Poseidonas:
Done — renamed `call` to `hook`. A couple of related things were bugging me while I was in there, so I fixed them too and figured you'd rather see them now than in a follow-up.

I put the state back to the `UaClientState` enum instead of the string I had — it lines up with the state listeners sitting right next to it and with `ha_client`, which already compares against `UaClientState`, and honestly it was the enum in your version to begin with. I also flipped `on_subscription_event` so `subscription_id` comes first, to match `on_notification`. And I renamed the publish counter from `delivered` to `produced` (and `item_count` to `event_count`) — `_deliver` can drop on overflow, so "delivered" was claiming more than actually happened.

ruff, mypy and the observer/reconnect/HA/subscription suites are all green. Should be good to go.

luke6Lh43:
Yes, this gives us everything we need.

`observe_request` as a context manager maps directly to a trace span — the adapter overrides it, opens a span on enter, records the error and closes it on exit. `on_request` feeds a request duration histogram, `on_state_change` a connection state gauge, `on_subscription_even`t an active-subscriptions counter, and `on_notification` a received-items counter. That's the full signal set for meaningful OPC UA telemetry.

Once this merges I'll put together a draft `opentelemetry-instrumentation-asyncua` package implementing the Observer interface and share it here for feedback — the one open question will be settling on an `opcua.*` attribute namespace, since OTel doesn't have official semconv for OPC UA yet.

PR #2029, opened 2026-08-26
Implement simple opentelemetry tracing and connection metric for UaClient

kohtala:
This is a simple start of native opentelemetry instrumentation for comparison with #2027 to provide for #1972.


oroulet:
I find these additions quite intrusive, but now I get the point of open telemetry. Any ideas to improve that?

PR #2031, opened 2026-08-27
add method history_update

AndreasHeine:
https://reference.opcfoundation.org/specs/OPC-10000-4/5.11.5


AndreasHeine:
@oroulet i stumbled over this in history.py:

```py
    def update_history(self, params: ua.HistoryUpdateParameters) -> list[ua.HistoryUpdateResult]:
        """
        Update history for a node
        This is the part AttributeService, but implemented as its own service
        since it requires more logic than other attribute service methods
        """
        results: list[ua.HistoryUpdateResult] = []
        for _ in params.HistoryUpdateDetails:
            result = ua.HistoryUpdateResult()
            # we do not accept to rewrite history
            result.StatusCode = ua.StatusCode(ua.StatusCodes.BadNotWritable)
            results.append(result)
        return results
```

why do we dont accept to rewrite history?

PR #2038, opened 2026-09-07
fix(server): await node deletion callbacks

### LINKED ISSUE #1771: Coroutine not awaited while deleting node
**Describe the bug**
When calling `asyncua_server.delete_nodes([self.node], recursive=True)` there appears message in terminal:
```
.local/lib/python3.12/site-packages/asyncua/server/address_space.py:428: RuntimeWarning: coroutine 'MonitoredItemService.datachange_callback' was never awaited
  callback(handle, None, ua.StatusCode(ua.StatusCodes.BadNodeIdUnknown))
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

It seems there are some sync/async issues, maybe some functionality doesn't work properly.

**To Reproduce**
Call `asyncua_server.delete_nodes([self.node], recursive=True)`

**Expected behavior**
No warnings

**Version**
Python-Version: 3.12
opcua-asyncio Version: 1.1.5

sankalpsthakur:
## Summary

Fixes #1771 by awaiting asynchronous node-deletion notifications after the entire batch finishes mutating the address space. Node/reference removal and callback-handle cleanup remain synchronous, so another asyncio task cannot observe a partially deleted batch. A failing callback does not prevent subsequent callbacks from running.

Pending status changes can trigger publishing without waiting for keepalive expiry. This preserves the existing `StatusChangeNotification(BadNodeIdUnknown)` path.

## Validation

- 224 callback, deletion and common tests passed.
- New client/server regression suspends a callback, verifies complete batch cleanup, and retries deletion concurrently. It fails on the previous implementation and passes with this fix.
- Ruff and mypy passed (146 source files).

AI-assisted implementation and tests; code and test evidence reviewed by Codex.

### DIFF
--- a/asyncua/server/address_space.py
+++ b/asyncua/server/address_space.py
@@ -414,11 +414,21 @@ class NodeManagementService:
         self, deletenodeitems: ua.DeleteNodesParameters, user: User = User(role=UserRole.Admin)
     ) -> list[ua.StatusCode]:
         results: list[ua.StatusCode] = []
+        callbacks: list[tuple[int, Callable[..., Any]]] = []
         for item in deletenodeitems.NodesToDelete:
-            results.append(await self._delete_node(item, user))
+            results.append(self._delete_node(item, user, callbacks))
+        # Complete the whole request's address-space changes before any callback
+        # can yield control to another request. Notification does not need a lock.
+        for handle, callback in callbacks:
+            try:
+                await callback(handle, None, ua.StatusCode(ua.StatusCodes.BadNodeIdUnknown))
+            except Exception:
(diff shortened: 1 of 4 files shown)


oroulet:
not sure this is correct, maybe f these adress spaces methods are sync to guarantee nothing happens in between. if we make them async we need to reintroduce locks...

sankalpsthakur:
@oroulet You are right. Batch mutation and callback cleanup now finish synchronously before any notification is awaited (6bec7dd). A regression suspends the first callback and retries deletion concurrently; it fails on the previous code and passes now. All 224 callback/common tests, Ruff and mypy pass.

